PR #4, opened 2019-06-11
scaffold apps from the monorepo

xavdid:
Before, we could pull the entire standalone example app. But, now that they're in the monorepo, we have to pull only those files. 

This will need tweaks once #2 is merged because the path has changed, but that should be an easy change.

Also, I didn't pull all of the example apps in when creating this repo. For apps that are more examples and not a starting-out point (github, onedrive, etc), we were basically doing a glorified git clone on behalf of the user, which seemed silly. There's code in this PR to handle that case. @eliangcs if you think we should go ahead and pull them all in, you can do the import and we can tweak this PR as needed. 

Long term, I think it makes sense to sunset these packages anyway. If we made our scaffolding code a little more complex, users could scaffold triggers into their existing apps or scaffold a new app, which is a blank template + trigger scaffold (the whole process becomes drier).

Lastly, this still needs some tests, though if you agree that the scaffolding plan above is a good one, maybe we can forgo the tests for now.


packages/zapier-platform-cli/src/commands/init.js
```diff
@@ -4,11 +4,7 @@ const constants = require('../constants');
 const appTemplates = require('../app-templates');
 
 const init = (context, location) => {
-  context.line('Welcome to the Zapier Platform! :-D');
-  context.line();
-  context.line(constants.ART);
-  context.line();
-  context.line("Let's initialize your app!");
+  context.line('Initializing New App.');
```
eliangcs: I like the ASCII art! Any reasons why we want to remove it?

xavdid: From [this comment](https://github.com/zapier/zapier-platform-cli/pull/427#discussion_r282629042), it seemed like we wanted to make this a little more "professional" looking, but I don't have strong feelings about the art (just the `:-D`, which I'm not a fan of). Feel free to re-add if you'd like!

xavdid:
@eliangcs ok! so I've merged master back and updated the code for the new locations. did a spot-test on the babel template (`init --template=babel, cd, yarn, test, ✅). 

Didn't have time to make the test, but it looks like the smoke tests should be pretty close to what we want already, so I'll leave that to you. 

Feel free to work on this and merge at your convenience. Thanks a ton!

eliangcs:
@xavdid I notice your original work uses GitHub API, which has a rate limit of 60 requests/hour if a dev doesn't provide `GITHUB_API_TOKEN`. It's very easy to hit this limit when I tested this PR. It's also a bit inconvenient (and too much to ask IMO) for devs having to provide their `GITHUB_API_TOKEN`. So I reworked a bit. Now it downloads the entire monorepo and saves to a cache directory so it doesn't have to download every time the dev runs `zapier init`.

PR #33, opened 2019-06-28
Publish with Travis CI

eliangcs:
Allows to publish npm packages from Travis CI. The process is going to be like:

1. Run `yarn bump` locally and select the versions to bump
2. Git tag `PACKAGE_NAME@VERSION` for each package (e.g. `zapier-platform-cli@8.2.1` and `zapier-platform-core@8.2.1`) is pushed to GitHub
3. Travis sees the git tag and runs `scripts/publish.sh`, which publishes a package to npm registry
4. In the case of publishing `zapier-platform-core`, a `postpublish` hook is configured to build and upload the boilerplate to S3


eliangcs:
We'll probably need to re-encrypt the AWS credentials in `.travis.yml` because Travis [generates a different RSA keypair for each repo](https://docs.travis-ci.com/user/encryption-keys/#encryption-scheme), but I'm not sure. Can @zapier/devops take a look? They are currently copied from the original [`.travis.yml`](https://github.com/zapier/zapier-platform-core/blob/0764ca6e210709f737ca11bbd345850448acfb25/.travis.yml#L29-L30) in `zapier-platform-core`.

.travis.yml
```diff
@@ -2,8 +2,23 @@ language: node_js
 node_js:
   - "8.10"
   - "10"
+before_install:
+  - npm install -g yarn
```
xavdid: as far as I can tell, the presence of `yarn.lock` means that yarn will already be available ([docs](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#yarn-is-supported))

package.json
```diff
@@ -9,6 +9,12 @@
     "test":
       "lerna run --stream --scope zapier-platform-cli --scope zapier-platform-core --scope zapier-platform-schema --scope zapier-platform-legacy-scripting-runner test",
     "smoke-test":
-      "lerna run --stream --scope zapier-platform-cli --scope zapier-platform-core --scope zapier-platform-schema --scope zapier-platform-legacy-scripting-runner smoke-test"
+      "lerna run --stream --scope zapier-platform-cli --scope zapier-platform-core --scope zapier-platform-schema --scope zapier-platform-legacy-scripting-runner smoke-test",
+    "bump":
+      "lerna version --exact --force-publish=zapier-platform-cli,zapier-platform-core,zapier-platform-schema",
```
xavdid: Since i'm scared to actually run and test this command, can you talk a little about what actually happens when this is run?

My main concern is if we'll be able to easily version packages (mostly) individually. I think these will be our most common release patterns:

1. `core` alone: `9.0.1` -> `9.0.2` with no change in the schema dep
1. maj/min/patch release of `schema`, which causes the same in `core` (while depending on the new version). Eg, `schema` goes `9.0.1` -> `9.0.2`, `core` goes from `9.0.2` -> `9.0.3` w/ a bugfix and updated schema dependency
3. major version of cli (`8` -> `9`), no change in cli / core

Are are all of those supported and straightfoward with this command?

packages/core/bin/build-boilerplate.sh
```diff
@@ -2,55 +2,62 @@
...
-rm -f package-lock.json
-rm -fr node_modules
+# Remove generated files
+rm -f zapierwrapper.js definition.json core-*.tgz legacy-scripting-runner-*.tgz
+rm -rf node_modules
 cd ..
 
-# Revert version
-bin/update-boilerplate-dependencies.js revert
+# Don't let yarn pile up useless caches of local packages.
```
xavdid: great attention to detail in these yarn usecases

scripts/publish.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Add auth token to .npmrc
+NPMRC_PATH="$HOME/.npmrc"
+NPM_REGISTRY_NO_PROTO=`echo $NPM_REGISTRY | sed -E 's/^https?://'`
+echo "$NPM_REGISTRY_NO_PROTO/:_authToken=\"$NPM_TOKEN\"" > $NPMRC_PATH
+
+# cd to the package and publish it!
+PKG_PATH=`yarn workspaces info -s | jq -r ".$PKG_NAME.location"`
+cd $PKG_PATH
+npm publish --registry $NPM_REGISTRY
```
xavdid: is there a reason we need to specify the registry here?

Also (though I imagine `yarn publish` uses `npm publish` under the hood), why not use `yarn`?

.travis.yml
eliangcs: The one that Travis preinstalls is old and doesn't support `yarn workspace` command. `npm install -g yarn` is to upgrade the yarn version.

scripts/publish.sh
eliangcs: Great questions!

It's there because I tested the script with a self-hosted registry ([Verdaccio](https://github.com/verdaccio/verdaccio) to be specific). Probably a good idea to leave it there to make testing easy.

I tried to use `yarn publish` but couldn't get it to work. There isn't an easy way to test `yarn publish` with a self-hosted registry. More importantly, unlike we can use an auth token with `npm publish`, it always asks for a password when publishing, which makes it impossible to work on CI.

package.json
eliangcs: Separate versioning is also something I love to see happening, but I narrowed down the scope of this PR, so it only deals with the current situation, which is to version altogether. On the operating level, I think it should be easy to do separate versioning. We probably only need to remove that `--force-publish` part from the `yarn bump` command. The hard part is the application-level dependencies, how we find and remove them in each package.

eliangcs:
@xavdid As a side note, I know we talked about dropping Lerna in favor of `yarn workspace`. But after looking into this, we probably should keep Lerna for its publish-related commands, which give us a nicer CLI.

PR #34, opened 2019-07-12
centralize husky code

xavdid:
The actual content of this is:

* remove `prettier` from devDeps (stays in cli since it's used in scaffold)
* remove `husky`, `lint-staged`, and their configuratioon from all packages
* per [the docs](https://github.com/typicode/husky/blob/master/DOCS.md#multi-package-repository-monorepo), only run husky in the root. 

the exported schema json got run through prettier again and the formatting changed. we jumped a number of minor versions, so something must have changed. In any case, stepped through the command and confirmed that what's in this PR is the output of `prettier` `v1.18.2`, so ¯\\\_(ツ)_/¯

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,4 @@
+packages/cli/scaffold/
+package.json
+package-lock.json
+packages/cli/snippets/next.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,13 @@
 language: node_js
 node_js:
-  - "8.10"
-  - "10"
+  - '8.10'
+  - '10'
 before_install:
   - npm install -g yarn
   - pip install --user awscli
+install:
+  # need to ignore engines until we can ditch node 8 after calendar 2019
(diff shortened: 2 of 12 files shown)


xavdid:
Build is failing because yarn is complaining about an [incompatible version](https://github.com/sindresorhus/execa/blob/071a8154f882d13116a6a91d8691ea150de31753/package.json#L13). 

I ran `yarn --ignore-engines` locally on node 8 and all the tests passed with no issue.

xavdid:
sweet, that did it! also needed to add `AWS_SECRET_ACCESS_KEY=fake` (same for secret) on the repo settings. It was doubly confusing because there's 2 identical settings pages (`travis-ci.org` and `.com`), but only one affects builds (`.com`). Got it now though!

eliangcs:
@xavdid that was my bad. I actually removed the ones on travis.com as I thought they're useless. Anyway, I think it's better to put the keys in `package.json` rather than on Travis. Addressed in #35.

PR #39, opened 2019-07-26
Add re-encrypted credentials for s3 publish

zrochler:


xavdid:
looks like the tests failed because the new value wasn't parsed correctly. once that's working correctly, this should be good to go.

zrochler:
@xavdid it looks like smoke tests are failing, but I have no idea if that's related to this change.

eliangcs:
The tests are failing because https://auth-json-server.zapier.ninja is shut down. Will need to switch to the new host once it's ready.

eliangcs:
Tests are passing now. Merging.

PR #40, opened 2019-07-27
Oclif proof of concept

xavdid:
Currently very in-progress.

### DIFF
--- a/packages/cli/bin/run
+++ b/packages/cli/bin/run
@@ -1,0 +1,6 @@
+#!/usr/bin/env node
+
+require('@oclif/command')
+  .run()
+  .then(require('@oclif/command/flush'))
+  .catch(require('@oclif/errors/handle'));
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -35,6 +35,10 @@
     "validate": "yarn test && yarn lint"
   },
   "dependencies": {
+    "@oclif/command": "1.5.18",
+    "@oclif/config": "1.13.2",
+    "@oclif/plugin-help": "2.2.0",
+    "@oclif/plugin-not-found": "1.2.2",
     "adm-zip": "0.4.13",
     "archiver": "1.3.0",
     "async": "2.1.5",
@@ -70,9 +74,11 @@
     "zip-stream": "1.1.0"
   },
   "devDependencies": {
+    "@oclif/dev-cli": "1.22.0",
     "decompress": "4.2.0",
     "eslint-plugin-mocha": "^5.3.0",
     "fetch-mock": "5.11.0",
+    "globby": "10.0.1",
     "jayson": "2.0.6",
     "litdoc": "1.5.6",
     "markdown-toc": "1.1.0",
@@ -85,5 +91,11 @@
   },
   "bin": {
     "zapier": "zapier.js"
+  },
+  "oclif": {
+    "commands": "./src/ocommands",
+    "plugins": [
+      "@oclif/plugin-not-found"
+    ]
   }
 }
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -45,6 +45,7 @@
     "browserify": "16.2.0",
     "buffer": "5.0.6",
     "cli-table3": "0.5.0",
+    "cli-ux": "5.3.1",
     "colors": "1.1.2",
     "compress-commons": "1.2.0",
     "core-js": "2.4.1",
--- a/packages/cli/src/baseCommand.js
+++ b/packages/cli/src/baseCommand.js
@@ -1,15 +1,58 @@
 const { Command, flags: Flags } = require('@oclif/command');
+const { isEmpty } = require('lodash');
+const { cli: ux } = require('cli-ux');
+
+const { readCredentials, checkCredentials } = require('./utils/api');
 
 class ZapierBaseCommand extends Command {
   _version() {
     return 'blah blah ok'; // this should overwrite the version string, but doesn't?
   }
 
-  async run() {
-    console.log('running in base');
+  log(message) {
+    if (!['json', 'raw'].includes(this.flags.format)) {
+      super.log(message || '');
+    }
   }
 
-  // log(message) {}
+  get prompt() {
+    return ux.prompt;
+  }
+  // AUTH
+  /**
+   * ensures credential file exists and looks correct. Doesn't guarantee creds are still valid, but the API call will blow up if that's the case. Cheap way to test early that user has a session
+   */
+  async probablyLoggedIn() {
+    const creds = readCredentials(false);
+    if (isEmpty(creds)) {
+      this.log('Before you can do that, you need to be logged in.\n');
+      // run login command, which we can't call from here since it'll eventually import this?
+      this.login();
+    }
+  }
+
+  /**
+   * verifies creds, helpful to call before putting the user through tedious input fields and expensive operations (like build)
+   */
+  async definitelyLoggedIn() {
+    try {
+      await checkCredentials();
+    } catch (e) {
+      this.log('Credentials invalid, please run `zapier login`');
+    }
+  }
+
+  // // this should really live on the command itself?
+  // // but i'm not sure how to organize that code and take advantage of the utils in oclif
+  // // or, we have instructions to run `zapier login` instead of actually piping them into it, simplifying our life
+  // async login(firstTime = true) {
+  //   // check credentials
+  //   // get a lot of user input using this.prompt
+  // }
+
+  // async registerApp(title, printWhenDone = true) {
+  //   await this.probablyLoggedIn();
+  // }
 }
 
 ZapierBaseCommand.flags = {
--- a/packages/cli/src/baseCommand.js
+++ b/packages/cli/src/baseCommand.js
@@ -1,0 +1,26 @@
+const { Command, flags: Flags } = require('@oclif/command');
+
+class ZapierBaseCommand extends Command {
+  _version() {
+    return 'blah blah ok'; // this should overwrite the version string, but doesn't?
+  }
+
+  async run() {
+    console.log('running in base');
+  }
+
+  // log(message) {}
+}
+
+ZapierBaseCommand.flags = {
+  format: Flags.string({
+    char: 'f',
+    options: ['a', 'b', 'c', 'raw'],
+    default: 'b'
+  })
+};
+
+ZapierBaseCommand.buildFlags = childFlags =>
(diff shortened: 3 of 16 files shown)


packages/cli/bin/run
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env node
+
+require('@oclif/command')
```
xavdid: this is the new entrypoint for the CLI. All of the code currently in `entry.js` (startup checks, etc) could either go in an [init hook](https://oclif.io/docs/hooks) or get called in a function from `BaseCommand.constructor`

packages/cli/package.json
```diff
@@ -85,5 +92,11 @@
   },
   "bin": {
     "zapier": "zapier.js"
+  },
+  "oclif": {
+    "commands": "./src/ocommands",
+    "plugins": [
+      "@oclif/plugin-not-found"
```
xavdid: this provides our "`buil` is not a command. did you mean `build`?" out of the box!

packages/cli/src/baseCommand.js
```diff
@@ -0,0 +1,58 @@
+const { Command } = require('@oclif/command');
+const { isEmpty } = require('lodash');
+const { cli: ux } = require('cli-ux');
+
+const { readCredentials, checkCredentials } = require('./utils/api');
+
+class ZapierBaseCommand extends Command {
+  _version() {
+    return 'blah blah ok'; // this should overwrite the version string, but doesn't?
```
xavdid: not sure why this doesn't work (we'd want to have it replicate what we have [here](https://github.com/zapier/zapier-platform/blob/9e776f07c8ecd21283b3f57cce1aa7de7af913ca/packages/cli/src/utils/misc.js#L158). There's an [open issue](https://github.com/oclif/oclif/issues/254) for it, so it'll be fixable.

packages/cli/src/baseCommand.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  log(message) {
```
xavdid: instead of doing `context.line`, we can run `this.log` from inside the commands. This means that all of our log lines will need to be in the commands themselves, but that'll probably help us split them apart and make them smaller.

packages/cli/src/baseCommand.js
```diff
@@ -0,0 +1,58 @@
...
+    if (!['json', 'raw'].includes(this.flags.format)) {
+      super.log(message || '');
+    }
+  }
+
+  get prompt() {
```
xavdid: we'll also get to use this instead of our own user-input, which is nice. I don't think we have any more complex inputs, but there are libraries (namely [Inquirer](https://github.com/SBoudrias/Inquirer.js/)) if we want to do checkboxes, etc.

packages/cli/src/baseCommand.js
```diff
@@ -0,0 +1,58 @@
...
+  // AUTH
+  /**
+   * ensures credential file exists and looks correct. Doesn't guarantee creds are still valid, but the API call will blow up if that's the case. Cheap way to test early that user has a session
+   */
+  async probablyLoggedIn() {
+    const creds = readCredentials(false);
+    if (isEmpty(creds)) {
+      this.log('Before you can do that, you need to be logged in.\n');
+      // run login command, which we can't call from here since it'll eventually import this?
+      this.login();
```
xavdid: this was one of the issues I ran into. In our currently CLI, we've got a [special file](https://github.com/zapier/zapier-platform/blob/e8d4fc588bdb2a15059a151057b607a8057b0a06/packages/cli/src/utils/command-invocations.js) that imports the `login` and `register` commands and can run them as functions. this lets us drop into the login flow in another command, so we interrupt the dev less (instead of throwing an error and asking them to run `zapier login`, it just prompts them when they've run an authed command).

So, we've got two major options:

1. move all the login code into a standalone function. It's easily re-useable, but won't be able to take advantage of any of the methods we add onto `BaseCommand` since it's a standalone. We could pass the relevant methods in, but that'll make it really annoying to call (eg `login(this.log, this.x, this.y)`)
2. make exceptions for `login` and `register` and build those full methods onto the base command. Then it's easy to call them in any other command, but seems like a bit of a code smell. They have a doc about [command code reuse](https://oclif.io/docs/running_programmatically#sharing-code-with-modules), but it's a little confusing. I've opened [an issue](https://github.com/oclif/oclif/issues/274) to get more info. Maybe we can eat our cake and have it too!

packages/cli/src/buildFlags.js
```diff
@@ -0,0 +1,26 @@
...
+    default: 'table'
+  }),
+  debug: flags.boolean({
+    char: 'd'
+  })
+};
+
+// didn't destruture these opts because I want them all on one object to be picked from
+const defaultOpts = { debug: true, format: false };
+const buildFlags = (rawFlags = {}, opts = {}) => {
```
xavdid: This was the cleanest way I could think of to let each command opt into what global args they want to inherit. All of them want `--debug`, but only some care about `--format=x` (the ones with table output). It's not the most elegant, but it provides good defaults and is easy to tweak per-command (with minimal boilerplate).

packages/cli/src/ocommands/build.js
```diff
@@ -0,0 +1,43 @@
...
+
+    await build(opts);
+
+    this.log(
+      `\nBuild complete! Moved ${BUILD_PATH} and ${SOURCE_PATH} ! Try the \`zapier upload\` command now.`
+    );
+  }
+}
+
+BuildCommand.rawFlags = {
```
xavdid: `Command.flags` is the standard place to store flags, but if we want to share them across commands (like build + push sharing the same ones), then we need to isolate `build`'s flags (before it's mixed with `debug`, etc).

packages/cli/src/ocommands/push.js
```diff
@@ -0,0 +1,41 @@
...
+    await build(opts, paths);
+    await upload(paths);
+
+    context.line(
+      `\nUpload of ${BUILD_PATH} and ${SOURCE_PATH} complete! Try \`zapier versions\` now!`
+    );
+  }
+}
+
+PushCommand.flags = BuildCommand.flags;
```
xavdid: We can (probably) safely do this. Alternatively, we can have `PushCommand.flags = buildFlags(BuildCommand.rawFlags)`. I think they're effectively the same but this approach shares the reference.

packages/cli/src/utils/api.js
```diff
@@ -84,7 +82,9 @@ const callAPI = (route, options, rawError = false) => {
         }
       }
 
-      if (constants.DEBUG || global.argOpts.debug) {
+      // TODO: pull this into the command? to easily access debug opt
+      // or we do better at using regular node debug lib
+      if (DEBUG || global.argOpts.debug) {
```
xavdid: We should probably use [debug](https://github.com/visionmedia/debug), which will let us not have to pass global args around (which is another area I wanted to address)

packages/cli/src/utils/build.js
```diff
@@ -322,10 +322,16 @@ const maybeNotifyAboutOutdated = () => {
...
-const build = async (zipPath, sourceZipPath, wdir) => {
-  wdir = wdir || process.cwd();
-  zipPath = zipPath || constants.BUILD_PATH;
-  sourceZipPath = sourceZipPath || constants.SOURCE_PATH;
+// TODO: use listr?
+// TODO: fix actually passing the opts around
+const build = async (
+  opts,
+  {
+    zipPath = constants.BUILD_PATH,
```
xavdid: this is a better pattern of defaults. `opts` are the command line flags (dumb paths & include maps).

packages/cli/src/utils/config.js
```diff
@@ -0,0 +1,41 @@
+const { resolve } = require('path');
```
xavdid: i spent a little time re-organizing all of the functions in `utlis/*` into more specific files and then importing directly from them (instead of re-exporting them all, which messes up VSCode's "jump to definition" and function signature previews; I recognize this is a more me-specific problem, but a little spring cleaning is probably in order. `upload` probably doesn't belong in `utils/api.js`)

packages/cli/src/utils/display.js
```diff
@@ -11,7 +11,7 @@ const notUndef = s => String(s === undefined ? '' : s).trim();
 
 const unBacktick = s => s.replace(/\n?`+(bash)?/g, '');
 
-const prettyJSONstringify = obj => JSON.stringify(obj, null, '  ');
+const prettyJSONstringify = obj => JSON.stringify(obj, null, 2);
```
xavdid: functionally the same, but easier to read as a numeral.

packages/cli/bin/run
eliangcs: Is `bin/run` the one we'll have to specify as `main` in `package.json`?

xavdid: yep!

packages/cli/src/baseCommand.js
xavdid: It's [definitely doable](https://oclif.io/docs/running_programmatically), but those docs include this:

> First, it is generally a bad idea to run a command directly as the command exports a user interface, not a code interface. It's a design smell that should rarely (if ever) be used. Generally speaking, it's better to break up the code so that it can be called directly rather than as a command. We'll show this better method first.

That (as I read it) encourages us to move it into a standalone function. But, then it's harder to use utils (namely logging) that are used elsewhere.

I think if I had to pick one, I would just run the other commands directly (like it says not to). I can't think of a cleaner way to do do it. ¯\\\_(ツ)_/¯

packages/cli/src/utils/build.js
xavdid: `opts` will replace calls to `global.argOpts` (such as [here]()) but I hadn't actually hooked that up yet.

xavdid:
I think at this point it's clear that there'd be a fair amount of reorganization/refactoring involved here. If everything looks good (which it seems to, based on your above comments), I think my next steps are:

1. set up the bifurcation so that we can add commands one at a time
2. get a sense of what order we want to do the commands in (easiest to hardest? core util functions first? `build` is probably not a good place to start 😁)
3. figure out how we can have both copies of docs (if needed)
3. estimate the project
4. get rolling

I think if we can re-use util commands for both new and old CLI we will. For some, they'll change enough that we'll just ship both copies of the function until we can safely pull the old one. not ideal, but that's the least destructive way to do that. 

Pending any red flags or objections from you, I'll probably do step 1 this week!

eliangcs:
@xavdid the plan looks good to me!

PR #45, opened 2019-08-07
Fix redundant request params when pulling dynamic dropdown

eliangcs:
Fixes one of the issues in [PDE-1078](https://zapierorg.atlassian.net/browse/PDE-1078). The test case should be self-explanatory.

### DIFF
--- a/packages/legacy-scripting-runner/test/integration-test.js
+++ b/packages/legacy-scripting-runner/test/integration-test.js
@@ -455,18 +455,26 @@ describe('Integration Test', () => {
         isFillingDynamicDropdown: true
       };
       input.bundle.inputData = {
-        dontSendMe: 'test'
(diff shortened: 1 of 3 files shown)


packages/legacy-scripting-runner/test/integration-test.js
```diff
@@ -434,6 +434,42 @@ describe('Integration Test', () => {
       });
     });
 
+    it.only('KEY_pre_poll, dynamic dropdown', () => {
```
xavdid: remember to pull this!

eliangcs:
Though it's a weird behavior of WB, it turns out `bundle.trigger_fields` should still contain the node params even if the node is not of the same trigger as the dynamic dropdown. Pushed commit d4f521c to fix that.

PR #46, opened 2019-08-08
move tooling into root & stop shipping eslint with CLI

xavdid:
rOk! So this is tacking some monorepo tooling debt we hadn't gotten around to doing. The PR is pretty noisy, so i'll summarize the important changes:

1. added a root `.eslintrc`
2. in each of the 4 main `package.json`s:
  a. moved `lint`, `test`, and `smoke-test` into a single command: `validate`
  b. removed `posttest: "yarn lint"`
3. added `zapier-platform-example-app-` to the name of the 3 `oauth1` examples. This doesn't change `init` behavior (we were looking them up by folder name, not `package.json/name`), but does let us easily ignore them in the root lerna commands
4. removed `.eslintrc` and `yarn.lock` from all sub-packages
5. pulled a lot of `devDeps` out of sub packages so they'll all use the same version in root
6. removed eslint app linting from CLI (only actual code change)

## Notes

1. I've verified locally that tests pass. 
2. Linting (and thus CI) will fail, since I didn't want to add more noise to already noisy PR. Once it's ✅, i'll fix all the linting issues. it's mostly using `const` instead of `let` and spacing after comments (`// asdf` instead of `//asdf`).
3. We should probably mark `master` branch as protected, since this is not the first time I've accidentally pushed straight there. I think i've scrubbed that history correctly.


packages/schema/docs/build/schema.md
```diff
@@ -1053,12 +1053,12 @@ How will we get a list of new objects? Will be turned into a trigger automatical
 
 * ```
   { display: { label: 'New User', description: 'Trigger when a new User is created in your account.' },
-    operation: 
+    operation:
```
eliangcs: The trailing spaces were re-added in my last PR. I don't know why they keep coming back. 🤷‍♂ Maybe your IDE automatically strips trailing spaces for all files in the repo? Mine only does that to a file when it's saved.

PR #47, opened 2019-08-08
bundle.request.data should be null for GET requests

eliangcs:
Fixes another issue in [PDE-1078](https://zapierorg.atlassian.net/browse/PDE-1078). For operations (triggers and searches) that default to a GET request, `bundle.request.data` should exactly be `null` instead of an empty object in their scripting methods (`pre_poll`, `pre_search`, etc).

### DIFF
--- a/packages/legacy-scripting-runner/bundle.js
+++ b/packages/legacy-scripting-runner/bundle.js
@@ -108,15 +108,12 @@ const addHookData = (event, bundle, convertedBundle) => {
       );
     }
     if (!convertedBundle.request.content) {
-      convertedBundle.request.content = convertedBundle.request.data;
+      convertedBundle.request.content = convertedBundle.request.data || '';
     }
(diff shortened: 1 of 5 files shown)


packages/legacy-scripting-runner/bundle.js
```diff
@@ -186,14 +183,20 @@ const addResponse = (event, bundle, convertedBundle) => {
...
+      event.name.startsWith('create') ||
+      event.name.startsWith('auth.oauth2') ||
+      event.name.startsWith('trigger.hook.subscribe')
+    ) {
+      requestMethod = 'POST';
+    } else if (event.name === 'trigger.hook.unsubscribe.pre') {
+      requestMethod = 'DELETE';
+    }
+  }
+  if (!requestMethod) {
```
xavdid: minor style thing, but couldn't this be moved a few lines up as an `else` clause? I think the logical flow is the same.

PR #48, opened 2019-08-09
[schema] Allow RefResources to have commas in the labels

xavdid:
This brings the `contact.id.firstName,lastName` dynamic dropdown pattern to CLI! I've confirmed it works:

`contact.id.lastName,firstName,lastName`: 

This involves couple of changes:

1. change `resource` property to `KeySchema` since `dynamic: 'contact'` is an error in the editor
2. That allows us to make more assertions about the shape of actual references, which needs the shape `x.y(.z)?`. We also now include an optional comma clause (but only 1, I got nervous the regex would get unwieldy. We could also simplify it and just allow commas anywhere, but I worry someone will mess it up and it'll be very unclear why their dynamic dropdown isn't working)

Very strictly speaking, this is a semver.major since before it was valid to have `dynamic: 'x'` and on this version, it won't be. That said, that wouldn't have produced a working app anyway, so i don't feel too bad about including this in a lesser release if it comes to that.

--- 

Note that I'm OOO next week, so if this is ✅, I'd appreciate if you merged it for me. Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,11 @@
     }
   },
   "lint-staged": {
-    "*.{js,json}": [
+    "*.{js}": [
       "eslint --fix",
+      "git add"
+    ],
+    "*.{js,json}": [
       "prettier --write",
       "git add"
     ]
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -34,8 +34,11 @@
     }
   },
   "lint-staged": {
-    "*.{js,json}": [
+    "*.{js}": [
```
xavdid: unrelated change, but there are apparently errors if you try to run `eslint` on json ¯\\\_(ツ)_/¯

PR #49, opened 2019-08-14
Add support for more than one commas in dynamic dropdown

eliangcs:
https://github.com/zapier/zapier-platform/pull/48 added support for one comma in a dynamic dropdown, but it looks like we still need to support more than one commas.

### DIFF
--- a/packages/schema/docs/build/schema.md
+++ b/packages/schema/docs/build/schema.md
@@ -887,14 +887,15 @@ Reference a resource by key and the data it returns. In the format of: `{resourc
 #### Details
 
 * **Type** - `string`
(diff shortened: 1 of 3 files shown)


packages/schema/docs/build/schema.md
```diff
@@ -1060,12 +1060,12 @@ How will we get a list of new objects? Will be turned into a trigger automatical
 
 * ```
   { display: { label: 'New User', description: 'Trigger when a new User is created in your account.' },
-    operation:
+    operation:
```
xavdid: @eliangcs Noticed these got re-added again. Not sure the culprit, but I would guess it's tooling and not editor related (since I don't save/format this generated file in the editor ever). 

I've got `prettier -v => 1.18.2`, though I can't imagine prettier is _adding_ trailing spaces?

PR #51, opened 2019-08-21
[cli] Add `debug` package and port `versions` command

xavdid:
Two major areas of changes: 

* cleanup around debugging. removed undocumented and inconsistently-used `constants.DEBUG`. Now everything goes through the `debug` package, which is turned on either in `BaseCommand` right after flag parsing or manually with old-style opts. It also gives us a new way to enable debugging (environment variables: `DEBUG=* zapier versions`). The cool thing about that is the user can be more selective: `DEBUG=zapier:api` will only show api call logs, not flag parsing or anything else. For backwards compatibility and simplicity I kept `--debug`, but we could pull that long term if we wanted to. 
* added `logTable` to the `BaseCommand` and ported `versions`. There's a number of commands that simply show array api responses, so those should be an easy fast-follow once this is merged.


packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -1,12 +1,22 @@
...
 
 const inquirer = require('inquirer');
 
+const DATA_FORMATS = ['json', 'raw'];
+
 class ZapierBaseCommand extends Command {
   run() {
     this._parseFlags();
+    if (this.flags.debug) {
+      this.debug.enabled = true; // enables this.debug on the command
```
eliangcs: I wonder where we initialized `this.debug`. Does oclif initialize it?

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -1,12 +1,22 @@
...
+      require('debug').enable('zapier:*'); // enables all further spawned functions, like API
```
eliangcs: When I run `zapier versions --debug`, I only see `zapier:versions` logs. Should it also include `zapier:api` logs?

packages/cli/src/oclif/ZapierBaseCommand.js
xavdid: yep, that's inherited from the base command: https://github.com/oclif/command/blob/70f67914d0a8a265a563b84c832e01c9f79b7014/src/command.ts#L74

packages/cli/src/oclif/ZapierBaseCommand.js
xavdid: you should, and that's what I see: 

that was running `../../../platform/packages/cli/zapier.js versions --debug`

xavdid: works as expected when running via `~/projects/zapier/platform/node_modules/.bin/zapier versions --debug` as well 🤔 

If you've got a sec, can you add the following line to the top of the `callAPI` method in `cli/src/utils/api.js`? Should be line 45 of that file.

```js
console.log(require('debug').names, debug.enabled);
```

When you run `versions` without `--debug`, it should show `[] false`. When it's true, it should be `[ /^zapier:.*?$/ ] true`

eliangcs: NVM, running `yarn` fixes the issue. Looks like I forgot to run `yarn` in `zapier-platform` so `.bin/zapier` wasn't up-to-date.

xavdid:
Valid json output is fixed in 6f6a6c8! Now we've got a `this.logJson()` that does a standard `console.log` with a nice format, just like before. good catch!

xavdid:
Oh dang, I didn't even notice the different output; that's unintentional. I'll fix that.

xavdid:
Ok! the catch was that we should just straight `console.log` whatever comes out of the formatter. this'll be an easy thing to get unit tests for, but that's out of scope for this PR. I'll merge once tests are green.

PR #53, opened 2019-08-23
bump cli deps

xavdid:
turns out there are a lot of them.

This may end up being a semver major release, mostly because a lot of the sub-deps are (rightfully) dropping support for node 6. We probably don't have an abundance of devs using it, but it's still technically a breaking change.

What's currently up is just CLI, and none of the "hard" ones, where actual code changes on our part are needed. The one i'm most worried about is `node-fetch`, which has material changes and is pretty core to development. we'll have to see what parts of that interface we actually surface to users and advise accordingly.


packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -7,20 +6,10 @@ require('should');
 
 const { getPackageLatestVersion, getPackageSize } = require('../utils/npm');
 const { makeTempDir } = require('../utils/files');
+const { runCommand } = require('../tests/_helpers');
```
eliangcs: Nice improvement 👍

PR #56, opened 2019-08-27
fix(cli): remove small tables

xavdid:
Added tests to ensure logging and properly-formatted json behavior was good. 

While I was doing that, I found out that `--format=small` always throws an error and wasn't used anywhere, so I pulled it. It looks like it was originally meant to be used for printing all the commands real small, but it got dropped somewhere along the way.

Given that it always errored before and now will continue to error (albeit with a different message), this is maybe just a patch change? but it's going out in a major version, so maybe it doesn't matter.

### DIFF
--- a/packages/cli/src/tests/baseCommand.js
+++ b/packages/cli/src/tests/baseCommand.js
@@ -5,7 +5,7 @@ const should = require('should');
 const { stdout } = require('stdout-stderr');
 
 const MESSAGE = 'very cool message';
-const ROWS = [{ id: 123, title: 'helddddddlo' }, { id: 456, title: 'world' }];
+const ROWS = [{ id: 123, title: 'hello' }, { id: 456, title: 'world' }];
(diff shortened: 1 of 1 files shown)


packages/cli/src/tests/baseCommand.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    it('should print valid transformed json', async () => {
+      await TestSampleCommand.run(['table', '--format', 'json']);
+
+      should(stdout.output).containEql('Contact ID');
+      should(stdout.output).containEql('123');
+      should(stdout.output).not.containEql('Here');
+
+      JSON.parse(stdout.output); // shouldn't throw
```
eliangcs: Instead of using `containEql`, I think it'd be more straightforward if we assert the parsed result directly. That is:

```js
const rows = JSON.parse(stdout.output);
rows.length.should.eql(2);
rows[0]['Contact ID'].should.eql('123');
rows[0]['Neat Title'].should.eql('helddddddlo');
``` 

The same applies to all the other places that produce a json output.

PR #59, opened 2019-08-30
feat(cli) port login command

xavdid:
Port the login command and fix some bugs.

Worth noting that this doesn't include the saml changes from https://github.com/zapier/zapier-platform-cli/pull/420. It could, but we're still not positive on the url for that dashboard, so i'm going to hold off. Should be a fast enough follow once this is in though. 

I'm not sure how we can test this or if it's easy to mock stdin.


packages/cli/src/oclif/commands/login.js
```diff
@@ -0,0 +1,99 @@
...
+        )
+      );
+    }
+    const username = await this.prompt(
+      'What is your Zapier login email address? (Ctrl-C to cancel)'
+    );
+    this.log(
+      `\n\nNow you'll enter your Zapier password.\nIf you log into Zapier via the ${colors.green(
+        'log in with Google button'
+      )}, you may not have a Zapier password.\nIf that's the case, go to https://zapier.com/app/login/forgot and create one.\n\n`
```
eliangcs: 💯

packages/cli/src/tests/baseCommand.js
```diff
@@ -1,51 +1,78 @@
...
+  async perform() {
+    this.log(MESSAGE);
+  }
+}
+NoFormatCommand.flags = buildFlags();
 
 describe('BaseCommand', () => {
   describe('log', () => {
     beforeEach(() => {
+      // stdout.print = true;
```
eliangcs: Does this commented line of code have any purpose? I suggest removing if not.

xavdid: it's helpful for testing, though it probably doesn't need to be included. Will pull!

PR #60, opened 2019-08-30
BREAKING CHANGE(cli): Port Logout Command

xavdid:
Add new logout command. Given that it includes the changes made in https://github.com/zapier/zapier-platform-cli/pull/422, it's technically major since it changes logout behavior. 

I don't anticipate people wanting the old "kill all keys" behavior back, but if they do, they'll be able to do the same thing in the UI. if we need to add it back here, we can add a flag. 

Testing wise, I may try and do it with a bunch of mocks and just validate that a bad server response is fine and that an auth file is deleted (or ignored silently).


packages/cli/src/utils/api.js
```diff
@@ -95,10 +95,15 @@ const callAPI = (route, options, rawError = false) => {
...
         const replacementStr = 'raw zip removed in logs';
-        const cleanedBody = _.assign({}, JSON.parse(requestOptions.body), {
-          zip_file: replacementStr,
-          source_zip_file: replacementStr
-        });
+        const requestBody = JSON.parse(requestOptions.body);
+        const cleanedBody = {};
+        for (const k in requestBody) {
+          if (k.includes('zip_file')) {
+            cleanedBody[k] = replacementStr;
```
xavdid: now this won't add `"zip removed"` to requests without zip files in the first place.

packages/cli/src/utils/files.js
```diff
@@ -45,16 +45,14 @@ const writeFile = (fileName, data) => {
   return fse.writeFile(fixHome(fileName), data);
 };
 
-// Returns a promise that deletes the file.
-const deleteFile = fileName => {
-  return new Promise(resolve => {
-    try {
-      fse.unlinkSync(fileName);
```
xavdid: I'm not sure why we were wrapping a sync method in a promise (instead of using the async version or no promise at all). Went ahead and did "just plain sync".

packages/cli/src/oclif/commands/logout.js
```diff
@@ -0,0 +1,36 @@
...
+    this.log(
+      'Preparing to deactivate local deploy key and reset local configs.'
+    );
+    this.startSpinner('Deactivating local deploy key');
+    try {
+      await callAPI('/keys', { method: 'DELETE', body: { single: true } });
+    } catch (e) {
+      this.warn(
+        'Deletion API request failed. If this is unexpected, rerun this command with `--debug` for more info.'
+      );
```
eliangcs: When the server fails to delete the key and returns an error, should we just exit? I know this follows the original implementation, but maybe it makes more sense not to continue if an error happens.

Another issue I found is when I tested `zapier logout` in a directory that doesn't have `.zapierapprc`, the output formatting is off:


I think adding a `this.endSpinner()` before printing the warning message should fix the formatting issue.

xavdid: used a `finally`! good catch.

PR #61, opened 2019-09-03
feat(cli): add saml support

xavdid:
Duplicates https://github.com/zapier/zapier-platform-cli/pull/420. The salient change is that users can paste a deploy key instead of typing an email or password. if we know they're a SAML user based on their email, we can force them into that flow. Otherwise, the experience is the same.

### DIFF
--- a/packages/cli/src/utils/credentials.js
+++ b/packages/cli/src/utils/credentials.js
@@ -1,0 +1,16 @@
+const fetch = require('node-fetch');
+
+const isSamlEmail = async email => {
+  const rawResponse = await fetch(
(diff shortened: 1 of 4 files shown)


packages/cli/src/utils/credentials.js
```diff
@@ -0,0 +1,16 @@
+const fetch = require('node-fetch');
+
+const isSamlEmail = async email => {
+  const rawResponse = await fetch(
+    `https://zapier.com/api/v4/idp-discovery/?email=${encodeURIComponent(
```
eliangcs: Should probably use `BASE_ENDPOINT` here too.

PR #62, opened 2019-09-04
fix(legacy-scripting-runner): add object-first flag to runCreate

ter0:
We need to cover the functionality in python, which returns empty dicts if an empty list or null value is provided.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -341,7 +341,13 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
       }
       throw new Error('JSON results array could not be located.');
     } else if (type.startsWith('object-')) {
-      if (_.isPlainObject(result)) {
+      if (['number', 'string', 'boolean'].includes(typeof result)) {
+        // primitive types that aren't null or undefined
+        return { message: result };
+      } else if (_.isEmpty(result)) {
+        // null, undefined, empty object or array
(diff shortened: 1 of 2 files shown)


packages/legacy-scripting-runner/index.js
```diff
@@ -341,7 +341,9 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
       }
       throw new Error('JSON results array could not be located.');
     } else if (type.startsWith('object-')) {
-      if (_.isPlainObject(result)) {
+      if (result == null) {
```
eliangcs: We can use `_.isEmpty(result)` to cover the cases of `result` being `null`, `undefined`, an empty object, or an empty array.

Also, looking at our backend [code](https://github.com/zapier/zapier/blob/68c3050bebefb954de0c0c8cbb7ac2764529852a/web/backend/zapier/developer/api_base.py#L1525-L1527), we also deal with primitive types (numbers and strings). So we should cover that as well. To summarize, the code should be:

```js
if (['number', 'string', 'boolean'].includes(typeof result)) {
    // primitive types that aren't null or undefined
    return {message: result};
} else if (_.isEmpty(result)) {
    // null, undefined, empty object or array
    return {};
} else if (_.isPlainObject(result)) {
// the rest is the same
```

ter0:
Thanks for the review! I can still never remember all the equality related bits for JS. 😛

PR #63, opened 2019-09-06
refactor(cli,schema): port and refactor validate command

eliangcs:
The original intent of this PR is to:

* switch the endpoint `zapier validate` hits from `/api/platform/cli/style-check` to `/api/platform/cli/check`.
* pass `app_id` and `version` to the validation endpoint, which enables the server to validate areas other than the app definition ("D" checks), such as states and marketing ("S" and "M" checks).
* add `User-Agent: zapier-platform-cli/x.y.z` header to all API calls to zapier.com. This helps the server identify the CLI client version and respond accordingly.
* adjust `zapier promote` command to accept the new format of error messages from the server.
* update schema doc links to the monorepo

But while I was there, I also ported the following commands to the oclif side, and it turned out to be a big effort because of the dependencies.

* `zapier validate`
* `zapier test` because it depends on `zapier validate`
* `zapier promote`
* `zapier migrate` because it depends on `zapier promote` (insert frodo-finally-over meme here)


packages/cli/src/oclif/commands/promote.js
```diff
@@ -0,0 +1,153 @@
...
+
+const BaseCommand = require('../ZapierBaseCommand');
+const utils = require('../../utils');
+const { buildFlags } = require('../buildFlags');
+
+const serializeErrors = errors => {
+  const opener = 'Promotion failed for the following reasons:\n\n';
+  if (typeof errors[0] === 'string') {
+    // errors is an array of strings
+    return opener + errors.map(e => `* ${e}`).join('\n');
```
xavdid: it wasn't like this before, but it might look nice to indent the bullets:

```js
errors.map(e => `  * ${e}`).join('\n')
```

Same below. In fact, I might even put this on `BaseCommand` as `printList(opener, errors)`

packages/cli/src/oclif/commands/promote.js
```diff
@@ -0,0 +1,153 @@
...
+        {
+          method: 'PUT',
+          body
+        },
+        true
+      );
+    } catch (response) {
+      this.stopSpinner();
+
+      const activationUrl = _.get(response, 'json.activationInfo.url');
```
xavdid: super minor, but I've seen a linting rule that enforces the second argument to `_.get` be an array of strings (`['json', 'activationInfo', 'url']`), and I like that. For longer chains, prettier will put each item on a new line, which makes it nice and readable.

Separately, if it's easy, I'd destructure the `lodash` require (`const { get } = require('lodash')`) to match the other more explicit requiring.

packages/cli/src/oclif/commands/promote.js
```diff
@@ -0,0 +1,153 @@
...
+      await utils.callAPI(
+        url,
+        {
+          method: 'PUT',
+          body
+        },
+        true
+      );
+    } catch (response) {
+      this.stopSpinner();
```
xavdid: maybe move this `stop` farther down so we can stop it with a ✅ or ❌ depending on if it's an actual error or not?

packages/cli/src/oclif/commands/test.js
```diff
@@ -0,0 +1,89 @@
...
+
+const BaseCommand = require('../ZapierBaseCommand');
+const ValidateCommand = require('./validate');
+const constants = require('../../constants');
+const utils = require('../../utils');
+const { buildFlags } = require('../buildFlags');
+
+class TestCommand extends BaseCommand {
+  async perform() {
+    if (!this.flags['skip-validate']) {
```
xavdid: minor preference for flags being camel case rather than hyphenated so we can use `this.flags.skipValidate` instead. That's a breaking change though.

packages/cli/src/oclif/commands/test.js
```diff
@@ -0,0 +1,89 @@
...
+
+TestCommand.args = [];
+
+TestCommand.examples = [
+  'zapier test',
+  'zapier test -t 30000 --grep api --skip-validate'
+];
+TestCommand.description = `Tests your app via \`npm test\`.
+
+This command is effectively the same as \`npm test\`, except we also validate your app and set up the environment. We recommend using mocha as your testing framework.`;
```
xavdid: Mocha is decent, but i wonder if we don't be a little more framework-neutral and not push them towards anything specific here. 

Also, rather than support specific flags to pass into any particular framework, we could pass along the rest of argv directly to npm/yarn. that way, we won't get anyone writing in saying "Could you add support for the `--xyz` flag to `zapier test`?" i'm down to mention these potentially helpful args in the docstring though!

packages/cli/src/oclif/commands/validate.js
```diff
@@ -0,0 +1,108 @@
...
+      );
+      process.exitCode = 1;
+    } else {
+      this.log('This project is structurally sound!');
+    }
+
+    let checkResult = [];
+    if (0 && (this.flags['no-server'] || process.exitCode === 1)) {
+      if (process.exitCode === 1) {
+        this.log(
```
xavdid: is there a reason not to use `this.error(message)` instead and skip the rest of the function?

packages/cli/src/oclif/commands/validate.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+ValidateCommand.flags = buildFlags({
+  commandFlags: {
+    format: flags.string({
+      char: 'f',
+      options: Object.keys(formatStyles),
+      default: 'row'
+    }),
+    'no-server': flags.boolean({
```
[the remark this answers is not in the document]
xavdid: also, thinking it over, i'm not sure that `no-server` is clear enough for the user. Without knowledge of CLI, it's opaque which parts of the check(s) happen on the server vs locally. I think for them, the distinction of "structural validity" and "style validity" is clear (though i know "style checks" are a deprecated/loaded term). 

All that is to say I'm not sure I'd call it this, but I don't have a better suggestion at this time.

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -0,0 +1,117 @@
+const { flags } = require('@oclif/command');
+
+const BaseCommand = require('../ZapierBaseCommand');
+const PromoteCommand = require('./promote');
+const utils = require('../../utils');
```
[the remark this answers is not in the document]
eliangcs: That makes sense. Fixed in 04c7a41.

packages/cli/src/oclif/commands/test.js
```diff
@@ -0,0 +1,89 @@
...
+const ValidateCommand = require('./validate');
+const constants = require('../../constants');
+const utils = require('../../utils');
+const { buildFlags } = require('../buildFlags');
+
+class TestCommand extends BaseCommand {
+  async perform() {
+    if (!this.flags['skip-validate']) {
+      await ValidateCommand.run([]);
+      if (process.exitCode) {
```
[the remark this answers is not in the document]
eliangcs: I thought it was in the old test command. Turns out we continue to run tests even if the validation didn't pass. Removed in c906340.

packages/cli/src/oclif/commands/test.js
```diff
@@ -0,0 +1,89 @@
...
+      '--',
+      `--timeout=${this.flags.timeout}`
+    ];
+    if (this.flags.grep) {
+      argv.push(`--grep=${this.flags.grep}`);
+    }
+
+    this.log('Running test suite.');
+
+    const packageManager = this.flags.yarn ? 'yarn' : 'npm';
```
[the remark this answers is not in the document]
eliangcs: `-y` isn't probably a good shortcut for `--yarn`. I think more commonly `-y` is used to suppress prompts by answering yes to all the questions.

I like the other ideas! But I'm gonna leave them in future PRs because it's kind of out of the scope of this PR.

packages/cli/src/oclif/commands/validate.js
```diff
@@ -0,0 +1,108 @@
...
+      this.log(
+        'Your app is structurally invalid. Address concerns and run this command again.'
+      );
+      process.exitCode = 1;
+    } else {
+      this.log('This project is structurally sound!');
+    }
+
+    let checkResult = [];
+    if (0 && (this.flags['no-server'] || process.exitCode === 1)) {
```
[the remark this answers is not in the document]
eliangcs: Good catch! I was debugging and forgot to remove. Removed in b1aa7d7.

packages/cli/src/oclif/commands/validate.js
```diff
@@ -0,0 +1,108 @@
...
+      );
+    } else {
+      this.log('Your app looks great!\n');
+    }
+  }
+}
+
+ValidateCommand.flags = buildFlags({
+  commandFlags: {
+    format: flags.string({
```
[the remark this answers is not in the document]
eliangcs: Yeah, it was repeated because I wanted to give it a different default ("row" rather than "table"). But on the second thought, I think it's fine to use the default default format of "table" because "table" format becomes "row" automatically when the screen width doesn't fit, which is very likely to happen. Fixed in 1632a8c.

packages/cli/src/oclif/commands/validate.js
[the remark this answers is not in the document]
eliangcs: That's a good callout! I don't have a better idea either. Let me just change back to `without-style` in 8899c20.

packages/cli/src/oclif/commands/validate.js
```diff
@@ -0,0 +1,108 @@
...
+const { flattenCheckResult, formatStyles } = require('../../utils/display');
+
+class ValidateCommand extends BaseCommand {
+  async perform() {
+    this.log('Validating project locally.');
+
+    const errors = await utils.localAppCommand({ command: 'validate' });
+
+    const newErrors = errors.map(error => {
+      error = Object.assign({}, error);
```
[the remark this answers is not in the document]
eliangcs: I really need to get used to JavaScript spread syntax. Thanks for the tip! Fixed in e0d5487.

packages/cli/src/utils/misc.js
```diff
@@ -29,9 +30,9 @@ const runCommand = (command, args, options) => {
   }
 
   options = options || {};
-  if (global.argOpts.debug) {
-    console.log('\n');
-    console.log(
+  if (_.get(global, ['argOpts', 'debug']) || debug.enabled) {
```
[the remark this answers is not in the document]
eliangcs: Makes sense! Fixed in c4796a6.

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -0,0 +1,117 @@
...
+    required: true,
+    description: 'The version FROM which to migrate users.'
+  },
+  {
+    name: 'toVersion',
+    required: true,
+    description: 'The version TO which to migrate users.'
+  },
+  {
+    name: 'percent',
```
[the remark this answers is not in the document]
eliangcs: I don't see oclif providing a `type` option for args. But it has a `parse` option for us to specify a parse function. Fixed in 3dd57f5.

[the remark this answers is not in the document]
eliangcs: BTW, `parseInt('100%')` correctly gives a number `100`. So despite we're removing the `%` from the command help text, `x%` still works. Not a breaking change.

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -0,0 +1,117 @@
...
+      promoteFirst = this.confirm(
+        `Do you want to promote ${toVersion} to production first?`,
+        false,
+        true
+      );
+    }
+
+    if (promoteFirst) {
+      PromoteCommand.printMigrateHint = false;
+      await PromoteCommand.run([toVersion]);
```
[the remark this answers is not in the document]
eliangcs: I'd vote for 2.

The oclif doc says  "(running a command from another command) a design smell". I think the statement is arguable. It doesn't give any objective reasons why it's bad. The only reason I see in the doc is a matter of personal taste.

For our use case, we have commands that are, by definition, are composed of subcommands. For example, `zapier push` *is* `zapier build` followed by `zapier upload`, so it's intuitive to call those subcommands from the container command. It also results in clean and easy-to-maintain code, which is the objective reason to take into account.

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -0,0 +1,117 @@
...
+      );
+      promoteFirst = this.confirm(
+        `Do you want to promote ${toVersion} to production first?`,
+        false,
+        true
+      );
+    }
+
+    if (promoteFirst) {
+      PromoteCommand.printMigrateHint = false;
```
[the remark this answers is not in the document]
eliangcs: Nice suggestion! Fixed in 76f96eb.

packages/cli/src/oclif/commands/promote.js
eliangcs: This is what it looks like without the indent:



The output message is thrown with `new Error(message)` instead of printing, so `printList()` on `BaseCommand` probably isn't what we want.

And since the entire error message is already indented, so I wouldn't add another indent to the bullets.

packages/cli/src/oclif/commands/promote.js
eliangcs: Changed it to array path in 93815f0.

I'd keep `_.get` though. I like explicit requiring most of the time, but this one is an exception. `get` is a super generic name. Keeping it `_.get` actually makes it clear that it's lodash's `get` without requiring us to look for where we import it in the top section.

packages/cli/src/oclif/commands/promote.js
eliangcs: Good catch! Fixed in 5abaa9c.

packages/cli/src/oclif/commands/test.js
eliangcs: Yeah, probably not worth it. Not only we have a lot of flags already using hyphens, but also most popular CLI tools (such as git and grep) follow this convention.

packages/cli/src/oclif/commands/validate.js
eliangcs: This was how it is in the original command. I also think for `zapier validate` this is a normal case instead of an error. `zapier validate`'s job is to find violations, so it's normal that the app didn't validate. I believe `this.error` is for real errors. For `zapier validate`, it could be `zapier-platform-core` hasn't been installed yet or the developer runs in a random directory.

packages/cli/src/oclif/commands/test.js
eliangcs: I like your idea! But I feel this is out of the scope of this PR. Maybe we should do it as seperate work?

eliangcs:
@xavdid thanks for the detailed review! I've adopted most of your suggestions. But there's some disagreement, so it'd be great if you can take a second 👀.

packages/cli/src/oclif/commands/validate.js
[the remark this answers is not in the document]
xavdid: no worries. 😁we've got a tough job going between python and js so much. spread and destructuring (especially w/ defaults) make js feel a lot cleaner to me.

packages/cli/src/oclif/commands/migrate.js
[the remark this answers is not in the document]
xavdid: oh nice! yeah, seems like 6 one way and a half-dozen the other.

packages/cli/src/oclif/commands/promote.js
xavdid: that's funny, the indentation relative to the parent message imparts ownership to me. It's the difference between:

```
promotion failed because:
  * a
  * b
  * c
validation failed because:
  * a
  * b
  * c
```

vs

```
promotion failed because:
* a
* b
* c
validation failed because:
* a
* b
* c
```

whereas the latter feels more clear. But, given that it's pretty clear and there's no secondary list, i'm happy either way here. 😁

packages/cli/src/oclif/commands/promote.js
xavdid: Fine by me! the other (maybe silly) option is to rename it (`const {get: lodashGet} = require('lodash')`), but I don't care that much. 

Worth nothing that [optional chaining](https://github.com/tc39/proposal-optional-chaining) is in stage 3, which means it'll be in JS for real in the next year (and typescript starting in nov 👀), so we'll eventually be able to do `response?.json?.activationInfo?.url`! That has no bearing on this pr, but it's fun info.

packages/cli/src/oclif/commands/test.js
xavdid: that's fine with me! the only reservation would be releasing the current items as CLI 9 and then having to do a second breaking change release soonish after. I'd rather do all the breaks together. 

In terms of this PR, i'm happy to see it merged as long as we take a holistic look at the args before we cut a release 👍

packages/cli/src/oclif/commands/validate.js
xavdid: fair! there's also a built-in `this.warn`. logs with a yellow `>`, iirc. A little more eye-catching, but doesn't throw.

packages/cli/src/oclif/commands/migrate.js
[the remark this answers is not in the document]
xavdid: I'm down for that. That's a good point about them not really backing up their claim - I should be a little more critical about their word as law.

A thing we should be cognizant of is maintaining consistent flags across shared commands. For instance, I would expect `push` to take all the flags available to `build` and `upload`. I think that's a lot easier now with the way we're building the flag objects (we can import them from a shared file if need be).

packages/cli/src/utils/misc.js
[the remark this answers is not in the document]
xavdid: Cool. we'll be able to pull it once everything is on oclif, but that's the way we'll need to do it for now.

packages/cli/src/oclif/commands/validate.js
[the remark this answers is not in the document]
xavdid: ok! we can also make the default configurable if that's a thing we find useful elsewhere.

packages/cli/src/oclif/commands/test.js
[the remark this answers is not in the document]
xavdid: It's totally more common for `--yes`, but I can't think of another letter we could use for a shorthand instead (besides picking one randomly). Given that this command has no user input, i'm not too bothered by breaking convention. 

And yeah, no worries on the out-of-scope bit. I'll add a note for my backlog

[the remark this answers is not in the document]
xavdid: to add some context, I ran into this when working on another project, [typed-install](https://github.com/xavdid/typed-install). I added a flag to use `yarn` over `npm`, defaulting to `npm`. That part worked great, but I'd always run it with the default args by accident (`typedi lodash`) and it would use `npm` even though I had previously used `yarn` in that project. I ended up fixing it with a [one liner](https://github.com/xavdid/typed-install/blob/ec81b012044725923a6f9c4592b0b8d6f81c0bc2/src/index.ts#L93), but before I did I was surprised how disruptive it was to need to remember to include a flag with every invocation.

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -0,0 +1,121 @@
...
+const BaseCommand = require('../ZapierBaseCommand');
+const PromoteCommand = require('./promote');
+const { callAPI, getLinkedApp } = require('../../utils/api');
+const { buildFlags } = require('../buildFlags');
+
+class MigrateCommand extends BaseCommand {
+  async perform() {
+    const percent = this.args.percent;
+    if (isNaN(percent) || percent < 1 || percent > 100) {
+      throw new Error('`PERCENT` must be a number between 1 and 100.');
```
xavdid: If you want, you could use `this.error(message)` instead of `throw new Error`. Their [custom error](https://github.com/oclif/errors/blob/b6903528d9d9a58a87347f3f86049393cc2e8d83/src/errors/cli.ts#L10-L51) doesn't do a ton, but it does clean the stack and do nice indenting and stuff. Same goes for all other thrown errors.

packages/cli/src/oclif/commands/test.js
eliangcs: I agree it's better to do all the breaks together. So yeah, we should do an overall review before we release 9.0.0.

packages/cli/src/oclif/buildFlags.js
```diff
@@ -13,11 +13,21 @@ const baseFlags = {
...
+    hidden: true
   })
 };
 
 // didn't destruture these opts because I want them all on one object to be picked from
-const defaultOpts = { debug: true, format: false };
+const defaultOpts = {
+  debug: true,
+  format: false,
+  invokedFromAnotherCommand: true
```
xavdid: probably want this to default to `false`, right? most of the time, we're calling the commands from CLI (unless i'm getting the implementation reversed)

eliangcs: Setting `true` here means all the commands have a hidden `--invokeFromAnotherCommand` flag by default. This flag is true only if you pass in `--invokeFromAnotherCommand` when running the command. Does that make sense?

xavdid: yep, this was totally me mixing up "defaults to being present on all commands" (👍) with "defaults to being true for all commands" (👎). all good here!

PR #67, opened 2019-09-19
fix(legacy-scripting-runner): allow hooks without a polling URL

eliangcs:


eliangcs:
Closing in favor of https://github.com/zapier/zapier/pull/31159.

PR #68, opened 2019-09-20
chore(example-apps) prettier & lint all apps

xavdid:
run prettier and our eslint over all example apps. Also, removed a few instances of `body: JSON.stringify(body)` since that's handled in core. 

No other serious code changes.


xavdid:
merging since it's just formatting and it's not worth the time to inspect each one.

PR #69, opened 2019-09-21
fix(schema): Make label & description optional for hidden actions

xavdid:
Noticed the other day that hidden triggers needed label/description. This loosens that requirement. 

The one thing I still need to test is that the editor plays nicely with this. I'll do that before I merge it.

### DIFF
--- a/packages/schema/lib/functional-constraints/labelWhenVisible.js
+++ b/packages/schema/lib/functional-constraints/labelWhenVisible.js
@@ -1,0 +1,31 @@
+const _ = require('lodash');
+const jsonschema = require('jsonschema');
+
+const actionTypes = ['triggers', 'searches', 'creates'];
+
+const labelWhenVisible = definition => {
+  const errors = [];
+
+  for (const actionType of actionTypes) {
+    const group = definition[actionType] || {};
+    _.each(group, (action, key) => {
+      const { display } = action;
+      if (!display.hidden && !(display.label && display.description)) {
+        errors.push(
+          new jsonschema.ValidationError(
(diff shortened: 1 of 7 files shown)


packages/schema/lib/functional-constraints/labelWhenVisible.js
```diff
@@ -0,0 +1,31 @@
...
+  for (const actionType of actionTypes) {
+    const group = definition[actionType] || {};
+    _.each(group, (action, key) => {
+      const { display } = action;
+      if (!display.hidden && !(display.label && display.description)) {
+        errors.push(
+          new jsonschema.ValidationError(
+            `visible actions must have a label and description`,
+            action,
+            `/${_.capitalize(actionType)}Schema`,
```
eliangcs: Should we link to `BasicDisplaySchema` instead?

packages/schema/lib/functional-constraints/labelWhenVisible.js
```diff
@@ -0,0 +1,31 @@
...
+            `instance.${key}.key`,
```
eliangcs: Ideally, I think the path should be `instance.${key}.display.label` or `instance.${key}.display.description` to be more accurate, but I'm ok with `instance.${key}.display` too.

xavdid: went with `instance.creates.foo.display`, which is the path to where the dev will need to fix the issue. Good catch!

xavdid:
confirmed locally this works without issue in the editor!

PR #72, opened 2019-09-27
fix(cli): handle passing checks in validate command

eliangcs:
Works with https://github.com/zapier/zapier/pull/31427.

### DIFF
--- a/packages/cli/src/oclif/ZapierBaseCommand.js
+++ b/packages/cli/src/oclif/ZapierBaseCommand.js
@@ -143,15 +143,11 @@ class ZapierBaseCommand extends Command {
   }
 
   startSpinner(message) {
-    if (this._shouldPrintData()) {
-      startSpinner(message);
-    }
+    startSpinner(message);
   }
 
   stopSpinner({ success = true, message = undefined } = {}) {
-    if (this._shouldPrintData()) {
-      endSpinner(success, message);
-    }
+    endSpinner(success, message);
   }
 }
 
--- a/packages/cli/src/oclif/ZapierBaseCommand.js
+++ b/packages/cli/src/oclif/ZapierBaseCommand.js
@@ -143,11 +143,15 @@ class ZapierBaseCommand extends Command {
   }
 
   startSpinner(message) {
-    startSpinner(message);
+    if (this._shouldPrintData()) {
+      startSpinner(message);
+    }
   }
 
   stopSpinner({ success = true, message = undefined } = {}) {
-    endSpinner(success, message);
+    if (this._shouldPrintData()) {
+      endSpinner(success, message);
+    }
   }
 }
 
--- a/packages/cli/src/oclif/commands/validate.js
+++ b/packages/cli/src/oclif/commands/validate.js
@@ -65,7 +65,7 @@ class ValidateCommand extends BaseCommand {
 
     const checkIssues = flattenCheckResult(checkResult);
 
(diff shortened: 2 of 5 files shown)


packages/schema/docs/build/schema.md
```diff
@@ -1059,12 +1059,12 @@ How will we get a list of new objects? Will be turned into a trigger automatical
 
 * ```
   { display: { label: 'New User', description: 'Trigger when a new User is created in your account.' },
-    operation:
+    operation:
```
xavdid: it's maybe not worth your time to fix, but might be nice to solve why these keep getting added.

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -143,11 +143,15 @@ class ZapierBaseCommand extends Command {
   }
 
   startSpinner(message) {
-    startSpinner(message);
+    if (this._shouldPrintData()) {
```
xavdid: were you seeing issues before adding this? my understanding is that `ora` uses `stderr`, so it doesn't actually interfere with formatting. 

For instance: 



Not a hard blocker, but the spinners add a much better UX (rather than the command seeming to hang), so I want to make sure they're shown as much as possible.

packages/cli/src/oclif/commands/validate.js
```diff
@@ -57,8 +58,14 @@ class ValidateCommand extends BaseCommand {
...
     }
 
+    const doneMessage = checkResult.passes
+      ? `Running app checks ... ${checkResult.passes.length} checks passed`
+      : undefined;
+    this.stopSpinner({ message: doneMessage });
+
     const checkIssues = flattenCheckResult(checkResult);
 
+    this.log("\nHere're the issues we found:");
```
xavdid: While I believe `here're` is [technically correct](https://english.stackexchange.com/questions/36532/using-the-contraction-of-are), I'm not sure I've ever read it like that and it comes off as very unusual. I think `Here are` is clearer here.

packages/schema/docs/build/schema.md
eliangcs: I finally spent some time looking into this. It's because `util.inspect` generates slightly different outputs. On Node 8, which I'm using, it outputs trailing spaces, while on Node 12, which you're probably using, it doesn't generate trailing spaces.

Added some code to ensure a consistent behavior in 72c6f77.

packages/cli/src/oclif/ZapierBaseCommand.js
eliangcs: Good catch! I didn't know spinners go to stderr. Undid it in a4ca9e7.

packages/cli/src/oclif/commands/validate.js
eliangcs: Thanks! I've been writing this for a long time until TIL **here're** (or **there're**) isn't a thing in English. Fixed in a455de9.

xavdid: For sure! `here` is contracted lots of other commonly-accepted ways, but for some reason that's not one of them. English is a mess.

packages/schema/docs/build/schema.md
xavdid: ah yeah, i'm mostly on 10 or 12. that'll do it! Thanks for digging in.

PR #74, opened 2019-09-28
chore(examples): Revalidate all examples, remove most of them

xavdid:
This is a long-winded way to do [PDE-1077](https://zapierorg.atlassian.net/browse/PDE-1077), where we condense the `example-apps` directory into the `examples` folder, which has way fewer files. 

This PR consisted of copying a selected set of examples into a new folder. Eventually, we'll scaffold from this new folder and pull those into the docs (removing the snippets folder). In the interest of keeping this PR limited in scope, the rest of that functionality is left for later.

The main goal was to reduce the amount of boilerplate that we need to maintain, which this mostly does. There's some repetition between the auth configs, but it's pretty good. I also added linting to them.

When we update the scaffolding, we'll also want to add testing to ensure they all work out of the box. I've validated them manually for now. The big exception is `session`, which uses the now-defunct `zapier-httpin`. Will need to see if there's another option available.

Each of these new examples exports an object with a `config` property (the previous export), plus optional `befores` and `afters` middleware arrays.


xavdid:
Oh nice! It seems like there are lots of test auth servers, json servers, httpbins, mock apis, etc floating around. do we have a full list anywhere? internal or external, it'd probably be a helpful reference.

good call on the httpbin. I did a big find-and-replace. also replaced all `http` urls with `https` (and verified that they're still fine).

eliangcs:
@xavdid I'm not aware of a full list of mock servers we're using, but I think you've listed them all. Ideally, we should also movie all the mockapi.io to auth-json-server if possible, so we only have httpbin.zapier-tooling.com and auth-json-server.

xavdid:
just to follow up - I fixed everything but `httpbin.org` -> `httpbin.zapier-tooling.com`. All the core tests fail after that switch: 



Looks like ours wraps things in arrays where the public one doesn't. since it's not urgent, i'll leave it for another day.

PR #76, opened 2019-10-03
docs(cli) add new commands to docs

xavdid:
We [noticed](https://zapier.slack.com/archives/C5Z9BP4U9/p1570028316375200) that the slack invite url was expired. I fixed it in master (https://github.com/zapier/zapier-platform/commit/74711f04f3bf1832f259387510e066f87cac24b8), and as a result, noticed that `cli.docs` wasn't getting run in the pre-commit. 

That's fixed here, plus adding code to format the command into markdown (so it's picked up by the `docs` command). 

This code was simplified from [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/master/src/command.ts), which we use. It would have been cool to be able to use the formatting functions without their `bold` and joining and stuff, so that's maybe a PR i'll make at some point. In any case, this is good for now.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -197,7 +197,7 @@ After the deprecation date has passed it will be safe to delete that app version
 * `-d, --debug` | Show extra debugging output
 
 **Examples**
-* zapier deprecate 1.2.3 2011-10-01
+* `zapier deprecate 1.2.3 2011-10-01`
 
 
 ## describe
@@ -389,8 +389,8 @@ After running this, you'll have a new example app in your directory. If you re-r
 * `-d, --debug` | Show extra debugging output
 
 **Examples**
-* zapier init ./some/path
-* zaper init . --template typescript
+* `zapier init ./some/path`
+* `zaper init . --template typescript`
 
 
 ## invite
@@ -587,9 +587,9 @@ Note: since a migration is only for non-breaking changes, users are not emailed 
 * `-d, --debug` | Show extra debugging output
 
 **Examples**
-* zapier migrate 1.0.0 1.0.1
-* zapier migrate 1.0.1 2.0.0 10
-* zapier migrate 2.0.0 2.0.1 --user=user@example.com
+* `zapier migrate 1.0.0 1.0.1`
+* `zapier migrate 1.0.1 2.0.0 10`
+* `zapier migrate 2.0.0 2.0.1 --user=user@example.com`
 
 
 ## promote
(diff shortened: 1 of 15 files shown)


xavdid:
Ah, looks like a naming collision with something else in oclif. Using a unique name fixes it (and I confirmed both code paths work 😬). Fixed in 045398d.

Removing `static` threw an error because in `bin/docs`, we call `command.usage(name)`. Without that being static, that line would need to be `new command().usage(name)`. No need to change that though, because we renamed it.

docs/cli.md
```diff
@@ -465,6 +559,69 @@ $ zapier logs --type=http
...
+
+## promote
+
+> Promotes a specific version to public access.
+
+**Usage**: `zapier promote VERSION`
+
+Promotes an app version into production (non-private) rotation, which means new users can use this app version.
+
+* This [1mdoes[22m mark the version as the official public version - all other versions & users are grandfathered.
```
eliangcs: In PR https://github.com/zapier/zapier-platform/pull/63, I used ANSI codes in the command help text. That's why it got misrendered here in this markdown doc. We don't have to fix it in this PR, but I just wanted to point out, before releasing 9.0.0, as part of the effort standardizing our CLI style, we should style the text properly based on the destination format. Currently, we have these formats:

* ANSI codes for the terminal. We should get rid of markdown syntax such as backticks (`` `code` ``) and double asterisks (`**bold**`) and use appropriate ANSI codes. 
* HTML
* Markdown

docs/cli.md
```diff
@@ -320,6 +358,41 @@ $ │ logout      │ zapier logout                         │ Deactivates all
...
+**Arguments**
+* (required) `path` | Where to create the new app. If the directory doesn't exist, it will be created. If the directory isn't empty, we'll ask for confirmation
+
+**Flags**
+* `-t, --template` | The template to start your app with. One of `[minimal | trigger | search | create | basic-auth | custom-auth | digest-auth | oauth2 | oauth1-trello | oauth1-tumblr | oauth1-twitter | session-auth | dynamic-dropdown | files | middleware | resource | rest-hooks | search-or-create | babel | typescript | github | onedrive]`. Defaults to `minimal`.
+* `-d, --debug` | Show extra debugging output
+
+**Examples**
+* `zapier init ./some/path`
+* `zaper init . --template typescript`
```
eliangcs: Typo `zaper`. Not introduced by this PR, but should be a quick fix.

docs/cli.md
xavdid: Yeah, good call! I think in the docs on the commands themselves we use `chalk.bold` (etc) for nice styling. We can use [strip-ansi](https://github.com/chalk/strip-ansi) to pull them out for the markdown. 

Or (and i'm not sure how hard this is; it might just be a straight replace?) we take the ansi docs and replace the ansi with markdown and then build html out of the markdown.

PR #80, opened 2019-10-11
feat(cli): add analytics

xavdid:
Adds simple, unobtrusive analytics to CLI commands.

I've triple checked that we don't store any sensitive information in `argOpts`, just in the args themselves. 

For example:

```
% DEBUG=zapier:analytics ../zapier.js env 1.0.0 CLIENT_SECRET 1234
  zapier:analytics sending { command: 'env', validCommand: true, numArgs: 3, flags: {} } +0ms
Preparing to set environment CLIENT_SECRET for your 1.0.0 "myApp".

✔ Setting CLIENT_SECRET to "1234"

Environment updated! Try viewing it with `zapier env 1.0.0`.
```

Also some tweaks to make sure we don't print the result message before we're actually done with the request.

TODO:

- [x] merge and deploy https://github.com/zapier/zapier/pull/32056
- [x] add command that enables/disables analytics
- [x] read/write the above configuration 
- [x] release this branch as `8.4.0`

**Note** that this branch was not made off master, but the `8.3.0` release. That means we'll be able to release a new version with minimal changes, but won't be able to (easily? or at all?) merge this directly into master. We'll either figure out how to do that merge, or just close this PR and recreate it in a new one. That'll make it hard to find this code in the future, so I hope to be able to merge it somehow.


xavdid:
The main reason I didn't want to collect OS in anonymous mode was that if there are users with more unique systems, then they'd be more identifiable (fewer people in that bucket).

The thing I was thinking about changing was if they're anonymous, create a deterministic (but non-identifiable) identifier for them. As it stands now, all anonymous users won't count towards a user count. It would be cool to still know how many different users are using the CLI on a weekly basis, even if we don't know who they are.

My guess is that most people won't use anon mode and we're fine as is, but we'll re-evaluate once this is out.

xavdid:
Ok! So the server is deployed and I've confirmed everything is working nicely. I had a lot of meetings today and didn't want to cut a release at the end of my day. 

If you could checkout this branch locally and deploy all 3 packages as `8.4.0` that would be awesome.

I'll fix merge conflicts tomorrow and then we should be back to a good state.

Thanks a ton!

PR #84, opened 2019-10-17
fix(cli): don't run analytics during tests

xavdid:
I noticed a big surge of analytics on `linux` with odd behavior, such as no command (which should never be possible). Turns out the culprit was traivs. Fixed some minor bugs while I was in there.

### DIFF
--- a/packages/cli/src/constants.js
+++ b/packages/cli/src/constants.js
@@ -45,26 +45,37 @@ const UPDATE_NOTIFICATION_INTERVAL = 1000 * 60 * 60 * 24 * 7; // one week
 const CHECK_REF_DOC_LINK =
   'https://platform.zapier.com/docs/integration-checks-reference';
 
+// can't just read from argv because they could have lots of extra data, such as
+// [ '/Users/david/.nvm/versions/node/v10.13.0/bin/node',
+//   '/Users/david/projects/zapier/platform/node_modules/.bin/mocha',
+//   'src/tests' ]
+const argvStr = process.argv.join(' ');
+const IS_TESTING =
+  argvStr.includes('mocha') ||
+  argvStr.includes('jest') ||
+  (process.env.NODE_ENV || '').toLowerCase().startsWith('test');
+
 module.exports = {
   ANALYTICS_KEY,
   ANALYTICS_MODES,
   API_PATH,
   AUTH_KEY,
-  AUTH_LOCATION,
   AUTH_LOCATION_RAW,
+  AUTH_LOCATION,
   BASE_ENDPOINT,
+  BLACKLISTED_PATHS,
   BUILD_DIR,
   BUILD_PATH,
   CHECK_REF_DOC_LINK,
-  SOURCE_PATH,
-  BLACKLISTED_PATHS,
   CURRENT_APP_FILE,
   DEFINITION_PATH,
   ENDPOINT,
+  IS_TESTING,
   LAMBDA_VERSION,
   PACKAGE_NAME,
   PACKAGE_VERSION,
   PLATFORM_PACKAGE,
+  SOURCE_PATH,
   STARTER_REPO,
   UPDATE_NOTIFICATION_INTERVAL
 };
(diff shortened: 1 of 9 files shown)


packages/cli/src/constants.js
```diff
@@ -45,26 +45,37 @@ const UPDATE_NOTIFICATION_INTERVAL = 1000 * 60 * 60 * 24 * 7; // one week
...
+  argvStr.includes('mocha') ||
+  argvStr.includes('jest') ||
+  (process.env.NODE_ENV || '').toLowerCase().startsWith('test');
+
 module.exports = {
   ANALYTICS_KEY,
   ANALYTICS_MODES,
   API_PATH,
   AUTH_KEY,
-  AUTH_LOCATION,
```
xavdid: these are just sorting

packages/cli/src/constants.js
```diff
@@ -45,26 +45,37 @@ const UPDATE_NOTIFICATION_INTERVAL = 1000 * 60 * 60 * 24 * 7; // one week
 const CHECK_REF_DOC_LINK =
   'https://platform.zapier.com/docs/integration-checks-reference';
 
+// can't just read from argv because they could have lots of extra data, such as
+// [ '/Users/david/.nvm/versions/node/v10.13.0/bin/node',
+//   '/Users/david/projects/zapier/platform/node_modules/.bin/mocha',
+//   'src/tests' ]
+const argvStr = process.argv.join(' ');
+const IS_TESTING =
```
xavdid: we have the same thing in core, helps us to be consistent

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -225,6 +225,14 @@ class ZapierBaseCommand extends Command {
       .join('\n')
       .trim();
   }
+
+  _recordAnalytics() {
```
xavdid: this could be overkill, but I did this before finding the bug in `analytics.js` itself. In any case, it's nice to centralize the behavior.

packages/cli/src/tests/testing.js
```diff
@@ -0,0 +1,9 @@
+require('should');
+
+const { IS_TESTING } = require('../constants');
+
+describe('testing setup', () => {
+  it('should set IS_TESTING to true', () => {
+    IS_TESTING.should.be.true();
```
xavdid: simple canary

packages/cli/src/tests/utils/analytics.js
```diff
@@ -0,0 +1,9 @@
+require('should');
+const { shouldSkipAnalytics } = require('../../utils/analytics');
+
+describe('analytics', () => {
+  // causes a lot of noise
+  it('should not run analytics when testing', () => {
+    shouldSkipAnalytics().should.be.true();
```
xavdid: this makes the functionality tough to test. it fixes the problem for now, but we'll probably have to rethink this approach.

packages/cli/src/utils/analytics.js
```diff
@@ -15,19 +15,22 @@ const setAnalyticsMode = newMode => {
   return writeUserConfig({ [ANALYTICS_KEY]: newMode });
 };
 
+const shouldSkipAnalytics = mode =>
```
xavdid: easier to read this way

packages/cli/src/utils/analytics.js
```diff
@@ -15,19 +15,22 @@ const setAnalyticsMode = newMode => {
...
+  IS_TESTING ||
+  process.env.DISABLE_ZAPIER_ANALYTICS ||
+  mode === ANALYTICS_MODES.disabled;
+
 const recordAnalytics = async (command, isValidCommand, args, flags) => {
   const analyticsMode = await currentAnalyticsMode();
 
-  const shouldRecordAnalytics =
-    process.env.DISABLE_ZAPIER_ANALYTICS ||
-    (process.NODE_ENV !== 'test' && analyticsMode !== ANALYTICS_MODES.disabled);
```
xavdid: ❗️ the actual bug was having `process.NODE_ENV` instead of `process.env.NODE_ENV`

PR #86, opened 2019-10-18
test(core): fix for analytics in tests & cleanup

xavdid:
Minor test fixes to make it easier to run smoke tests locally.

Also disables analytics from running in the tests, but doesn't actually work because of the bug fixed in #84 (that isn't deployed yet)

### DIFF
--- a/packages/core/smoke-test/smoke-test.js
+++ b/packages/core/smoke-test/smoke-test.js
@@ -25,16 +25,17 @@ const REGEX_VERSION = /\d+\.\d+\.\d+/;
 
 const setupZapierRC = () => {
   let hasRC = false;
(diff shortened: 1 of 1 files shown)


packages/core/smoke-test/smoke-test.js
```diff
@@ -90,7 +91,7 @@ const downloadRepoZip = async workdir => {
   const zipPath = path.join(workdir, 'repo.zip');
   const dest = fs.createWriteStream(zipPath);
 
-  return await new Promise((resolve, reject) => {
```
xavdid: Had a lint failure here - no need to `return await` since the result is already getting `await/.then`'d

PR #87, opened 2019-10-21
fix(core): hide sensitive data from error messages

eliangcs:
DO NOT MERGE! This PR is exactly the same as https://github.com/zapier/zapier-platform/pull/85. The branch is checked out from 8.4.0 and is just for us to release it in 8.4.1.


xavdid:
I was having trouble with the merge, so I re-committed the pertinent bits of #84 in 1d56b26. 

Given that both changes are already in master, I think we can safely release this as `8.4.1` and then close the branch and never merge it (like you mentioned).

PR #91, opened 2019-10-29
fix(schema) be a little more permissive with mutually exclusive fields

xavdid:
As part of https://github.com/zapier/zapier/pull/32518, I noticed that the frontend could get into a bad state where it was sending `{dict: false, list: false}` and schema was complaining that those were mutually exclusive. 

We can be a little more permissive here- it's only bad if they're truthy. My one unknown is that I'm not sure how the editor reads these values. If it also uses `_.has`, then this will cause weird behavior. @stevelikesmusic that's what I wanted to loop you in on if you could double check.

### DIFF
--- a/packages/schema/lib/constants.js
+++ b/packages/schema/lib/constants.js
@@ -16,5 +16,6 @@ module.exports = {
     ['dict', 'list'], // Use only one or the other
     ['dynamic', 'dict'], // dict is ignored
     ['dynamic', 'choices'] // choices are ignored
-  ]
+  ],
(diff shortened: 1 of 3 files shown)


xavdid:
@stevelikesmusic the was that if you check "list" and then "uncheck it", the state goes `{}` -> `{list: true}` -> `{list: false}` (which makes sense). But, the way that this worked, it was throwing an error.

To be clear, that's only using code from the above linked PR, where we validate VB apps in lambda. so no issue now, but I wanted to get ahead of this one.

PR #92, opened 2019-10-29
fix(legacy-scripting-runner): clean headers like legacy WB

eliangcs:
legacy-scripting-runner should clean headers just as Web Builder [does it](https://github.com/zapier/zapier/blob/0b95996f00238b3e092547a1478e4dc23972c5de/web/backend/zapier/useful/request_extras.py#L505).


eliangcs:
@ter0 Yes, there was a conversion error. I believe it happens because users can accidentally put whitespace chars (spaces, tabs and newlines) into their credentials.

PR #95, opened 2019-10-31
[NO MERGE] Bump deps fork

xavdid:
The only actual new change here was to do a find-and-replace on the lodash dep. 

This is forked from the 1d56b26ff1cefcadbbf7619300a0d95691a1d68e commit, since the one where `8.4.0` was actually bumped to `8.4.1` isn't in the tree. 

🚨 So, all the packages need to be moved to `8.4.2`, which isn't what'll happen when we run our regular `version` command.

This has all the changes that have already been deployed (which is good). 

I think we're safe to deploy off of this and then close the branch.


eliangcs:
8.4.2 has been published. Closing.

PR #96, opened 2019-11-01
chore: rebuild yarn.lock

eliangcs:
I found in our local dev, `zapier-platform/node_modules/lodash` was always 4.17.11 even though we bumped it to 4.17.15. I had to delete `yarn.lock` and rebuild a new one for yarn to correctly resolve Lodash to 4.17.15.


eliangcs:
Pretty sure this is good to go. I'm going ahead and merging this preemptively.

PR #99, opened 2019-11-14
BREAKING CHANGE(cli) Port build and upload

xavdid:
Port `build`, `upload`, and `push`.

Breaking - remove `--include-js-map` from `build`.

One thing I was looking at was how it's harder to ensure that a user is logged in. We can probably do the same setup we had before, but if it's a function we're doing a lot, then it would be cool to have `this.ensureValidSession()` to use in any command. 

Past that, we verify the session & app id more often than we probably need to. Not something to fix here, but that's a candidate for a little speed boost. Assuming the server gives reasonable errors, then we can let that handle a little more work here.


packages/cli/src/oclif/commands/build.js
```diff
@@ -0,0 +1,55 @@
...
+const { buildAndOrUpload } = require('../../utils/build');
+
+class BuildCommand extends BaseCommand {
+  async perform() {
+    this.log('Building Project\n');
+    await buildAndOrUpload(
+      { build: true },
+      {
+        skipNpmInstall: this.flags['skip-npm-install'],
+        disableDependencyInjection: this.flags['disable-dependency-injection']
```
eliangcs: This was named `disable-dependency-detection`. Was it renamed on purpose? I think the original name is more accurate.

packages/cli/src/tests/utils/build.js
```diff
@@ -148,6 +148,58 @@ describe('build', () => {
       });
   });
 
+  it('should make a build.zip without .zapierapprc', () => {
```
eliangcs: Can you explain why this test is needed?

packages/cli/src/oclif/commands/build.js
xavdid: whoops, definitely an accident.

packages/cli/src/tests/utils/build.js
xavdid: I changed the build command to no longer rely on there being a `.zapierapprc`. Users probably won't notice, but I ran into it a lot when testing build sizes - it's nice not to have to register the app (or use a fake file). 

I later found a bug where build would fail without an `.zapierapprc`, so I fixed it and added a test.

PR #104, opened 2019-11-15
BREAKING CHANGE(cli) port env command

xavdid:
A rewrite of the UX in the `env` command. Now it's called with the following:

* `env:get 1.0.0`
* `env:set 1.0.0 a=1 b=2`
* `env:unset 1.0.0 a b`

It looks like you currently cant set var args and strict parsing at the same time, so we're doing it custom a little bit. This could be made a little more dry, but it's ok for now. 

Also to note - we could also set it up to use only flags, which we could do in strict mode:

`env:set --version 1.0.0 --kvpair a=1 --kvpair b=2` or something. That felt a little more verbose, so I didn't go for it.


packages/cli/docs/cli.md
```diff
@@ -267,49 +267,19 @@ $ zapier describe
...
 ## env
 
-  > Read, write, and delete environment variables.
+> Gets Environment variables for a version.
 
-  **Usage:** `zapier env 1.0.0 CLIENT_SECRET 12345`
-
-  
-Manage the environment of your app so that `process.env` has the necessary variables, making it easy to match a local environment with a production environment via `CLIENT_SECRET=12345 zapier test`.
+**Usage**: `zapier env VERSION`
```
eliangcs: The doc doesn't seem to fully reflect the command. Maybe change it to:

```
> Gets, sets, or unsets environment variables for a version.

**Usage**: `zapier env:COMMAND VERSION`
```

Would be nice if we have at least one example for `env:get`, `env:set`, and `env:unset`, each.

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -18,6 +18,7 @@ class ZapierBaseCommand extends Command {
       require('debug').enable('zapier:*'); // enables all further spawned functions, like API
     }
 
+    this.debug('arv is', this.argv);
```
eliangcs: Typo? Should be `argv`.

packages/cli/src/oclif/commands/env/set.js
```diff
@@ -0,0 +1,86 @@
...
+
+class SetEnvCommand extends BaseCommand {
+  async perform() {
+    const { version } = this.args;
+    this.throwForInvalidVersion(version);
+    // args should be [ '1.0.0', 'qer=123', 'qwer=123' ]
+    const valuesToSet = this.argv
+      .slice(1)
+      .filter(k => !k.startsWith('-'))
+      .map(k => k.toUpperCase());
```
eliangcs: This also transforms the value into upper cases:

```
$ zapier env:set 1.0.0 hello=world
Successfully wrote the following to the environment of version 1.0.0:
{
  "HELLO": "WORLD"
}
```

We should probably delete the `toUpperCase` line here because it looks like you uppercase the environment key in the later code at line 37.

packages/cli/src/oclif/commands/env/unset.js
```diff
@@ -0,0 +1,71 @@
...
+      .filter(k => !k.startsWith('-'))
+      .map(k => k.toUpperCase());
+
+    if (!keysToUnset.length) {
+      this.error('Must specify at least one key to unset (like `SOME_KEY`)');
+    }
+
+    if (keysToUnset.some(v => v.includes('='))) {
+      this.error(
+        'Do not specify values using the unset operation, only values'
```
eliangcs: I think you meant to write `only keys` at the end.

packages/cli/docs/cli.md
xavdid: good call! tweaked it so subcommands all show up properly in docs & help.

PR #105, opened 2019-11-18
docs(cli) deprecate apps command

xavdid:
Renames `apps` to `integrations`, keeps the old command, and adds a deprecation warning handler. 

There are double docs right now, since `entry.js` (which will eventually be removed once the oclif migration is complete) doesn't know about aliases.


packages/cli/src/oclif/commands/integrations.js
```diff
@@ -3,11 +3,16 @@ const { buildFlags } = require('../buildFlags');
 
 const { listApps } = require('../../utils/api');
 
-class AppsCommand extends BaseCommand {
+class Integrations extends BaseCommand {
```
eliangcs: Shouldn't we name it `IntegrationsCommand`?

packages/cli/src/oclif/commands/integrations.js
```diff
@@ -3,11 +3,16 @@ const { buildFlags } = require('../buildFlags');
...
   async perform() {
+    if (this.argv.includes('apps')) {
+      this.warn('The `apps` command is deprecated. Use `integrations` instead');
```
eliangcs: This also appears in `hooks/deprecated.js`. Do we still need it here?

packages/cli/src/oclif/commands/integrations.js
xavdid: yep, copy pasting 😁 doesn't really matter, but it's probably good. We could also name them all `Command`, since it's renamed at import

packages/cli/src/oclif/commands/integrations.js
xavdid: ah, it appears there because it doesn't work here. `this.id` and `this.args` always use the actual version, not the alias. Will remove.

PR #106, opened 2019-11-19
BREAKING CHANGE(cli): add user management commands, remove old ones

xavdid:
Ok! This was a long one.

Refactors the `zapier invite` command into:

* `users:add`
* `users:remove`
* `users:get`
* `users:links`

Refactors the `zapier collaborate` command into:

* `team:add`
* `team:remove`
* `team:get`

As is, the code is fairly repetitive. We're under a little bit of a deadline to get this out, so rather than spend too much time here drying it up, I wanted to get it out the door. In particular, we can probably clean up reused code between the `:add` commands and how we split api calls between subscribers and collaborators.

For the review, pay particular attention to:

* the help text, ensuring it's clear and not copy-pasted wrong
* the UX of each command. I tried to stay with args and flags instead of interactive, but on one command it was pretty unavoidable. I'm not too bothered though, I honestly really like the interactive experiences.

Also, it would be great to be able to call `this.exit()` without it looking like an error. The only way I could think to do that is to throw an error with a specific message that we handle at the top-level. I don't love using errors for control flow though. In any case, a simple `return` from our `perform()` is perfectly serviceable.


packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -133,14 +133,15 @@ class ZapierBaseCommand extends Command {
...
-    return ans;
+  prompt(question, opts = {}) {
+    return inquirer
+      .prompt({
+        type: 'string',
+        ...opts,
+        name: 'ans',
+        message: question
+      })
+      .then(({ ans }) => ans);
```
eliangcs: Any reasons why changing it to promise-style vs. async/await?

packages/cli/src/utils/api.js
```diff
@@ -249,18 +249,34 @@ const listApps = async () => {
...
+    const route = _.isFunction(endpoint)
+      ? endpoint(app)
+      : `/apps/${app.id}/${endpoint}`;
+
+    const results = await callAPI(route, {
+      // if a full url comes out of the function, we have to use that
+      url: route.startsWith('http') ? route : undefined
     });
+
+    const { objects, theRest } = results;
```
eliangcs:
```suggestion
    const { objects, ...theRest } = results;
```

Suggesting this because I was running into this error with `zapier users:links`:

```
$ zapier-dev users:links
✔ Loading links

You can invite users to all versions of your app using the following link:


undefined


You can invite users to specific app versions using the following links:

 ›   Error: Cannot convert undefined or null to object
 ›
```

packages/cli/src/oclif/ZapierBaseCommand.js
xavdid: it _probably_ doesn't matter, but it brought it more in line with the other prompts.

In theory this has a performance benefit because it returns a promise to the calling function instead of waiting here and then returning. In practice, it probably doesn't matter. I'll revert.

PR #109, opened 2019-11-21
BREAKING CHANGE(cli): port delete command

xavdid:
Simple refactor of the `delete` command. Split it into two simple subcommands.


packages/cli/docs/cli.md
```diff
@@ -136,30 +136,34 @@ $ zapier convert 1234 .
...
 
-  > Delete a version of your app (or the whole app) as long as it has no users/Zaps.
+> Deletes your integration (including all versions).
 
-  **Usage:** `zapier delete version 1.0.0`
+**Usage**: `zapier delete:integration`
 
-  
-A utility to allow deleting app versions that aren't used.
+This only works if there are no active users or Zaps on any version. If you only want to delete certain versions, use the zapier delete:version command instead. It's unlikely that you'll be able to run this on an app that you've pushed publicly, since there are usually still users.
```
eliangcs:
```suggestion
This only works if there are no active users or Zaps on any version. If you only want to delete certain versions, use the `zapier delete:version` command instead. It's unlikely that you'll be able to run this on an app that you've pushed publicly, since there are usually still users.
```

PR #111, opened 2019-11-21
fix(cli) Run validation during build against `/check` endpoint

xavdid:
We were still using `/style-check` for the api call that happens during the `build` validation step. We should combine these at some point, but for now it's ok.


xavdid:
This is a simple change, so I'll go ahead and merge so we can ship

xavdid:
i actually didn't end up shipping. Even though, I've already merged this, @eliangcs you can take a look and tweak in master if you'd like!

PR #112, opened 2019-11-22
Release 9.0.0

xavdid:
This does the following:

* edits the changelog
* bumps the `version` in all 3 `package.json` files
* adds a line to `versionStore.js`
* tweaks our handling of semver slightly to accommodate AWS not caring about exact versions anymore
* bumps the boilerplate zip build func

---

Still need to:

- [ ] manually add git tags (x3)
- [ ] push to master
- [ ] upload boilerplate
- [ ] modify the `PlatformVersion`


.travis.yml
```diff
@@ -17,7 +16,7 @@ deploy:
   script: ./scripts/publish.sh
   on:
     tags: true
-    node: '8.10'
+    node: '10.16.3'
```
xavdid: @eliangcs please verify that this is what we want

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
-  - '8.10'
-  - '10'
+  - '10.16.3' # lambda version as of 2019-11-21
```
xavdid: no joke, AWS release node 12 support this week. We should add support for that soon, but not today

packages/cli/src/utils/misc.js
```diff
@@ -74,11 +74,8 @@ const runCommand = (command, args, options) => {
   });
 };
 
-const isValidNodeVersion = () => {
-  const v = semver(LAMBDA_VERSION);
-
-  return semver.satisfies(process.version, `>=${v.major}.${v.minor}`);
-};
+const isValidNodeVersion = (version = process.version) =>
+  semver.satisfies(version, `>=${semver.coerce(LAMBDA_VERSION)}`);
```
xavdid: `.coerce` plays nicely with `10.x` or whatever we need to put in as the lamba version. didn't end up needing it, but it's a good thing to use anyway.

packages/cli/src/version-store.js
```diff
@@ -11,5 +10,6 @@ module.exports = [
   { nodeVersion: '6.10.3', npmVersion: '>=3.10.10' }, // 5.x // patch version change on AWS
   { nodeVersion: '6.10.3', npmVersion: '>=3.10.10' }, // 6.x
   { nodeVersion: '8.10.0', npmVersion: '>=5.6.0' }, // 7.x
-  { nodeVersion: '8.10.0', npmVersion: '>=5.6.0' } // 8.x
+  { nodeVersion: '8.10.0', npmVersion: '>=5.6.0' }, // 8.x
+  { nodeVersion: '10', npmVersion: '>=5.6.0' } // 9.x; it's no longer a specific version
```
xavdid: Officially, lamba says they run it on `10.x`, so I figure we don't need such a hard requirement anymore. ¯\\\_(ツ)_/¯

packages/cli/src/constants.js
```diff
@@ -24,9 +23,7 @@ const BLACKLISTED_PATHS = [
   '.environment',
   'build'
 ];
-const nodeVersion = semver.Comparator(
```
xavdid: I have no idea why we needed to do this just to get out the exact same string we were putting in 🤔

xavdid:
tests are failing because the example apps can't validate because the server doesn't know about 9.0.0 yet, merging.

PR #114, opened 2019-11-22
 docs(cli) fix double `App.App.` in unit test snippet

kaznovac:


xavdid:
Ah, good catch. thanks!

PR #121, opened 2019-12-12
refactor(cli) port logs command

xavdid:
User-invisible change that updates the log command.


docs/cli.md
```diff
@@ -426,53 +426,21 @@ $ zapier link
...
-# │     Status      │ 200                                  │
-# │     URL         │ https://httpbin.org/get               │
-# │     Querystring │ hello=world                          │
-# │     Version     │ 1.0.0                                │
-# │     Step        │ 99c16565-1547-4b16-bcb5-45189d9d8afa │
-# │     Timestamp   │ 2016-01-01T23:04:36-05:00            │
-# └─────────────────┴──────────────────────────────────────┘
-```
+**Flags**
+* `-v, --version` | Filter logs to the specified version.
```
xavdid: this luckily doesn't conflict with `zapier --version` 😌

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -104,7 +104,7 @@ class ZapierBaseCommand extends Command {
    * @param {any[]} opts.rows The data to display
    * @param {string[][]} opts.headers Array of pairs of the column header and the key in the row that that header applies to
    * @param {string} opts.emptyMessage a message to print if there's no data. Printed in grey
-   * @param {boolean} opts.usedRowBasedTable override format and use `row` instead
+   * @param {boolean} opts.formatOverride override format and use this instead
```
xavdid: doc-only change

packages/cli/src/oclif/commands/logs.js
```diff
@@ -0,0 +1,119 @@
...
+      'Cap the number of logs returned. Max is 50 (also the default)',
+    default: 50
+  })
+};
+
+class LogsCommand extends BaseCommand {
+  async perform() {
+    this.startSpinner('Loading logs');
+
+    const flags = pick(this.flags, Object.keys(commandFlags));
```
xavdid: I want to pick here because the server rejects extra args. Currently it accepts but ignores `debug` and `format`, but it's better to not rely on that always being the case.

packages/cli/src/oclif/commands/logs.js
```diff
@@ -0,0 +1,119 @@
...
+        ['Timestamp', 'timestamp']
+      ];
+
+      if (this.flags.detailed) {
+        headers.concat([
+          ['Request Headers', 'request_headers'],
+          ['Request Body', 'request_data'],
+          ['Response Headers', 'response_headers'],
+          ['Response Body', 'response_content']
+        ]);
```
eliangcs: When I run `zapier logs --type http --detailed`, details are not showing. The problem is `Array.concat` doesn't do an in-place update. Should change it to:

```
headers = headers.concat([ ...
```

packages/cli/src/oclif/commands/logs.js
```diff
@@ -0,0 +1,119 @@
...
+      headers,
+      emptyMessage:
+        'No logs found. Try adding some `z.request()`, `z.console.log()` and doing a `zapier push`!\n'
+    });
+
+    if (hasLogs) {
+      this.log(grey('  Most recent logs near the bottom.'));
+
+      if (this.flags.type === 'http' && !this.flags.detailed) {
+        grey(
```
eliangcs: The message isn't showing because we should be calling `this.log` here.

packages/cli/docs/cli.md
```diff
@@ -426,53 +426,21 @@ $ zapier link
...
-# └─────────────────┴──────────────────────────────────────┘
-```
+**Flags**
+* `-v, --version` | Filter logs to the specified version.
+* `-s, --status` | Filter logs to only see errors or successes One of `[any | success | error]`. Defaults to `any`.
+* `-t, --type` | See logs of the specified type One of `[console | bundle | http]`. Defaults to `console`.
+* `--detailed` | See extra info, like request/response body and headers.
+* `-u, --user` | Only show logs for this user. Defaults to your account.  Defaults to `me`.
+* `--limit` | Cap the number of logs returned. Max is 50 (also the default)  Defaults to `50`.
+* `-f, --format` | undefined One of `[plain | json | raw | row | table]`. Defaults to `table`.
```
eliangcs: I wonder why there's an `undefined` here?

packages/cli/src/oclif/commands/logs.js
xavdid: Ah, good catch. TS would catch this too 😁

packages/cli/docs/cli.md
xavdid: ah thanks! little bug with our custom markdown.

PR #122, opened 2019-12-12
refactor(cli) Port Register Command

xavdid:
This fixes a long-standing frustration, which is calling `zapier register` with no args. Now title is optional and we'll ask at runtime if we need it. Everything else is the same.


packages/cli/src/utils/command-invocations.js
```diff
@@ -1,47 +0,0 @@
...
-// returns a promise
-const maybeRegisterApp = async context => {
-  if (!(await hasRegisteredApp())) {
-    context.line(
-      "Looks like this is your first push. Let's register your app on Zapier."
-    );
-    const title = await getInput('Enter app title (Ctrl-C to cancel):\n\n  ');
-    await register(context, title, { printWhenDone: false });
-  }
-};
```
eliangcs: I wonder why is this file removed? Originally we had use cases like calling `maybeRegisterApp` in `zapier push` if the app hasn't been registered. But it looks like we removed those. Maybe there's a reason but I forgot. Can you refresh my memory?

xavdid: good question! During the refactor we pulled that functionality out, so this file wasn't being used. With `registration` being ported, none of the commands this file references exist anymore.

As it is now, commands fail appropriately when auth is wrong or the app isn't registered. So it's not misleading, but not great.

I never loved this previous code approach, so I was going to think through a better way to it in oclif. But for now, we don't need this file anymore.

PR #124, opened 2019-12-13
fix(core) use NodeJS ReadableStream instead of DOM in TS definitions

kaznovac:
* use NodeJS.ReadableStream instead of ReadableStream from DOM
* set node tsconfig [recommended settings](https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping) for node 10.x in example app
* explicitly provide only es2018 lib in tsconfig (to avoid accidental importing of DOM modules)

### DIFF
--- a/example-apps/typescript/tsconfig.json
+++ b/example-apps/typescript/tsconfig.json
@@ -1,7 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "es2018",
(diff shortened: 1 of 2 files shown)


kaznovac:
this is node-fetch typings for reference about ReadableStream https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node-fetch/index.d.ts

PR #128, opened 2019-12-17
refactor(cli): port link command

xavdid:
We got to simplify this command a lot!

The only downgrade is that the table currently shows more info like "date" and "id". Could maybe add `id` back into the label in the chooser.

I don't think this counts as a breaking change since it was interactive before and now it's a different UI, but the same amount of interactive.


xavdid:
Cool, i'll change the title to `AppName (1234)`. We use [InquirerJS](https://github.com/SBoudrias/Inquirer.js/) under the hood, so we're a little bit limited with what we can customize. That said, item order and page size are both customizable. I'll knock the page size up to... 15? 20?

As for the sort order, the old version did it by `id ASC`. Do we want to keep that or do it alphabetical instead?

Lastly, the old command had a ✔️on the currently linked app (there's a `.zapierapprc`). That didn't feel important to me, so I pulled it. We can re add it in the title though. It would be something like `AppName (1234) [currently linked]`. Thoughts?

eliangcs:
> I'll knock the page size up to... 15? 20?

Yeah, increasing the page size would help. I think 15 should be good.

> Do we want to keep that or do it alphabetical instead?

I would vote for alphabetical order.

> It would be something like AppName (1234) [currently linked]. Thoughts?

I don't feel it's important either, but that would be a good improvement.

xavdid:
Done! Great suggestions.

PR #130, opened 2019-12-17
fix(core): don't require both username and password for basic auth

eliangcs:
In Web Builder, we allow username or password to be empty. For example, a developer can use an auth mapping like the following to do basic-auth-backed API key auth:

```
{
  "username": "{{api_key}}",
  "password": ""
}
```

In CLI core, however, we currently require both username and password to be non-empty for the basic auth middleware to add the auth header to the request. This causes trouble if we convert a WB app like that to CLI. Not sure why we did that in the first place, but I think it should be good if we allow either username or password to be empty.

### DIFF
--- a/packages/core/src/http-middlewares/before/add-basic-auth-header.js
+++ b/packages/core/src/http-middlewares/before/add-basic-auth-header.js
@@ -2,11 +2,14 @@
 
 // Computes the basic auth header for the request
 const addBasicAuthHeader = (req, z, bundle) => {
-  if (bundle.authData && bundle.authData.username && bundle.authData.password) {
-    const buff = Buffer.from(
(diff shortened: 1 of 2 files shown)


eliangcs:
> this is technically a breaking change.

I thought about that too. To avoid that, we could make this change in legacy-scripting-runner by adding a don't-care character like `x` if username or password is empty. So when it comes to core middleware, the check `bundle.authData.username && bundle.authData.password` would pass.

However, I think it's totally acceptable for basic auth to have only one of username/password, so I made the change in core. It should be a safe assumption that most of the servers don't care if we add the basic auth header while previously we didn't.

xavdid:
Cool, I thought about this a little more and I think it's fine. Since it only affects apps that already have `basic` as the auth type, impact should be limited.

PR #132, opened 2019-12-18
style(cli): make language of validate command consistent with UI

eliangcs:
Addresses [PDE-1251](https://zapierorg.atlassian.net/browse/PDE-1251).

Before:

After:

### DIFF
--- a/packages/cli/src/oclif/ZapierBaseCommand.js
+++ b/packages/cli/src/oclif/ZapierBaseCommand.js
@@ -133,8 +133,8 @@ class ZapierBaseCommand extends Command {
    * Print text in a list style.
    * @param {string[][]} items
    */
-  logList(items, { maxWidth = stdtermwidth } = {}) {
-    this.log(renderList(items, { spacer: '\n', maxWidth }));
+  logList(items) {
+    this.log(renderList(items, { spacer: '\n', maxWidth: stdtermwidth }));
   }
 
   /**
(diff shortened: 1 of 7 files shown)


packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -127,6 +129,14 @@ class ZapierBaseCommand extends Command {
     }
   }
 
+  /**
+   * Print text in a list style.
+   * @param {string[][]} items
+   */
+  logList(items, { maxWidth = stdtermwidth } = {}) {
```
xavdid: I like having the `maxWidth` param here, but is there a particular reason you added it? I don't see it used anywhere.

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -1,4 +1,6 @@
 const { Command } = require('@oclif/command');
+const { stdtermwidth } = require('@oclif/plugin-help/lib/screen');
```
xavdid: This should be fine, but my understanding is that if they move or change their file structure at all, this import would break. Just a thing to be wary of.

eliangcs: If the import breaks, the smoke test will break too, so I'm not too worried.

packages/cli/src/oclif/ZapierBaseCommand.js
eliangcs: Yeah, it's not used and probably won't. Removed it in fc9fbeb.

PR #137, opened 2019-12-27
style(cli): consistent CLI style

eliangcs:
Addresses [PDE-1271](https://zapierorg.atlassian.net/browse/PDE-1271) with some differences. The goal is to make sure the CLI uses a consistent writing style and format across all the commands. Summary:

* Make sure the first line of the help text of a command starts with a verb and ends with a period, such as:
  - `Get environment variables for a version.`
  - `Migrate users from one version of your integration to another.`
* Format the Markdown help text appropriately in the terminal. This is done using [marked-terminal](https://github.com/mikaelbr/marked-terminal) with an oclif hook `src/oclif/hooks/renderMarkdownHelp.js`.
* Replace all occurrences of "app" with "integration".
* Update some of the command help text to make it more reflect the current reality.


packages/cli/package.json
```diff
@@ -89,15 +91,22 @@
...
       "delete": {
-        "description": "Delete your entire integration or a specific integration version"
+        "description": "Delete your entire integration or a specific integration version."
       },
       "env": {
-        "description": "Interact with your integration's environment"
+        "description": "Interact with your integration's environment."
+      },
+      "team": {
+        "description": "Add, remove, get team members of your integration."
```
xavdid: Might change to `"Add, remove, and get team members`

packages/cli/src/oclif/commands/convert.js
```diff
@@ -118,7 +118,7 @@ ConvertCommand.flags = buildFlags({
     })
   }
 });
-ConvertCommand.description = `Converts a Legacy Web Builder or Visual Builder app to a CLI app.
+ConvertCommand.description = `Convert a Legacy Web Builder or Visual Builder integration to a CLI integration.
```
xavdid: Might swap to `Convert a Legacy Web Builder app or Visual Builder integration to a CLI integration`, since "web builder app" is a known phrase.

packages/cli/src/oclif/commands/users/add.js
```diff
@@ -50,11 +50,10 @@ UsersAddCommand.flags = buildFlags({
     })
   }
 });
+UsersAddCommand.examples = ['zapier users:add bruce@wayne.com'];
```
xavdid: might include an example with a version, too: `zapier users:add alfred@wayne.com 1.2.3`

packages/cli/src/utils/api.js
```diff
@@ -235,7 +235,7 @@ const getVersionInfo = () => {
   });
 };
 
-const checkCredentials = () => {
+const checkCredentials = async () => {
```
xavdid: this doesn't need to be async since we don't use `await` in the function. We're already returning a promise, so `async` doesn't get us anything extra. There's some related reading [here](https://eslint.org/docs/rules/no-return-await). 

Honestly, this function could be re-written as `const checkCredentials = () => callAPI('/check')`

packages/cli/package.json
```diff
@@ -53,6 +53,8 @@
     "inquirer": "6.5.0",
     "klaw": "3.0.0",
     "lodash": "4.17.15",
+    "marked": "0.8.0",
```
xavdid: When running `yarn`, I get `warning "workspace-aggregator-... > zapier-platform-cli > marked-terminal@3.3.0" has incorrect peer dependency "marked@^0.4.0 || ^0.5.0 || ^0.6.0 || ^0.7.0".` I'd probably use `marked@0.7.0` unless there's a big reason to use 0.8.0.

packages/cli/src/oclif/hooks/renderMarkdownHelp.js
```diff
@@ -0,0 +1,24 @@
...
+    codespan: chalk.underline.bold
+  })
+});
+
+module.exports = options => {
+  const cmdId = options.id === 'help' ? options.argv[0] : options.id;
+  const cmd = options.config.findCommand(cmdId);
+  if (cmd) {
+    if (cmd.description) {
+      cmd.description = marked(cmd.description).trim();
```
xavdid: this feels like an odd way to do this, but it works. ¯\\\_(ツ)_/¯

packages/cli/package.json
eliangcs: Sure, but I'd use "or" instead of "and".

PR #146, opened 2020-01-09
feat(cli) Scaffold improvements

xavdid:
I still have a little cleanup to do, but this is the lion's share of the work!

I had a little trouble testing the command itself. I think we'll need to do the same thing we do in smoke tests (make a temp dir and invoke commands there from the CLI, not from the code itself). I was going to use `mock-fs`, but then we need to replicate a ton of things into the mock (the templates, plus all of `oclif`, since it uses lazy requires). 

In absence of that, I tested all of the utils under the hood. Notably, the flags aren't well tested here. 

This PR got big (sorry!), so I'm limiting the scope a little. The next step will be to change `init --template` to use `scaffold` under the hood

### open tasks

- [x] be smarter about finding the export obj
- [x] handle spacing in noun
- [x] allow a non-checked install when being used from init
- [x] rollback logic?
- [x] spruce up other templates


packages/cli/scaffold/trigger.template.js
```diff
@@ -1,13 +1,13 @@
-// triggers on <%= LOWER_NOUN %> with a certain tag
-const trigger<%= CAMEL %> = (z, bundle) => {
-  const responsePromise = z.request({
+// triggers on a new <%= LOWER_NOUN %> with a certain tag
+const perform = async (z, bundle) => {
```
xavdid: This is the updated version of the template; I still need to spruce up the others. They'll match the improvements in this one though.

packages/cli/src/oclif/commands/scaffold.js
```diff
@@ -1,181 +1,116 @@
...
+  createTemplateContext
+} = require('../../utils/scaffold');
+
+const getNewFileDirectory = (action, test = false) =>
+  `${test ? 'test/' : ''}${plural(action)}`;
+
+/**
+ * both the string to `require` and later, the filepath to write to
+ */
+const getFilename = (directory, noun) =>
```
xavdid: I was having trouble being satisfied about where these tiny util functions should live. They're here, but they're a little redundant and I don't love them

packages/cli/src/oclif/commands/scaffold.js
```diff
@@ -1,181 +1,116 @@
...
 
 ScaffoldCommand.args = [
   {
-    name: 'type',
-    help: 'What type of thing are you creating',
+    name: 'actionType',
+    help: 'What type of step type are you creating?',
     required: true,
-    options: ['resource', 'trigger', 'search', 'create']
+    options: ['trigger', 'search', 'create', 'resource']
```
xavdid: trying to deprioritize resources as a thing

packages/cli/src/oclif/commands/scaffold.js
```diff
@@ -1,181 +1,116 @@
...
-    }
+    this.log(`\nFinished! Your new ${actionType} is ready to use.`);
   }
 }
 
 ScaffoldCommand.args = [
   {
-    name: 'type',
-    help: 'What type of thing are you creating',
+    name: 'actionType',
```
xavdid: this is non-breaking since this just describes the variable we access this at in the code

packages/cli/src/tests/utils/build.js
```diff
@@ -1,26 +1,21 @@
...
+describe('build (runs slowly)', () => {
   let tmpDir, entryPoint;
   before(async () => {
     // basically does what `zapier init` does
-    const osTmpDir = fse.realpathSync(os.tmpdir());
-    tmpDir = path.join(
-      osTmpDir,
-      'zapier-' + crypto.randomBytes(4).toString('hex')
-    );
+    tmpDir = getNewTempDirPath();
```
xavdid: I have more similar changes to make here

packages/cli/src/utils/ast.js
```diff
@@ -0,0 +1,108 @@
...
+        { init: { type: 'CallExpression', callee: { name: 'require' } } }
+      ]
+    })
+    // filters for top-level require statements by filtering only for statements whose parents are type Program, the root
+    .filter(path => j.Program.check(path.parent.value));
+
+  if (reqStatements.length) {
+    reqStatements
+      .at(-1)
+      .insertAfter(
```
xavdid: I'm going to DRY this up

packages/cli/src/utils/ast.js
```diff
@@ -0,0 +1,108 @@
...
+              ),
+              value: j.identifier(varName),
+              computed: true
+            })
+          ])
+        )
+      );
+    } else {
+      throw new Error(
+        `Unable to add new property "${property}" to exported app. Can't find variable declaration for "App". To fix, either ensure there is a variable called "App" in your root file or add \`${property}: {},\` to your existing app. Then, re-run this command.\``
```
xavdid: You'll notice this code acts on a variable explicitly called `App` despite trying to move away from that terminology. All of our templates thusfar have used that and looking for that explicitly gives us a better shot at being able to find the right variable declaration. 

I don't anticipate this error being thrown much (since our templates have also always included `triggers: {}` by default) but pushing `App` here could be confusing. 

Do you think we should:

* check both properties, but have the error instruct users to add a variable called `Integration`
* just have folks `module.exports = { triggers: {} }` since the variable here is sort of redundant anyway
* something else?

packages/cli/src/utils/scaffold.js
```diff
@@ -0,0 +1,109 @@
...
+
+  try {
+    codeStr = createRootRequire(codeStr, varName, `./${relativePath}`);
+    codeStr = addKeyToPropertyOnApp(codeStr, plural(actionType), varName);
+    await writeFile(entryFilePath, codeStr);
+
+    // validate the edit happened correctly
+    // can't think of why it wouldn't, but it doesn't hurt to double check
+    const rewrittenIndex = require(entryFilePath);
+    if (!_.get(rewrittenIndex, [plural(actionType), newActionKey])) {
```
eliangcs: When I test, I always see the "Oops, we could not reliably rewrite" message even if my `index.js` seems to be correctly edited. I think it's because `require(entryFilePath)` is cached, which would always give us the unmodified file content. To get around the cache, we should probably use `readFile` or find a way to force `require` to reload.

packages/cli/src/utils/ast.js
eliangcs: I'd vote for the second option because it's more robust. `App` could be undefined. Try converting a VB app with `zapier convert` for example.

packages/cli/src/utils/scaffold.js
xavdid: interesting, outside of bugs mine has worked every time :thinking:



I can clear the require cache though, to make sure we're getting a fresh copy.

eliangcs: I re-tested and it wasn't the cache. It's because I haven't run `npm|yarn install` within the app directory, so when you `require(entryFilePath)`, it raises an error. But the entry was correctly inserted to `index.js`. I think it's because before the `require`, you only parse `index.js` syntactically, which doesn't require dependencies.

I wonder is there a way to double check the edit without requiring a developer run `npm/yarn install` beforehand? After all, when scaffolding, there are high chances the dependencies haven't been installed yet.

xavdid: I've been thinking about this. I think we can safely assume that when this is run via `init`, it will always succeed (no need to re-check). So we only need to worry about the case where a dev is adding it to an existing app. We could probably do logic that this depends on a valid app install (aka they've installed `node_modules`) if directly as a command. Either bail early or do the work and not check it. I sort of like the former, since that's how we can ensure it worked (and not leave code in a bad state).

packages/cli/src/oclif/commands/scaffold.js
```diff
@@ -1,181 +1,115 @@
...
+  plural,
+  updateEntryFile,
+  writeTemplateFile,
+  createTemplateContext
+} = require('../../utils/scaffold');
+
+const getNewFileDirectory = (action, test = false) =>
+  `${test ? 'test/' : ''}${plural(action)}`;
+
+const getLocalFilePath = (directory, noun) => `${directory}/${noun}`;
```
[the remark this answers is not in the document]
xavdid: Ah, fair point. I would argue that the noun should be `Contact` instead of `New Contact` (and this input's place in the template makes more sense as `contact`), but I realize:

* our examples don't always reflect that
* this should at least work as well as it did before

I'll fix that.

packages/cli/src/utils/scaffold.js
eliangcs: I like the former too: check if they've installed `node_modules` before we re-check.

packages/cli/src/utils/ast.js
xavdid: I re-wrote a lot of this logic and it's a lot more sturdy now!

packages/cli/scaffold/index.template.js
```diff
@@ -1,24 +0,0 @@
-<%= REQUIRES %>
-
-const App = {
```
xavdid: I couldn't find anywhere this was used, so I binned it

packages/cli/scaffold/resource.template.js
```diff
@@ -1,86 +1,88 @@
...
-    url: `https://jsonplaceholder.typicode.com/posts/${bundle.inputData.id}`,
-  });
-  return responsePromise
-    .then(response => z.JSON.parse(response.content));
-};
-
 // get a list of <%= LOWER_NOUN %>s
-const list<%= CAMEL %>s = (z) => {
-  const responsePromise = z.request({
+const performList = (z, bundle) => {
```
xavdid: this is a lot of repeated code, but I think it's fine for now. we'll have to do more passes on these later anyway when we condense everything

packages/cli/src/oclif/commands/convert.js
```diff
@@ -129,4 +129,6 @@ If you re-run this command on an existing directory it will leave existing files
 
 You'll need to do a \`zapier push\` before the new version is visible in the editor, but otherwise you're good to go.`;
 
+ConvertCommand.skipValidInstallCheck = true;
```
xavdid: these are moved so each class can control whether or not the check happens

packages/cli/src/oclif/commands/scaffold.js
```diff
@@ -1,181 +1,168 @@
...
+      actionType,
+      templateContext.KEY
+    );
 
-      const updatedEntry = await this.updateEntry(
-        entryFile,
-        entry,
-        templateContext.CAMEL,
-        dest
+    if (isValidAppInstall().valid) {
```
xavdid: `isValidAppInstall()` doesn't check the `skipValidInstallCheck` property, so this will run as expected. this way, we only `require` it if `node_modules` is present

packages/cli/src/oclif/commands/scaffold.js
```diff
@@ -1,181 +1,168 @@
...
+      const success = isValidEntryFileUpdate(
```
xavdid: this controls the rollback behavior. to see it, you can either modify your `index.json` to fail on one of the listed errors or change this to `!isValidEntryFileUpdate`

packages/cli/src/tests/utils/ast.js
```diff
@@ -0,0 +1,179 @@
...
+        // assertions about what comes in the trigger property
+        should(codeByLine.indexOf('[findThing.key]: findThing')).eql(
+          firstIndex + 1
+        );
+        should(codeByLine[firstIndex + 2]).eql('}');
+      });
+    });
+  });
+
+  describe('errors', () => {});
```
xavdid: I'll wrap this up tomorrow, should be straightforward. I'll have an index.js that specifically will be tripped over by each thrown error.

packages/cli/src/tests/utils/build.js
```diff
@@ -1,26 +1,20 @@
 const should = require('should');
 
-const crypto = require('crypto');
```
xavdid: just refactoring, nothing to see here.

packages/cli/src/utils/ast.js
```diff
@@ -0,0 +1,139 @@
+// tools for modifiyng an AST
+
+const j = require('jscodeshift');
```
xavdid: this code is a little confusing, so let me know if the comments aren't clear enough. do you feel like you would be able to modify this later?

packages/cli/src/utils/scaffold.js
```diff
@@ -0,0 +1,120 @@
...
+  const originalCodeStr = codeStr; // untouched copy in case we need to bail
+  const relativePath = getRelativeRequirePath(entryFilePath, newFilePath);
+
+  codeStr = createRootRequire(codeStr, varName, `./${relativePath}`);
+  codeStr = addKeyToPropertyOnApp(codeStr, plural(actionType), varName);
+  await writeFile(entryFilePath, codeStr);
+  return originalCodeStr;
+};
+
+const isValidEntryFileUpdate = (entryFilePath, actionType, newActionKey) => {
```
xavdid: I separated the validation out from the modification, so we don't always have to do it

packages/cli/src/utils/ast.js
eliangcs: Yeah, some serious stuff here! Playing around in AST Explorer helps me understand it a bit. The comments help too. I guess I've got the idea, so 👍

packages/cli/src/utils/scaffold.js
```diff
@@ -0,0 +1,120 @@
...
+
+const { ensureDir, fileExistsSync, readFile, writeFile } = require('./files');
+const { splitFileFromPath } = require('./string');
+const { createRootRequire, addKeyToPropertyOnApp } = require('./ast');
+const { snakeCase } = require('./misc');
+
+const plural = type => (type === 'search' ? `${type}es` : `${type}s`);
+
+const getTemplatePath = actionType =>
+  path.join(__dirname, '../..', `scaffold/${actionType}.template.js`);
```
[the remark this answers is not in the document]
xavdid: good call! dang windows. 

I went with `path.join(__dirname, '..', '..', 'scaffold', actionType, '.template.js');`, which should be suitably cross-platform and a little less dreadful

xavdid:
Sweet, addressed your comments and added analytics as a no-login command. 🚢❗️

PR #148, opened 2020-01-24
tooling: custom bump script

eliangcs:
Part of [PDE-870](https://zapierorg.atlassian.net/browse/PDE-870) to fix our currently broken release process.

Originally, we used `lerna version`, which doesn't fit our needs entirely and isn't easy to customize. So I ended up manually editing `package.json`s and adding git tags a lot when releases packages.

This PR adds a standalone and interactive script `./scripts/bump.js` (can also be executed with `yarn bump`), which does the following things:

1. ask us which package(s) and the version(s) to bump
   
2. bump the selected packages by editing their `package.json`
3. update example apps' dependencies in `package.json`
4. git commit the changes
5. add the necessary git tags

Once the script finished, the only thing we need to do is to `git push origin master --tags`. When Travis CI sees the git tags on GitHub, it will publish the packages to npm for us.

### DIFF
--- a/scripts/bump.js
+++ b/scripts/bump.js
@@ -49,6 +49,46 @@ const ensureMainPackageVersionsAreSame = () => {
   }
 };
 
+const confirmIfNotMasterBranch = async () => {
+  const result = spawnSync('git', ['branch', '--show-current'], {
+    encoding: 'utf8'
+  });
+  const branch = (result.stdout || '').trim();
+  if (branch !== 'master') {
+    const answer = await inquirer.prompt([
+      {
+        type: 'confirm',
+        name: 'continue',
+        message:
+          'This was supposed to be run on master branch. ' +
+          `You want to proceed with branch ${underline(branch)} anyway?`,
+        default: false
+      }
+    ]);
+    if (!answer.continue) {
+      throw new Error('Canceled.');
+    }
+  }
+};
+
+const ensureNoUncommittedChanges = () => {
+  const result = spawnSync(
+    'git',
+    ['status', '--untracked-files=no', '--porcelain'],
+    { encoding: 'utf8' }
+  );
+  const lines = result.stdout
+    .split(/\r?\n/)
+    .map(line => line.trim())
+    .filter(line => line);
+  if (lines.length > 0) {
+    throw new Error(
+      `${bold.underline('git status')} shows you have uncommitted changes. ` +
+        'Commit or discard those before you try again.'
+    );
+  }
+};
+
 const promptPackagesToBump = async () => {
   const answer = await inquirer.prompt([
     {
@@ -237,6 +277,8 @@ const gitTag = versionsToBump => {
 const main = async () => {
   try {
     ensureMainPackageVersionsAreSame();
+    await confirmIfNotMasterBranch();
+    ensureNoUncommittedChanges();
   } catch (err) {
     console.error(err.message);
     return 1;
@@ -247,8 +289,8 @@ const main = async () => {
 
   const packagesToBump = await promptPackagesToBump();
   if (packagesToBump.length === 0) {
-    console.log('No packages selected, nothing to do here');
-    return 0;
+    console.error('No packages selected, nothing to do here.');
+    return 1;
   }
 
   const versionsToBump = {};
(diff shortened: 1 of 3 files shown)


scripts/bump.js
```diff
@@ -0,0 +1,284 @@
...
+
+const fs = require('fs');
+const path = require('path');
+
+const { bold, underline } = require('chalk');
+const { spawnSync } = require('child_process');
+const inquirer = require('inquirer');
+const semver = require('semver');
+
+const REPO_DIR = path.dirname(__dirname);
```
xavdid: the double `dirname` here reads a little funny. maybe use `process.cwd()`, since this command will always be run from the platform root?

scripts/bump.js
```diff
@@ -0,0 +1,284 @@
...
+  'legacy-scripting-runner'
+].reduce((result, packageName) => {
+  const packageJsonPath = path.join(
+    REPO_DIR,
+    'packages',
+    packageName,
+    'package.json'
+  );
+  const packageJson = readJson(packageJsonPath);
+  result[packageName] = packageJson.version;
```
xavdid: It's totally fine here (and is not worth changing) but it's worth noting (in case you weren't aware) that writing a reduce this way modifies the accumulator object. Since we're passing in `{}` it doesn't matter. If we were instead acting on an existing variable, like this: 

```js
const originalSettings = {a: 1, b: 2}
const newSettings = ['c', 'd', 'e'].reduce((result, someVar) => {
  result[someVar] = true; 
  return result
}, originalSettings)

console.log(newSettings, originalSettings)
// => {a: 1, b: 2, c: true, d: true, e: true}, {a: 1, b: 2, c: true, d: true, e: true}
```

Instead, you'd use the spread operator to return a new object each time:

```js
const originalSettings = {a: 1, b: 2}
const newSettings = ['c', 'd', 'e'].reduce((result, someVar) => {
  return {...result, [someVar]: true}
}, originalSettings)

console.log(newSettings, originalSettings)
// => {a: 1, b: 2, c: true, d: true, e: true}, {a: 1, b: 2}
```

like I said, totally reasonable here. But, the more you know!

scripts/bump.js
```diff
@@ -0,0 +1,284 @@
...
+  return result;
+}, {});
+
+const ensureMainPackageVersionsAreSame = () => {
+  if (
+    !(
+      PACKAGE_ORIG_VERSIONS.cli === PACKAGE_ORIG_VERSIONS.core &&
```
xavdid: it would be great to break this assumption eventually (I still can't think of a reason we need them to match up as long as all versions in play are a semi-ish-recent one), but this is fine here for now.

scripts/bump.js
```diff
@@ -0,0 +1,284 @@
...
+    gitAdd();
+    gitCommit(versionsToBump);
+    gitTag(versionsToBump);
+  } catch (err) {
+    console.error(err.message);
+    return 1;
+  }
+
+  console.log(
+    `\nDone! Now you can ${bold.underline('git push origin master --tags')}.`
```
xavdid: is this a "can" or a "should"? seems like a required step

scripts/bump.js
```diff
@@ -0,0 +1,284 @@
...
+  );
+
+  return 0;
+};
+
+main().then(exitCode => {
+  if (exitCode) {
+    process.exit(exitCode);
```
xavdid: if there's an error partway through, it might be nice to include instructions to roll back changes (in this case, probably discarding unstaged git changes) so the command can be safely run again

scripts/bump.js
eliangcs: It's actually intended so we can run it from any directory. Using `process.cwd()` makes an assumption where we run the script.

xavdid: that's true! do you anticipate running it without using `yarn bump` or from other directories often?

eliangcs: Probably not often. Sometimes I create a symlink to common-used command and put it under `PATH` so I can run it fast from anywhere. `bump.js` isn't going to be a common command, but I think it's a good practice for scripts not to assume the current working directory.

scripts/bump.js
eliangcs: I was aware of this but the spread operator is a good callout! Fixed in aed592d.

scripts/bump.js
eliangcs: Yeah, we'll need to revisit here when we want to make an effort to decouple cli and core versions.

scripts/bump.js
eliangcs: It'd be great to roll back automatically, but I'm taking a pass on that for now. Added a vague instruction in 1ca74d9.

PR #150, opened 2020-01-28
docs(cli): Add clarification for Computed Fields support

kreddlear:
The VB docs specify that Computed Fields only work with OAuth and Session Auth. I had a CLI user who didn't realize this, so I'm adding this to the CLI docs as well.

Please let me know if I should have changed this doc instead! https://github.com/zapier/visual-builder/blob/master/docs/_cli_docs/docs.md


xavdid:
@kreddlear ah, this is great! But, you need to edit `readme-source.md` and then build the docs. If you don't have yarn on your computer, i'd be happy to run it for you!

xavdid:
I added this but accidentally sent it straight to master (https://github.com/zapier/zapier-platform/commit/ee5176ba45af36ca6f481443f53d48999dace102). In any case, we're good. thanks!

PR #151, opened 2020-01-28
fix(legacy-scripting-runner): 'this' should be bound to 'Zap'

eliangcs:
Using `this.something` in a scripting method should equivalent to `Zap.something`.

### DIFF
--- a/packages/legacy-scripting-runner/test/integration-test.js
+++ b/packages/legacy-scripting-runner/test/integration-test.js
@@ -646,6 +646,25 @@ describe('Integration Test', () => {
       });
     });
 
+    it('KEY_pre_poll, this binding', () => {
+      const appDef = _.cloneDeep(appDefinition);
+      appDef.legacy.scriptingSource = appDef.legacy.scriptingSource.replace(
+        'movie_pre_poll_this_binding',
+        'movie_pre_poll'
+      );
+      const _compiledApp = schemaTools.prepareApp(appDef);
+      const _app = createApp(appDef);
+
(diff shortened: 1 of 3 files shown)


packages/legacy-scripting-runner/test/integration-test.js
```diff
@@ -646,6 +646,28 @@ describe('Integration Test', () => {
...
+      const _app = createApp(_appDefWithAuth);
+
+      const input = createTestInput(
+        _compiledApp,
+        'triggers.movie.operation.perform'
+      );
+      input.bundle.authData = { api_key: 'secret' };
+      return _app(input).then(output => {
+        const movies = output.results;
+        movies.length.should.greaterThan(1);
```
codebycaleb: From my understanding of the issue, this is the line that fails before the fix, right? Because `movies` is `undefined`? It's awesome to hear that a one-line fix can resolve that! Nice work! 🥇

eliangcs: The test would error when `this.getMovesUrl()` is invoked so that's before this line. The error message would be `this.getMoviesUrl is not a function`, something like [this](https://travis-ci.com/zapier/zapier-platform/builds/146512707#L451).

xavdid:
for my own edification, I tried running the new test (without the `.bind` call) on node `8.10` hoping that it would pass. It didn't, so now I have no idea why this only manifested after we moved those apps from 8 -> 10 ¯\\\_(ツ)_/¯

in any case, thanks for grabbing!

PR #152, opened 2020-01-29
tooling: fully automate boilerplate build process

eliangcs:
Last piece of [PDE-870](https://zapierorg.atlassian.net/browse/PDE-870).

Automates the boilerplate build process. See the below comments for details.


boilerplate/README.md
```diff
@@ -0,0 +1,5 @@
+# Zapier Platform Boilerplate
+
+Boilerplate is an empty Zapier integration project. We use it to build a zip
+providing all the necessary dependencies for integrations developed on visual
+builder.
```
eliangcs: The `boilerplate` has been moved from the `REPO_ROOT/packages/core/boilerplate` to its own directory `REPO_ROOT/boilerplate`. Putting `boilerplate` in `core` is awkward. It forces `core` to aware of the existence of `legacy-scripting-runner` because `boilerplate` depends on both `core` and `legacy-scripting-runner`.  It makes more sense for `boilerplate` to be isolated.

I thought about putting it to `REPO_ROOT/packages/boilerplate`. It wasn't an idea because `boilerplate` is not a yarn workspace. We don't want yarn to hoist dependencies to workspace root's `node_modules` when building the boilerplate zip.

boilerplate/scripts/build.sh
```diff
@@ -0,0 +1,136 @@
...
+    legacy_version=$3
+
+    cmd_for_core="s|\"zapier-platform-core\": \"[^\"]*\""
+    cmd_for_core+="|\"zapier-platform-core\": \"$core_version\"|g"
+
+    cmd_for_legacy="s|\"zapier-platform-legacy-scripting-runner\": \"[^\"]*\""
+    cmd_for_legacy+="|\"zapier-platform-legacy-scripting-runner\": \"$legacy_version\"|g"
+
+    sed -i.bak "$cmd_for_core" "$package_json"
+    sed -i.bak "$cmd_for_legacy" "$package_json"
```
eliangcs: Originally, we have a Node script [packages/core/bin/update-boilerplate-dependencies.js](https://github.com/zapier/zapier-platform/blob/2659f50ed953ce231012ed57d410936db3b84204/packages/core/bin/update-boilerplate-dependencies.js) that does the dep version replacement in package.json. I deleted the Node.js in favor of this `sed` one-liner.

.gitignore
```diff
@@ -92,3 +92,5 @@ typings/
 
 example-apps/*/yarn.lock
 example-apps/*/package-lock.json
+
+boilerplate/build
```
eliangcs: This is the place where we can find the output boilerplate zip.

package.json
```diff
@@ -28,9 +28,7 @@
     "lint": "lerna run --stream --ignore zapier-platform-example-app-* lint",
     "lint-examples": "eslint examples",
     "validate": "lerna run --ignore zapier-platform-example-app-* validate",
-    "bump": "./scripts/bump.js",
-    "build-boilerplate": "yarn workspace zapier-platform-core build-boilerplate",
-    "upload-boilerplate": "yarn workspace zapier-platform-core upload-boilerplate"
```
eliangcs: If fully automated, running the scripts locally would be less necessary. Let's remove them from yarn scripts to keep `package.json` clean and simple.

packages/core/package.json
```diff
@@ -27,9 +27,6 @@
     "upload-integration-test": "bin/upload-lambda.js local.bundle.zip",
     "deploy-integration-test":
       "yarn build-integration-test && yarn upload-integration-test",
-    "build-boilerplate": "bin/build-boilerplate.sh",
-    "upload-boilerplate": "bin/upload-boilerplate.sh",
-    "postpublish": "bin/build-boilerplate.sh && bin/upload-boilerplate.sh",
```
eliangcs: The `postpublish` part has been moved to the bottom of `REPO_ROOT/scripts/publish.sh`. The boilerplate needs updating not only when there's a new version of `core` but also when `legacy-scripting-runner` has a new version.

scripts/publish.sh
```diff
@@ -28,5 +28,28 @@ echo "$NPM_REGISTRY_NO_PROTO/:_authToken=\"$NPM_TOKEN\"" > $NPMRC_PATH
...
+    CORE_VERSION=$(curl https://registry.npmjs.org/zapier-platform-core/latest | jq -r .version)
+    LEGACY_VERSION=$(cat $PKG_PATH/package.json | jq -r .version)
+fi
+
+if [[ "$CORE_VERSION" != "" && "$LEGACY_VERSION" != "" ]]; then
+    # Build boilerplate and let Zapier know about this
+    ./boilerplate/scripts/build.sh $CORE_VERSION $LEGACY_VERSION &&
+    ./boilerplate/scripts/upload.sh $CORE_VERSION &&
+    ZIP_HASH=$(sha1sum "./boilerplate/build/$CORE_VERSION.zip" | awk '{print $1}') &&
+    curl -H 'content-type:application/json' -d "{\"version\":\"$CORE_VERSION\",\"id\":\"$ZIP_HASH\"}" $BOILERPLATE_UPDATE_URL | jq .status
```
eliangcs: `BOILERPLATE_UPDATE_URL` has been configured on Travis. It's a hook to which we can post requests to notify Zapier to update any internal resources.

scripts/publish.sh
```diff
@@ -28,5 +28,28 @@ echo "$NPM_REGISTRY_NO_PROTO/:_authToken=\"$NPM_TOKEN\"" > $NPMRC_PATH
...
+
+if [[ "$PKG_NAME" == "zapier-platform-core" ]]; then
+    LEGACY_VERSION=$(curl https://registry.npmjs.org/zapier-platform-legacy-scripting-runner/latest | jq -r .version)
+    CORE_VERSION=$(cat $PKG_PATH/package.json | jq -r .version)
+fi
+
+if [[ "$PKG_NAME" == "zapier-platform-legacy-scripting-runner" ]]; then
+    CORE_VERSION=$(curl https://registry.npmjs.org/zapier-platform-core/latest | jq -r .version)
+    LEGACY_VERSION=$(cat $PKG_PATH/package.json | jq -r .version)
+fi
```
eliangcs: So this means we're going to update the boilerplate when either `core` or `legacy-scripting-runner` has a new version.

- If `core` has a new version, we build a new boilerplate with the new `core` and the existing latest `legacy-scripting-runner`.
- If `legacy-scripting-runner` has a new version, we build a boilerplate with the existing latest `core` and the new `legacy-scripting-runner`. Then we replace the boilerplate with the new one we just built.

boilerplate/scripts/build.sh
```diff
@@ -0,0 +1,136 @@
...
+# Monkey patch boilerplate package.json so zapier-platform-core and
+# zapier-platform-legacy-scripting-runner have version numbers specified instead
+# of local file path
+update_deps "$BOILERPLATE_DIR/package.json" $CORE_VERSION $LEGACY_VERSION
+
+pushd $BOILERPLATE_DIR > /dev/null
+
+# Build the zip!
+# the node-X segment in the next line should match the latest major version
+zip -R $TARGET_FILE '*.js' '*.json' '*/linux-x64-node-10/*.node'
```
xavdid: I think this is fine for now, but we'll eventually have to include node 12 as well to avoid the issue we had with the node 8 deprecation (where converted apps using this were running on node 10 instead of 8).

No action for now, but it's a thing to consider.

eliangcs: That's a good point! I tested and the zip size only increases by 0.3% if we include the Node 12 binding as well. Included both Node 10 and 12 in 178f837.

PR #154, opened 2020-02-13
fix(legacy-scripting-runner): stop immediately when scripting method errors

eliangcs:
See 👇 for why.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -530,6 +530,7 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
             result = method(convertedBundle, optionalCallback);
           } catch (err) {
             reject(err);
+            return;
           }
 
           // Handle sync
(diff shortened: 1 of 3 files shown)


packages/legacy-scripting-runner/index.js
```diff
@@ -530,6 +530,7 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
             result = method(convertedBundle, optionalCallback);
           } catch (err) {
             reject(err);
+            return;
```
eliangcs: A tricky one-liner fix. We should stop immediately if the scripting method throws an error. Without the `return` after the `reject`, the program would keep running using the undefined `result`, which leads to an error like `TypeError: Cannot read property 'files' of undefined` happened at this [line](https://github.com/zapier/zapier-platform/blob/9e2a6945b3a8edf95b030e599e5afb016431df8c/packages/legacy-scripting-runner/index.js#L166).

PR #159, opened 2020-02-24
fix(cli): fix convert command so it correctly handles a VB app converted from WB

eliangcs:
Addresses [PDE-1339](https://zapierorg.atlassian.net/browse/PDE-1339). Fixes `zapier convert` command so it exports a VB app that was converted from WB more properly. Fixes two things:

* In `package.json`, include `legacy-scripting-runner` dependency
* In `.zapierapprc`, add `"includeInBuild": ["scripting.js"]`

Also removes an unused script `test-convert.js`.


packages/cli/src/utils/convert.js
```diff
@@ -97,35 +97,46 @@ const getAuthFieldKeys = appDefinition => {
...
-    description,
-    appId: legacyApp.general.app_id || 'null',
-    cliVersion: PACKAGE_VERSION,
-    coreVersion: appDefinition.platformVersion,
-    runnerVersion
+  const dependencies = {
+    [PLATFORM_PACKAGE]: appDefinition.platformVersion
   };
+  if (appDefinition.legacy) {
+    const RUNNER_PACKAGE = 'zapier-platform-legacy-scripting-runner';
```
xavdid: this declaration should probably be moved to https://github.com/zapier/zapier-platform/blob/master/packages/cli/src/constants.js

PR #162, opened 2020-02-26
Preserve values that also include empty curlies

stevelikesmusic:
fix(core): Preserve values that also include empty curlies in request body

Fixes a bug where we expect `body: { partial: 'text {{123__unresolved_curly}}' } }` to include `'text'` in the value sent to the API. 

Instead we were removing the whole field from the request if we found an unresolved curly. This fix preserves the hardcoded value so that:
```js
body: { partial: 'text {{123__unresolved_curly}}' } }

// becomes

body: { partial: 'text' } }
```

### DIFF
--- a/packages/core/src/tools/cleaner.js
+++ b/packages/core/src/tools/cleaner.js
@@ -18,6 +18,8 @@ const DEFAULT_BUNDLE = {
   targetUrl: ''
 };
 
+const isCurlies = /{{.*?}}/g;
+
 const recurseCleanFuncs = (obj, path) => {
   // mainly turn functions into $func${arity}${arguments}$
   path = path || [];
@@ -120,10 +122,25 @@ const createBundleBank = (appRaw, event = {}) => {
 
 const maskOutput = output => _.pick(output, 'results', 'status');
 
+// These normalize functions are called after the initial before middleware that
+// cleans the request. The reason is that we need to know why a value is empty
+// later on. If we resolve all templates (even with undefined values) first, we
+// don't know _why_ it was an empty string. Was it from a user supplied value in
+// an earlier Zap step? Or was it a null value? Each has different results depending
+// on how the partner has configued their integration.
 const normalizeEmptyRequestFields = (shouldCleanup, field, req) => {
-  const handleEmpty = req.removeMissingValuesFrom[field]
-    ? key => delete req[field][key]
-    : key => (req[field][key] = '');
+  const handleEmpty = key => {
+    const value = req[field][key] || '';
+    const cleaned = value.replace(isCurlies, '');
+
+    if (value !== cleaned) {
+      req[field][key] = cleaned;
+    }
+
+    if (!cleaned && req.removeMissingValuesFrom[field]) {
+      delete req[field][key];
+    }
+  };
 
   Object.entries(req[field]).forEach(([key, value]) => {
     if (shouldCleanup(value)) {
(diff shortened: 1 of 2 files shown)


docs/index.html
```diff
@@ -895,25 +895,25 @@ <h3 id="local-app-definition">Local App Definition</h3>
     <div class="col-md-7 col-sm-12 col-height  docs-code">
       <pre><code class="lang-js"><span class="hljs-keyword">const</span> App = {
   <span class="hljs-comment">// both version strings are required</span>
-  <span class="hljs-attr">version</span>: <span class="hljs-built_in">require</span>(<span class="hljs-string">&apos;./package.json&apos;</span>).version,
```
stevelikesmusic: Not sure what is causing this change to the docs

stevelikesmusic:
Thanks for the feedback @eliangcs! So we want to preserve the trailing whitespace then?

I think that does make sense. We don't want to combine normalizing fields with pruning from the request in the same middelware operation.

eliangcs:
@stevelikesmusic yes, I think we should preserve the trailing whitespace.

packages/core/src/http-middlewares/before/prepare-request.js
```diff
@@ -60,8 +60,6 @@ const coerceBody = req => {
   } else if (Buffer.isBuffer(req.body)) {
     // leave a buffer alone!
   } else if (req.body && !_.isString(req.body)) {
-    normalizeEmptyBodyFields(req);
```
eliangcs: I think we still need to prune empty body fields. Can we add this back?

stevelikesmusic: @eliangcs the tricky thing here now is that we apply all templates first. Now we'll never get `{{unresolved_curly}}` by the time. 

I'm remembering the reason for the order of middleware now.

If we apply all templates (even without values) then by the time we're ready to prune values from body/params, we have already replaced everything with an empty string. For instance, we should have a different behavior for a user supplied empty string and an empty string from an unresolved curly. 

I removed `normalizeEmptyBodyFields` because we allowed empty strings to exist in the body (as we assumed they were user supplied and valid) since we removed unresolved curlies. Now, we resolve all curlies first.

In that case, would it make sense to go back to the old flow of pruning and/or resolving in the same middleware step similar to the first commit?

eliangcs:
> For instance, we should have a different behavior for a user supplied empty string and an empty string from an unresolved curly.

I thought that was the intended behavior. But I know the main user of `removeMissingValuesFrom` is VB. We need a way to only prune empty values that were caused by unresolved curlies because users can't delete those themselves if they're using Form Mode. Am I understanding it correctly?

> In that case, would it make sense to go back to the old flow of pruning and/or resolving in the same middleware step similar to the first commit?

If so, then yes, let's go back with the old flow. Maybe add a comment to clarify why we do it. Sorry for the hassle 🙇

packages/core/src/http-middlewares/before/prepare-request.js
eliangcs: > For instance, we should have a different behavior for a user supplied empty string and an empty string from an unresolved curly.

I thought that was the intended behavior. But I know the main user of `removeMissingValuesFrom` is VB. We need a way to only prune empty values that were caused by unresolved curlies because users can't delete those themselves if they're using Form Mode. Am I understanding it correctly?

> In that case, would it make sense to go back to the old flow of pruning and/or resolving in the same middleware step similar to the first commit?

If so, then yes, let's go back with the old flow. Maybe add a comment to clarify why we do it. Sorry for the hassle 🙇

cvanderkolk:
Thanks for getting this out the door @stevelikesmusic! 👏

PR #170, opened 2020-03-05
fix(legacy-scripting-runner) non-string-like headers shouldn't blow up

xavdid:
Change header code to be a little more defensive around calling `.replace`


xavdid:
@eliangcs I'll go ahead and deploy this when it's ready to unblock the customer. 

All headers should be strings (afaik), so maybe we make that change. I wasn't sure of the implications of that, so I went with the safer change. Feel free to take this another direction.

Issue #8, opened 2019-06-25
Typescript bindings


xavdid:
This won't happen until we've got a 1.0 and a stable API, but bindings would be great. I'd be hapy to jump on it once we're ready!

All it'll be a `.ts` file in the repo and a line in the `package.json`

philonor:
@xavdid We are experiencing problems when trying to use Zapier's own error types as described [here](https://platform.zapier.com/cli_docs/docs#halting-execution) in combination with TypeScript.

The described code example `throw new z.errors.HaltedError('Your reason.');` does not work with TypeScript for us. Are we missing something?
- ```
   import {ZObject} from 'zapier-platform-core';
   throw new ZObject.errors.HaltedError('Your reason'). // TS2693: 'ZObject' only refers to a type, but is being used as a value here.
  ```
- ```
   import {HaltedError} from 'zapier-platform-core';
   throw new HaltedError('Your reason'); // passes with tsc, but yields "zapier_platform_core_1.HaltedError is not a constructor" error when run.
  ```
- ```
   import {HaltedError} from 'zapier-platform-core';
   throw new Error ('Your reason.') as HaltedError; // Passes tsc and works when run, but HaltedError is undefined
   ```
Reproduction repository: https://github.com/philonor/zapier-platform/tree/master

https://github.com/philonor/zapier-platform/commit/7fd74fd9d1aa4f9f88ad724af354d9a381e67af6

For now, we can only use default `Error` type which will turn off the Zap after a while instead of halting execution ([source](https://zapier.github.io/zapier-platform/#error-handling)):

> Note that if a Zap raises too many error messages it will be automatically turned off, so only use these if the scenario is truly an error that needs to be fixed.

xavdid:
@philonor the issue is that `ZObject` is just a type, not the actual object. Take the following:

```ts
import {ZObject} from 'zapier-platform-core'
const perform = (z: ZObject, ) => {
  throw z.errors.HaltedError('something wrong')
}
```

Once you type `z.errors.H`, `HaltedError` should auto-complete, thanks to marking `z` as a `ZObject`. Does that make sense?

sampsonbryce:
@xavdid We are having issues with the perform type when testing triggers that use webhooks. The type of createAppTester is
```javascript
export const createAppTester: (
  appRaw: object,
  options?: { customStoreKey?: string }
) => <T extends any, B extends Bundle>(
  func: (z: ZObject, bundle: B) => Promise<T>,
[...]

PR #172, opened 2020-03-06
fix(core) improve `appTester` types and bump node versions

xavdid:
Per discussion in [this comment](https://github.com/zapier/zapier-platform/issues/8#issuecomment-591877355), tweaking the types for `appTester` so it accepts perform methods that are synchronous.

In addition to changing the output type of `perform`, I simplified what `appTester` itself returns, since it always returns a promise. This is safe because there was a logic error before and it always evaluated to `Promise<T>` anyway. You can test that in [the playground](https://www.typescriptlang.org/play/#code/C4TwDgpgBA5hB2EBOBLAxgMQK7zVAvFADwAqAfABQCGSMAXFCQJQFmNQQAewCAJgM5QACkgD2AWxT8IpNgH4o-YKngwoDeFnEAjZAG4AUAbSj4SqCdEAbBnESpMOPIQrAAFilUMq8EC3xs7p4wRgD0oVDioljSUKIAbsgcVGhuUAAGllbpUJ7xomhUwCimADRQ7tBIEMBYSPAV4NBSUFRWAO5UIILpmjrI6QbhrYIAZjQjUACSFj4VEFZW5ZXVAOSC8KK58GBYwBVuRVCHglRQ1bX1jZAZSiowg1kUIhJSEAB01fzWiRQAzH8mExjKJrM8xJJpJ8IN8rL9VlR+LxRqsgSCwQAmQHoqwUBFIlHAp6aRZE0G4nC8CCjTwQXhksEAb0q8AYqxpiFWAF8GbiANp-AC6vPBryhXx+EFFkI+ErhUsBQKAA).

Also bumps the included node types. Doesn't matter much, but it's nice to have up to date ones.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -14,10 +14,10 @@ export const tools: { env: { inject: (filename?: string) => void } };
 export const createAppTester: (
   appRaw: object,
   options?: { customStoreKey?: string }
-) => <T extends any, B extends Bundle>(
-  func: (z: ZObject, bundle: B) => Promise<T>,
+) => <T, B extends Bundle>(
+  func: (z: ZObject, bundle: B) => T | Promise<T>,
   bundle?: Partial<B> // partial so we don't have to make a full bundle in tests
-) => T extends Promise<T> ? T : Promise<T>;
+) => Promise<T>; // appTester always returns a promise
 
 // internal only
 // export const integrationTestHandler: () => any;
@@ -51,12 +51,14 @@ export interface Bundle<InputData = { [x: string]: any }> {
     headers: { [x: string]: string };
     content: string;
   }>;
-  cleanedRequest?: Partial<{
-    method: HttpMethod;
-    querystring: { [x: string]: string };
-    headers: { [x: string]: string };
-    content: { [x: string]: string };
-  }> | any;
+  cleanedRequest?:
+    | Partial<{
+        method: HttpMethod;
+        querystring: { [x: string]: string };
+        headers: { [x: string]: string };
+        content: { [x: string]: string };
+      }>
+    | any;
   subscribeData?: { id: string };
   targetUrl?: string;
 }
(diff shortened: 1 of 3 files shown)


packages/core/index.d.ts
```diff
@@ -14,10 +14,10 @@ export const tools: { env: { inject: (filename?: string) => void } };
 export const createAppTester: (
   appRaw: object,
   options?: { customStoreKey?: string }
-) => <T extends any, B extends Bundle>(
-  func: (z: ZObject, bundle: B) => Promise<T>,
+) => <T, B extends Bundle>(
+  func: (z: ZObject, bundle: B) => T | Promise<T>,
   bundle?: Partial<B> // partial so we don't have to make a full bundle in tests
-) => T extends Promise<T> ? T : Promise<T>;
+) => Promise<T>; // appTester always returns a promise
```
eliangcs: This is mind-boggling, took me a while to realize it's a function returning a function. The change makes sense, but I wonder why we need that `bundle: B` part? Why not:

```ts
export const createAppTester: (
  appRaw: object,
  options?: { customStoreKey?: string }
) => <T>(
  func: (z: ZObject, bundle: Bundle) => Promise<T>,
  bundle?: Partial<Bundle> // partial so we don't have to make a full bundle in tests
) => Promise<T>;
```

Unless it's possible to pass in a `Bundle` subclass, i.e., a custom bundle?

xavdid: yeah, great question! 

that was tweaked in response to https://github.com/zapier/zapier-platform-cli/issues/339; 

Basically the default `inputData` is `{[x: string]: any}`, since that's the basic version of what it is. If you pass in a more specific `inputData` interface (which is great, because then `bundle.inputData.x` autocompletes in your editor (and warns you of missing properties)), then TS complained that the more specific type didn't line up with the more generic one. 

`extends` fixes that!

PR #174, opened 2020-03-11
fix(cli): analytics shouldn't need a valid app install

eliangcs:
`zapier analytics` is not an app-specific command, so it shouldn't need a valid app install.


eliangcs:
@xavdid let's do it in https://github.com/zapier/zapier-platform/pull/146 then. I'm closing this.

Issue #14, opened 2019-06-25
Zapier fails to build when the module depends on a relative file


itsuryev:
In my app I need to use local package, so I specify it in `package.json`:
```
  "dependencies": {
    "fibery-schema": "file:../fibery-schema",
    "lodash": "4.17.4",
    "zapier-platform-core": "5.1.0"
  },
```

zapier-cli when building copies project to a temp folder and installs dependencies there and of course can not find this package:
```
$ zapier build
Building project.

  Copying project to temp directory ⠃ done!
  Installing project dependencies ⠆

Error! npm ERR! code ENOLOCAL
npm ERR! Could not install from "../fibery-schema" as it does not contain a package.json file.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/itsu/.npm/_logs/2018-04-23T14_46_40_886Z-debug.log
```
In logs:
```
56 verbose stack Error: ENOENT: no such file or directory, open '/private/var/folders/bf/9_3xh2310kl_p8_wxwpfpb180000gn/T/fibery-schema/package.json'
57 verbose cwd /private/var/folders/bf/9_3xh2310kl_p8_wxwpfpb180000gn/T/zapier-2ec52d0b
```

Zapier information:
```
$ zapier -v
zapier-platform-cli/5.2.0
zapier-platform-core/5.1.0
node/v9.11.1
```
While I can use absolute path I can not commit it as it will not work for other developers of our app.

bradleyayers:
I'm running into a similar problem when using yarn workspaces.

xavdid:
can you give us a quick overview of the shape of the repo? That's a thing we'll want to fix at some point, and having a reproducible example will go a long way.

bradleyayers:
Hi @xavdid sure, here's the rough outline.

```
.
├── README.md
├── docker-compose.yml
├── tsconfig-base.json
├── tslint.json
├── node_modules
├── workspaces
│   ├── app-lib
│   │   ├── package.json
│   │   ├── node_modules
│   │   └── …
│   ├── emails
│   │   ├── package.json
│   │   ├── node_modules
│   │   └── …
│   ├── hal-9000
│   │   ├── package.json
│   │   ├── node_modules
│   │   └── …
│   ├── toolbox
│   │   ├── package.json
│   │   ├── node_modules
│   │   └── …
│   ├── ui-components
│   │   ├── package.json
│   │   ├── node_modules
│   │   └── …
│   ├── website
[...]

PR #175, opened 2020-03-11
refactor(core): fix smoke test cyclic dependency

eliangcs:
Addresses [PDE-1318](https://zapierorg.atlassian.net/browse/PDE-1318).

There's a cyclic dependency in our release process:
1. `yarn bump` bumps example apps' core version
2. core's smoke tests do live `npm install` on the example apps. Since the new core version hasn't published to npm yet, the install would fail.
3. To successfully `npm publish`, smoke tests must pass. But smoke tests couldn't pass because of 2.

This PR addresses 2 so core's smoke tests no longer download the new core version from npm.

There are some minor changes. I'll annotate them in the comments.

### DIFF
--- a/packages/cli/src/smoke-tests/smoke-tests.js
+++ b/packages/cli/src/smoke-tests/smoke-tests.js
@@ -82,12 +82,14 @@ describe('smoke tests - setup will take some time', () => {
     fs.removeSync(context.workdir);
   });
 
-  it('package size should not change much', async () => {
+  it('package size should not change much', async function() {
     const packageName = 'zapier-platform-cli';
     const latestVersion = await getPackageLatestVersion(packageName);
     const baselineSize = await getPackageSize(packageName, latestVersion);
     const newSize = fs.statSync(context.package.path).size;
     newSize.should.be.within(baselineSize * 0.7, baselineSize * 1.3);
+
+    this.test.title += ` (${baselineSize} -> ${newSize} bytes)`;
   });
 
   it('cli executable should exist', () => {
@@ -134,11 +136,14 @@ describe('smoke tests - setup will take some time', () => {
     pkg.name.should.containEql('babel');
   });
 
-  it('zapier apps', function() {
+  it('zapier integrations', function() {
     if (!context.hasRC) {
       this.skip();
     }
-    const stdout = runCommand(context.cliBin, ['apps', '--format=json']);
+    const stdout = runCommand(context.cliBin, [
+      'integrations',
+      '--format=json'
+    ]);
     const result = JSON.parse(stdout);
     result.should.be.Array();
   });
--- a/packages/core/smoke-test/smoke-test.js
+++ b/packages/core/smoke-test/smoke-test.js
@@ -9,6 +9,8 @@ const path = require('path');
 require('should');
 const fetch = require('node-fetch');
 
+const CORE_PACKAGE_NAME = 'zapier-platform-core';
+
 const TEST_APPS = [
   'basic-auth',
   'create',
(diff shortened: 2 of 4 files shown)


packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -82,12 +82,14 @@ describe('smoke tests - setup will take some time', () => {
...
 
-  it('package size should not change much', async () => {
+  it('package size should not change much', async function() {
     const packageName = 'zapier-platform-cli';
     const latestVersion = await getPackageLatestVersion(packageName);
     const baselineSize = await getPackageSize(packageName, latestVersion);
     const newSize = fs.statSync(context.package.path).size;
     newSize.should.be.within(baselineSize * 0.7, baselineSize * 1.3);
+
+    this.test.title += ` (${baselineSize} -> ${newSize} bytes)`;
```
eliangcs: This would give us how exactly the file size changes in the test output: https://travis-ci.com/github/zapier/zapier-platform/builds/152761505#L571

packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -82,12 +82,14 @@ describe('smoke tests - setup will take some time', () => {
     fs.removeSync(context.workdir);
   });
 
-  it('package size should not change much', async () => {
+  it('package size should not change much', async function() {
```
eliangcs: To access `this` in the test case, we have to use `function()` instead of an arrow function.

packages/core/smoke-test/smoke-test.js
```diff
@@ -88,23 +90,38 @@ const copyTestApps = workdir => {
   TEST_APPS.forEach(appName => {
     const srcAppDir = path.join(repoRoot, 'example-apps', appName);
     const destAppDir = path.join(workdir, appName);
-    fs.copySync(srcAppDir, destAppDir);
+    fs.copySync(srcAppDir, destAppDir, {
+      filter: (src, dest) => !src.includes('node_modules/')
```
eliangcs: This is a small bug I found when we copy example apps to temporary directories. Example apps could have a `node_module/` locally. We should avoid copying that to the temporary directory.

packages/core/smoke-test/smoke-test.js
```diff
@@ -194,15 +213,18 @@ describe('smoke tests - setup will take some time', () => {
...
-        const proc = spawnSync(context.cliBin, ['build'], {
-          encoding: 'utf8',
-          cwd: context.workAppDir,
-          env: {
-            SKIP_NPM_INSTALL: '1',
-            PATH: process.env.PATH,
-            DISABLE_ZAPIER_ANALYTICS: 1
+        const proc = spawnSync(
+          context.cliBin,
+          ['build', '--skip-npm-install'],
```
eliangcs: `SKIP_NPM_INSTALL` no longer works. Should change it to `--skip-npm-install`.

scripts/bump.js
```diff
@@ -185,6 +210,7 @@ const bumpExtensionPackage = (packageName, versionToBump) => {
 const bumpPackages = (packageName, versionToBump) => {
   if (packageName === 'cli, core, schema') {
     bumpMainPackages(versionToBump);
+    bumpMainPackagesForExampleApps(versionToBump);
```
eliangcs: I removed bumping example apps in bf48591e8a to unblock my release of 9.1.0. This only adds it back.

packages/core/smoke-test/smoke-test.js
```diff
@@ -88,23 +90,38 @@ const copyTestApps = workdir => {
...
-  });
+const npmInstalls = (coreZipPath, workdir) => {
+  // When releasing a new core version, example apps would have bumped their
+  // core before the new core version is published to npm. So we need to patch
+  // example app's package.json to use local core temporarily to avoid `npm
+  // install` error.
+  const packageJsonPath = path.join(workdir, 'package.json');
+  const origPackageJsonText = fs.readFileSync(packageJsonPath);
+  const packageJson = JSON.parse(origPackageJsonText);
+  packageJson.dependencies[CORE_PACKAGE_NAME] = `file:${coreZipPath}`;
```
xavdid: this might run afoul of #14 but if it works, then that's great!

eliangcs: Thanks for the callout! `coreZipPath` is an absolute path so it works here.

PR #179, opened 2020-03-12
Allow env variables containing equals

robwa10:
fix(cli): Correctly parse out env variables that contain an equal sign

Fixes a bug where we don't correctly parse env variables that contain an equal sign as part of the variable. An equal sign is a valid special character and occurs reasonably often in things such as CLIENT_SECRETS, etc. This seems to be a regression as the previous `env` command didn't parse those incorrectly.

### DIFF
--- a/packages/cli/src/oclif/commands/env/set.js
+++ b/packages/cli/src/oclif/commands/env/set.js
@@ -30,7 +30,8 @@ class SetEnvCommand extends BaseCommand {
     // if we get here, we should have well-formed input
 
     const payload = valuesToSet.reduce((result, kvPair) => {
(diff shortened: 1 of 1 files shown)


robwa10:
Ah, yes, spread syntax. I knew there was a more concise way forward here! Also, never considered multi-line inputs like public keys so a great catch there. 👍

xavdid:
cool! feel free to merge at your leisure.

robwa10:
@xavdid Github doesn't trust me...to be fair I've given them some good reasons early on in my Github experiences... 😂

xavdid:
odd! I'll merge it and take a closer look at why you can't

PR #181, opened 2020-03-18
fix(legacy-scripting-runner): full jQuery support

eliangcs:
Fixes [PDE-1357](https://zapierorg.atlassian.net/browse/PDE-1357).

We've been using `cheerio` and `jquery-param` for a lighter alternative for `jquery` and `jsdom`. But we found it hard to recreate identical jQuery capabilities this way. This PR includes `jquery` and `jsdom` back to dependencies and removes `cheerio` and `jquery-param`. There's a new test case demonstrating a common usage of jQuery DOM manipulation, which is hard to replicate if we do it on our own.

After adding this, the boilerplate zip will grow from 3.3 MB to 4.7 MB.

### DIFF
--- a/packages/legacy-scripting-runner/package.json
+++ b/packages/legacy-scripting-runner/package.json
@@ -24,10 +24,10 @@
   },
   "dependencies": {
     "async": "2.6.1",
-    "cheerio": "1.0.0-rc.2",
     "deasync": "0.1.15",
     "flat": "4.1.0",
-    "jquery-param": "1.0.1",
+    "jquery": "2.1.4",
+    "jsdom": "7.0.0",
     "lodash": "4.17.15",
     "moment-timezone": "0.5.17",
     "request": "2.87.0",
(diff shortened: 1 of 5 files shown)


packages/legacy-scripting-runner/package.json
```diff
@@ -24,10 +24,10 @@
   },
   "dependencies": {
     "async": "2.6.1",
-    "cheerio": "1.0.0-rc.2",
     "deasync": "0.1.15",
     "flat": "4.1.0",
-    "jquery-param": "1.0.1",
+    "jquery": "2.1.4",
```
xavdid: In the [legacy docs](https://platform.zapier.com/legacy/scripting#available-libraries), we indicate that `jquery` is v 1.8.3. Is this version compatible enough?

Similarly, is that old underscore version compatible with the lodash we provide?

eliangcs: The doc is likely outdated. The dependency versions we use for Web Builder can be found [here](https://github.com/zapier/lambda-wrapper/blob/2e9d7523faf9bb1a12cf3ca3010b0a0568865fc8/package.json). I also double checked in Web Builder with this script:

```js
    movie_post_poll: function(bundle) {
         return [{
             'id': new Date().getTime(),
             'jqueryVersion': $.fn.jquery,
             'underscoreVersion': _.VERSION
         }];
    }
```

The result is:



So we're good ✌️

PR #189, opened 2020-04-02
feat(core): New error classes that support error codes and statuses

FokkeZB:
This adds `z.errors.Error` (`AppError`) and `z.errors.ResponseError` and uses the latter in `throwForStatus` and `throwForStaleAuth`.

- **JIRA:** `SXP-1286` [Create new Node z.errors classes](https://zapierorg.atlassian.net/browse/SXP-1286) and `SXP-1294` [Use new Node z.errors in throwForStatus](https://zapierorg.atlassian.net/browse/SXP-1294)
- **Spec:** https://github.com/zapier/sandbox-integration-errors/issues/6

The `z.error.Error` class is for developers to use instead of `Error` and takes an optional second `code` (string) and third `status` (number) argument. All arguments are `JSON.stringify`-ed as the message of the error. 

In [Python-land](https://github.com/zapier/zapier/blob/e03c532d97e4e647f44038eb5a35ce4b30de2b48/web/backend/zapier/developer_cli/api_base.py#L1114-L1126) we `json.loads` the message and raise the Python `AppError` class, which takes the same arguments. The code and status get retained in our error logs and task and editor states so that the frontend can show relevant troubleshooting docs for the code and/or status.

The `z.error.ResponseError` class is primarily for internal use and used in `throwForStatus` and `throwForStaleAuth`. It takes the `response` and `JSON.stringify`s what we need in [Python-land](https://github.com/zapier/zapier/blob/e03c532d97e4e647f44038eb5a35ce4b30de2b48/web/backend/zapier/developer_cli/api_base.py#L1129) to rebuild the response as a Python `requests.Response` and use our existing v1/v2 tooling to extract an error message, code and status from the response, and then raise `AppError`.

# Testing

- Includes updated and new tests.
- Use https://zapier.com/developer/public-invite/97300/9d0a334481cf89c39060d0b72e21c836/ to invite yourself to the publication of https://github.com/zapier/zapier-platform-example-app-errors to test with different errors.
- Or re-test the steps in the shared Zap at: https://zapier.com/app/editor/82925946 that uses the above app. Keep an eye on the network requests to `.. events` for the event `rendererdError` and look for the `errorCode` and `errorStatus` properties.

    

# Notes

- I use a `doNotContextify` flag to opt out the new classes for appending `What happened:` and `Console logs:` to their (`JSON.stringify`-ed) `message`.
  - To be honest, I don't like that we append to errors their messages at all and wonder why we need this and if, [like suggested here](https://github.com/zapier/zapier/blob/e03c532d97e4e647f44038eb5a35ce4b30de2b48/web/backend/zapier/developer_cli/api_base.py#L1082-L1099), getting them the step UUID and letting them search the existing bundle and request logs in their monitor by that wouldn't get them the same if not more information. Think of Sentry for Zapier Platform, give the ID of an error and we show you what led to it.
    - We could also choose to always turn `message` into a JSON object so that we're able to pass along any context, in a more structured way then appending it to the error message.
- @eliangcs brought up that part of why `throwForStatus` didn't use the response body was that it may contain sensitive data. We talked about how this merely bring v3 on par with v1/v2, so any security concerns would apply to them as well and probably are best address in the Python error message/code extraction logic. For example, by making sure none of the authentication details are part of the extracted message.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -704,6 +704,7 @@ We provide several methods off of the `z` object, which is provided as the first
 
 The available errors are:
 
+* Error - Stops the current operation, allowing for (auto) replay. Read more on [General Errors](#general-errors)
 * HaltedError - Stops current operation, but will never turn off Zap. Read more on [Halting Execution](#halting-execution)
 * ExpiredAuthError - Turns off Zap and emails user to manually reconnect. Read more on [Stale Authentication Credentials](#stale-authentication-credentials)
 * RefreshAuthError - (OAuth2 or Session Auth) Tells Zapier to refresh credentials and retry operation. Read more on [Stale Authentication Credentials](#stale-authentication-credentials)
@@ -1222,17 +1223,22 @@ various kinds of errors occur.
 ### General Errors
 
 Errors due to a misconfiguration in a user's Zap should be handled in your app
-by throwing a standard JavaScript `Error` with a user-friendly message.
-Typically, this will be prettifying 4xx responses or APIs that return errors as
-200s with a payload that describes the error.
+by throwing `z.errors.Error` with a user-friendly message and optional error and
+status code. Typically, this will be prettifying 4xx responses or APIs that return
+errors as 200s with a payload that describes the error.
 
-Example: `throw new Error('Your error message.');`
+Example: `throw new z.errors.Error('Contact name is too long.', 'InvalidData', 400);`
 
 A couple best practices to keep in mind:
 
   * Elaborate on terse messages. "not_authenticated" -> "Your API Key is invalid. Please reconnect your account."
-  * If the error calls out a specific field, surface that information to the user. "Invalid Request" -> "contact name is invalid"
-  * If the error provides details about why a field is invalid, add that in too! "contact name is invalid" -> "contact name is too long"
+  * If the error calls out a specific field, surface that information to the user. "Provided data is invalid" -> "Contact name is invalid"
+  * If the error provides details about why a field is invalid, add that in too! "Contact name is invalid" -> "Contact name is too long"
+  * The second, optional argument should be a code that a computer could use to identify the type of error.
+  * The last, optional argument should be the HTTP status code, if any.
+
+The code and status can be used by us to provide relevant troubleshooting to the
+user when we communicate the error.
 
 Note that if a Zap raises too many error messages it will be automatically
 turned off, so only use these if the scenario is truly an error that needs to
@@ -1247,7 +1253,7 @@ in a create to add an email address to a list where duplicates are not allowed,
 you would want to throw a `HaltedError` if the Zap attempted to add a duplicate.
 This would indicate failure, but it would be treated as a soft failure.
 
-Unlike throwing `Error`, a Zap will never by turned off when this error is thrown
+Unlike throwing `z.errors.Error`, a Zap will never by turned off when this error is thrown
 (even if it is raised more often than not).
 
 Example: `throw new z.errors.HaltedError('Your reason.');`
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -63,6 +63,9 @@ export interface Bundle<InputData = { [x: string]: any }> {
   targetUrl?: string;
 }
 
+declare class AppError extends Error {
+  constructor(message: string, code?: string, status?: number);
+}
 declare class HaltedError extends Error {}
 declare class ExpiredAuthError extends Error {}
 declare class RefreshAuthError extends Error {}
@@ -171,6 +174,7 @@ export interface ZObject {
   ) => string;
 
   errors: {
+    Error: typeof AppError;
     HaltedError: typeof HaltedError;
     ExpiredAuthError: typeof ExpiredAuthError;
     RefreshAuthError: typeof RefreshAuthError;
--- a/packages/core/src/errors.js
+++ b/packages/core/src/errors.js
@@ -3,6 +3,39 @@
 const util = require('util');
 const _ = require('lodash');
 
+class AppError extends Error {
+  constructor(message, code, status) {
+    super(
+      JSON.stringify({
+        message,
+        code,
+        status
+      })
+    );
+    this.name = 'AppError';
+    this.doNotContextify = true;
+  }
+}
+
+class ResponseError extends Error {
+  constructor(response) {
+    super(
+      JSON.stringify({
+        status: response.status,
+        headers: {
+          'content-type': response.headers.get('content-type')
+        },
+        content: response.content,
+        request: {
+          url: response.request.url
+        }
+      })
+    );
+    this.name = 'ResponseError';
+    this.doNotContextify = true;
+  }
+}
+
 // Make some of the errors we'll use!
 const createError = name => {
   const NewError = function(message = '') {
@@ -33,7 +66,10 @@ const exceptions = _.reduce(
     col[name] = createError(name);
     return col;
   },
-  {}
+  {
+    Error: AppError,
+    ResponseError
+  }
 );
 
 const isRequireError = ({ name, message }) =>
--- a/packages/core/src/http-middlewares/after/throw-for-stale-auth.js
+++ b/packages/core/src/http-middlewares/after/throw-for-stale-auth.js
@@ -1,14 +1,14 @@
 'use strict';
 
-const { stripQueryFromURL } = require('../../tools/http');
-
 const errors = require('../../errors');
 
+/**
+ * The server-side will raise `RefreshAuthError` when `autoRefresh === true`.
+ * Once we always run `throwForStatus` or a custom `afterResponse`, we can drop `throwForStaleAuth`.
+ */
 const throwForStaleAuth = resp => {
   if (resp.status === 401) {
-    const cleanURL = stripQueryFromURL(resp.request.url);
-    const message = `Got ${resp.status} calling ${resp.request.method} ${cleanURL}, triggering auth refresh.`;
-    throw new errors.RefreshAuthError(message);
+    throw new errors.ResponseError(resp);
   }
 
   return resp;
--- a/packages/core/src/http-middlewares/after/throw-for-status.js
+++ b/packages/core/src/http-middlewares/after/throw-for-status.js
@@ -1,12 +1,10 @@
 'use strict';
 
-const { stripQueryFromURL } = require('../../tools/http');
+const errors = require('../../errors');
 
 const throwForStatus = resp => {
   if (resp.status > 300) {
-    const cleanURL = stripQueryFromURL(resp.request.url);
-    const message = `Got ${resp.status} calling ${resp.request.method} ${cleanURL}, expected 2xx.`;
-    throw new Error(message);
+    throw new errors.ResponseError(resp);
   }
 
   return resp;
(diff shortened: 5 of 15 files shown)


packages/core/src/errors.js
```diff
@@ -3,6 +3,38 @@
...
+      })
+    );
+    this.name = 'ResponseError';
+    this.contextify = false;
+  }
+}
+
 // Make some of the errors we'll use!
 const createError = name => {
   const NewError = function(message = '') {
```
FokkeZB, line 40: I'm not sure why we use this pattern, but it makes it impossible to do:

```
const error = new z.errors.HaltedError();
assert error instanceof z.errors.HaltedError
```

as `error` will be an instance of the `NewError` function here.

packages/core/test/errors.js
```diff
@@ -0,0 +1,44 @@
...
+      error.message.should.eql(
+        '{"message":"My Message","code":"MyCode","status":400}'
+      );
+    });
+  });
+
+  describe('ResponseError', () => {
+    it('should stringify arguments', async () => {
+      const testLogger = () => Promise.resolve({});
+      const input = createInput({}, {}, testLogger);
```
mikesol314, line 31: Personal learning question - what is going on here? I'm not sure why we're resolving a promise with an empty object or what createInput is doing for us.

FokkeZB: Good question - I copy pasted this from places like https://github.com/zapier/zapier-platform/blob/f3e81cab5c3fbcc68e4b0954de7c4c5edf9f2dd3/packages/core/test/tools/file-stasher.js#L17

FokkeZB:
Thanks @mikesol314 - good point on the docs - I'll update anything in this repo as part of this PR.

xavdid:
I ran out of time this week, but am planning on taking a look at this early next week!

FokkeZB:
Thanks for the pairing @eliangcs! I've updated the code to use a separate flag for letting the new errors skip context while still making sure existing errors get `Console logs:` added.

packages/core/src/tools/promise.js
```diff
@@ -23,7 +23,7 @@ const enrichErrorMessage = (err, frameStack) => {
 
 // Because reject callbacks stack, we have to skip errors we've already seen.
 const contextifyOnce = (err, attachErrorTrace, frameStack) => {
-  if (!err || err.isContextified) {
+  if (!err || err.doNotContextify || err.isContextified) {
```
FokkeZB, line 26: This makes sure that we don't append `What happened:` to the `JSON.stringify`-ed message of the new errors.

packages/core/src/tools/create-lambda-handler.js
```diff
@@ -165,7 +165,7 @@ const createLambdaHandler = appRawOrPath => {
       // the default behavior with callbacks anyway, but don't want
       // to rely on that.
       logger(logMsg, logData).then(() => {
-        if (!constants.IS_TESTING && err) {
+        if (!constants.IS_TESTING && err && !err.doNotContextify) {
```
FokkeZB, line 168: This makes sure that we don't append `Console logs:` to the `JSON.stringify`-ed message of the new errors.

packages/core/src/http-middlewares/after/throw-for-status.js
```diff
@@ -1,12 +1,10 @@
 'use strict';
 
-const { stripQueryFromURL } = require('../../tools/http');
+const errors = require('../../errors');
 
 const throwForStatus = resp => {
   if (resp.status > 300) {
-    const cleanURL = stripQueryFromURL(resp.request.url);
```
eliangcs, line 7: Looks like `stripQueryFromURL` will be no longer used after this change. If so, let's remove it.

packages/core/test/errors.js
```diff
@@ -0,0 +1,44 @@
...
+      const request = createAppRequestClient(input);
+
+      const response = await request({ url: 'https://httpbin.org/status/400' });
```
eliangcs, line 34: We should use our in-house httpbin `https://httpbin.zapier-tooling.com/status/400` here.

packages/core/src/http-middlewares/after/throw-for-status.js
```diff
@@ -1,12 +1,10 @@
...
-    const message = `Got ${resp.status} calling ${resp.request.method} ${cleanURL}, expected 2xx.`;
-    throw new Error(message);
+    throw new errors.ResponseError(resp);
```
eliangcs, line 7: (IMPORTANT) As discussed in our call, we should double check we don't introduce a regression of https://github.com/zapier/zapier/issues/32319, where users can see private info of the app in the request query params when an error happens.

packages/core/test/errors.js
FokkeZB: Fine with me, but https://github.com/zapier/zapier-platform/blob/3a3fd6a43160e6bee36b919dd9ded1c1baf57e3b/packages/core/test/http-middleware.js#L349 is the only test that uses that at the moment and all other tests use `https://httpbin.org/`, so if we prefer our in-house I suggest we switch all tests in a separate PR.

eliangcs: > we prefer our in-house I suggest we switch all tests in a separate PR

Sounds good!

packages/core/src/http-middlewares/after/throw-for-status.js
FokkeZB: We only use the URL in https://github.com/zapier/zapier/blob/a69df1a0e98159383489e99547e1e7a2cc30c2bb/web/backend/zapier/errorbox/extract.py#L403-L412 and we probably gonna drop that code so then I could actually update `ResponseError` to not include it in the `JSON.stringify`-ed message at all.

That said, I do like the idea we discussed to update https://github.com/zapier/zapier/blob/a69df1a0e98159383489e99547e1e7a2cc30c2bb/web/backend/zapier/errorbox/extract.py#L205 to use a passed authentication to ensure that we don't extract sensitive stuff from a response body.

However, as this would be a `zapier/zapier` PR and I don't think it needs to be blocking as this PR simply brings v3 on par with v1/2 error extraction.

FokkeZB:
Thanks @eliangcs. The documentation changes are already part of this PR: https://github.com/zapier/zapier-platform/pull/189/files#diff-7bc92a96c133e615b583d45c88a68516

packages/core/src/http-middlewares/after/throw-for-status.js
eliangcs: > I don't think it needs to be blocking as this PR simply brings v3 on par with v1/2 error extraction.

I meant we should do a quick test to see if it would create a security issue like https://github.com/zapier/zapier/issues/32319. I just did test with an app that uses a `beforeRequest` middleware that adds the API key to the request URL:

```
const includeApiKey = (request, z, bundle) => {
  if (bundle.authData.api_key) {
    request.params = request.params || {};
    request.params.api_key = bundle.authData.api_key;
  }
  return request;
};
```

and with a trigger that always throws an error:

```
    url: 'https://auth-json-server.zapier-staging.com/movies',
  });
  throw new z.errors.ResponseError(response);
};
```

When testing in the Editor, the API key isn't in the error message (which is good):



The part that says `app.com` is weird though, but that doesn't block this PR.

FokkeZB: Ha, good catch - that is because I'm not actually using the URL in https://github.com/zapier/zapier/blob/a69df1a0e98159383489e99547e1e7a2cc30c2bb/web/backend/zapier/developer_cli/utils.py#L692. I'll fix that.

FokkeZB: PR: https://github.com/zapier/zapier/pull/38615

packages/cli/README-source.md
```diff
@@ -1222,17 +1223,22 @@ various kinds of errors occur.
...
 
-Example: `throw new Error('Your error message.');`
+Example: `throw new z.errors.Error('Contact name is too long.', 'InvalidData', 400);`
 
 A couple best practices to keep in mind:
 
-  * Elaborate on terse messages. "not_authenticated" -> "Your API Key is invalid. Please reconnect your account."
-  * If the error calls out a specific field, surface that information to the user. "Invalid Request" -> "contact name is invalid"
-  * If the error provides details about why a field is invalid, add that in too! "contact name is invalid" -> "contact name is too long"
+  * Elaborate on terse messages. "InvalidData" -> "Provided data is invalid"
```
xavdid, line 1234: I'd re-add the `"not_authenticated" -> "Your API Key is invalid. Please reconnect your account."` line somewhere, which helps illustrate turning a code-like thing into actual directions for the user. `InvalidData` already communicates that `data is invalid`

packages/core/src/errors.js
xavdid: Used mostly so that we didn't need to write `class X extends Error {` over and over. But we can probably clean this up with newer JS features.

packages/core/index.d.ts
```diff
@@ -63,6 +63,7 @@ export interface Bundle<InputData = { [x: string]: any }> {
   targetUrl?: string;
 }
 
+declare class AppError extends Error {}
```
xavdid, line 66: `AppError` has a different signature than `Error` with its 3 arguments; that should be reflected here

packages/core/src/http-middlewares/after/throw-for-stale-auth.js
```diff
@@ -1,14 +1,14 @@
 'use strict';
 
-const { stripQueryFromURL } = require('../../tools/http');
-
 const errors = require('../../errors');
 
+/**
+ * The Python-side will raise `RefreshAuthError` when `autoRefresh === true`.
```
xavdid, line 6: super minor, but I might refer to the `server` instead of the `Python-side`. it's helps clarify that this happens in the runner, not just in something that uses python.

packages/core/test/errors.js
xavdid: nothin fancy, that's calling this function: https://github.com/zapier/zapier-platform/blob/864229ce495d0a9b6244a78910f173ba696af55d/packages/core/src/tools/create-input.js#L12

In production, we call it here: https://github.com/zapier/zapier-platform/blob/864229ce495d0a9b6244a78910f173ba696af55d/packages/core/src/tools/create-lambda-handler.js#L208-L209

packages/core/test/errors.js
xavdid: Ours behaves slightly differently (iirc wraps everything in arrays when the public one doesn't), so it's just enough friction that we haven't bothered to update it yet.

packages/core/src/errors.js
FokkeZB: A good one for a separate PR - would be nice to then be able to use `instanceof` on all of them.

packages/core/test/errors.js
FokkeZB: Looks like y'all are 🆗 with swapping all in a separate PR.

packages/cli/README-source.md
FokkeZB: The reason I changed the first one is so hat it tied back to the code argument of the example, but it doesn't add much so I'm happy to change it back.

packages/core/index.d.ts
FokkeZB: Nice catch - done!

FokkeZB:
@xavdid Thanks for the review. I've made the last changed and merged.

@eliangcs how do we go about shipping this? STAR would like to use this ASAP. Is there a release cadence of some sort?

eliangcs:
@FokkeZB we don't have a release schedule. We ship it on demand. How about we cut a release  9.3.0 my tomorrow?

FokkeZB:
Sounds good! Thanks!

PR #190, opened 2020-04-07
feat(core,schema,legacy-scripting-runner): allow to "reliably interpolate arrays or objects to a string"

eliangcs:
Addresses [PDE-1468](https://zapierorg.atlassian.net/browse/PDE-1468).

A common error of a converted app is:

"Cannot reliably interpolate objects or arrays into a string"

This is because the app uses curlies in their polling/action URL and the variable in the curlies is an array, such as `https://example.com?param={{an_array_variable}}`. Web Builder would replace `{{an_array_variable}}` with a comma-separated string such as `a,b,c`. But in VB, since core 8.0.0 with this breaking change (https://github.com/zapier/zapier-platform-core/pull/139), that syntax is no longer supported and would raise an error.

We've been advising partners to adjust code on their side. But there are just too many of them for us to handle it efficiently. I think to fix it, we can add an optional option `serializeValueForCurlies` to the `RequestSchema`. When specified, `serializeValueForCurlies` is a function that transforms an unknown type of a value to a string. The usage would be like:

```js
  url: 'https://httpbin.zapier-tooling.com/get?{{bundle.inputData.anArray}}',
  serializeValueForCurlies: value => {
    if (Array.isArray(value)) {
      return value.join(',');
    }
    return value;
  }
});
```

Then the `prepareRequest` middleware in the core will invoke this function to serialize values in the bank before it replaces curlies. legacy-scripting-runner can also use this feature to serialize arrays or objects like we used to do prior to core 8.0.0.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -630,6 +630,16 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
       }
 
       request.headers = cleanHeaders(request.headers);
+      request.serializeValueForCurlies = value => {
+        if (Array.isArray(value)) {
+          return value.join(',');
+        } else if (_.isPlainObject(value)) {
+          // Not sure if anyone would ever expect a '[object Object]',
+          // but who knows?
+          return value.toString();
+        }
+        return value;
+      };
 
       const response = await zcli.request(request);
 
--- a/packages/legacy-scripting-runner/test/integration-test.js
+++ b/packages/legacy-scripting-runner/test/integration-test.js
@@ -756,6 +756,37 @@ describe('Integration Test', () => {
       });
     });
 
+    it('KEY_pre_poll, array curlies', () => {
+      const appDef = _.cloneDeep(appDefinition);
+      appDef.legacy.scriptingSource = appDef.legacy.scriptingSource.replace(
+        'movie_pre_poll_array_curlies',
+        'movie_pre_poll'
+      );
+      appDef.legacy.scriptingSource = appDef.legacy.scriptingSource.replace(
+        'movie_post_poll_make_array',
+        'movie_post_poll'
+      );
+      const compiledApp = schemaTools.prepareApp(appDef);
+      const app = createApp(appDefWithAuth);
+
+      const input = createTestInput(
+        compiledApp,
+        'triggers.movie.operation.perform'
+      );
+      input.bundle.inputData = {
+        things: ['eyedrops', 'cyclops', 'ipod']
+      };
+      return app(input).then(output => {
+        const req = output.results[0];
+        req.args.should.deepEqual({
+          things: ['eyedrops,cyclops,ipod']
+        });
+        req.url.should.equal(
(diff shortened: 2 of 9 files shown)


eliangcs:
@xavdid I'm not sure why [local integration test](https://travis-ci.org/github/zapier/zapier-platform/builds/672012685#L859) is failing. I'll keep working on it tomorrow. But it'd great if you know what happened offhand.

packages/legacy-scripting-runner/index.js
```diff
@@ -630,6 +630,16 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
       }
 
       request.headers = cleanHeaders(request.headers);
+      request.serializeValueForCurlies = value => {
+        if (Array.isArray(value)) {
+          return value.join(',');
+        } else if (_.isPlainObject(value)) {
+          // Not sure if anyone would ever expect a '[object Object]',
```
xavdid, line 637: ¯\\\_(ツ)_/¯ https://xkcd.com/1172/

packages/schema/lib/schemas/RequestSchema.js
```diff
@@ -74,6 +75,11 @@ module.exports = makeSchema(
           }
         },
         additionalProperties: false
+      },
+      serializeValueForCurlies: {
+        description:
+          'A function to customize how to serialize a value for curlies `{{var}}`` in the request object. By default, when this is unspecified, the request client only replaces curlies where variables are strings, and would throw an error for non-strings. The function should accepts a single argument as the value to be serialized and return the string representation of the argument.',
```
xavdid, line 81: there's an extra backtick on your `{{var}}` making the rest of this line highlight weird in markdown

packages/legacy-scripting-runner/test/integration-test.js
```diff
@@ -756,6 +756,37 @@ describe('Integration Test', () => {
...
+      const appDef = _.cloneDeep(appDefinition);
+      appDef.legacy.scriptingSource = appDef.legacy.scriptingSource.replace(
+        'movie_pre_poll_array_curlies',
+        'movie_pre_poll'
+      );
+      appDef.legacy.scriptingSource = appDef.legacy.scriptingSource.replace(
+        'movie_post_poll_make_array',
+        'movie_post_poll'
+      );
+      const compiledApp = schemaTools.prepareApp(appDef);
```
xavdid, line 769: not a change here, but most of these tests go through these same setup steps. I wonder if it's worth moving setup into a callable function

xavdid:
The test is failing because you need to declare `RequestSchema` dependent on `FunctionSchema`.

Add `FunctionSchema` to this line:

https://github.com/zapier/zapier-platform/blob/10d1d71fc7088376f3c4462b939dabceb59492ef/packages/schema/lib/schemas/RequestSchema.js#L81

packages/legacy-scripting-runner/test/integration-test.js
eliangcs: Yep, that's something I wanted to do but it's never been a priority (or don't have much motivation).

PR #193, opened 2020-04-08
BREAKING CHANGE(core): Make throwForStatus an opt-out, also for code operations

FokkeZB:
Replaced by #210


packages/core/src/execute.js
```diff
@@ -9,31 +9,27 @@ const createJSONtool = require('./tools/create-json-tool');
...
 const ZapierPromise = require('./tools/promise');
 const { FORM_TYPE } = require('./tools/http');
 
 const constants = require('./constants');
 
 const executeHttpRequest = (input, options) => {
   options = _.extend({}, options, constants.REQUEST_OBJECT_SHORTHAND_OPTIONS);
-  return input.z
-    .request(options)
-    .then(throwForStatus)
```
FokkeZB, line 22: No need to append `throwForStatus` middleware for configured requests as all requests now default to it.

packages/core/src/http-middlewares/after/throw-for-stale-auth.js
```diff
@@ -1,17 +0,0 @@
-'use strict';
-
-const errors = require('../../errors');
-
-/**
- * The server-side will raise `RefreshAuthError` when `autoRefresh === true`.
- * Once we always run `throwForStatus` or a custom `afterResponse`, we can drop `throwForStaleAuth`.
- */
-const throwForStaleAuth = resp => {
```
FokkeZB, line 9: No more need for this as `throwForStatus` will throw for 401 and on the server-side we check if that should result in a refresh for the app's definition.

packages/core/src/tools/create-app-request-client.js
```diff
@@ -53,17 +53,14 @@ const createAppRequestClient = (input, options) => {
 
   const httpOriginalAfters = [prepareResponse, logResponse];
 
-  if (app.authentication) {
-    if (
-      app.authentication.type === 'session' ||
-      (app.authentication.type === 'oauth2' &&
-        _.get(app, 'authentication.oauth2Config.autoRefresh'))
-    ) {
-      httpOriginalAfters.push(throwForStaleAuth);
```
FokkeZB, line 62: No more need for this as we now default `afterResponse` to `[throwForStatus]`, which will throw for 401 and on the server-side we check if that should result in a refresh for the app's definition.

packages/core/test/userapp/index.js
```diff
@@ -507,20 +507,18 @@ const addRequestHeader = (request, z, bundle) => {
...
+    }
   }
 
-  const data = JSON.parse(response.content);
-  const error = data.args.error;
-  if (response.status === 200 && error) {
-    response.status = 500;
-  }
+  // defining `afterResponse` makes use responsible for calling `throwForStatus()`
+  response.throwForStatus();
```
FokkeZB, line 520: Our test app was a good example of an app that now needs to to take its responsibility for error handling.

example-apps/middleware/index.js
```diff
@@ -15,14 +15,26 @@ const addSortingParams = (request /*, z */) => {
 };
 
 // HTTP after middleware that checks for errors in the response.
-const checkForErrors = (response, z) => {
-  // If we get a bad status code, throw an error. This will halt the zap.
-  if (response.status >= 300) {
-    throw new z.errors.HaltedError(
-      `Unexpected status code ${response.status} from ${response.request.url}`
+// When you don't use `afterResponse` it default to `[throwForStatus]`.
+const handleErrors = (response, z) => {
```
FokkeZB, line 19: Updated the middleware example to show:

1. Using `afterResponse` makes you responsible for error response handling.
2. How to throw for statuses that `throwForStatus` won't (correctly).
3. How to prevent `throwForStatus` from throwing.
4. That you should always call `throwForStatus`.

example-apps/oauth1-trello/authentication.js
```diff
@@ -17,7 +17,6 @@ const getRequestToken = async (z, bundle) => {
       oauth_callback: bundle.inputData.redirect_uri
     }
   });
-  response.throwForStatus();
```
FokkeZB, line 20: No more need to call `throwForStatus` in perform-like functions.

example-apps/session-auth/index.js
```diff
@@ -17,6 +17,8 @@ const sessionRefreshIf401 = (response, z, bundle) => {
       throw new z.errors.RefreshAuthError('Session key needs refreshing.');
     }
   }
+  // When you define an `afterResponse`, make sure one of them calls `throwForStatus`
+  response.throwForStatus();
```
FokkeZB, line 21: Updated examples that use `afterResponse` to show that they are now responsible for calling `throwForStatus`. Of course they could also do this in their perform-like functions but this is the preferred way.

packages/cli/README-source.md
```diff
@@ -994,6 +994,10 @@ Middleware functions can be asynchronous - just return a promise from the middle
 
 The second argument for middleware is the `z` object, but it does *not* include `z.request()` as using that would easily create infinite loops.
 
+#### Error Response Handling
+
+Since v10 `afterResponse` defaults to one that calls `response.throwForStatus()`. This makes you responsible for error response  handling when you do use `afterResponse`. Like in the above example, first handle any cases that `throwForStatus` shouldn't or doesn't (correctly) throw for.
```
FokkeZB, line 999: Documented how using `afterResponse` makes you responsible for error handling.

packages/cli/README-source.md
```diff
@@ -1215,7 +1219,11 @@ have incomprehensible messages for end users, or possibly go uncaught.
 
 Zapier provides a couple tools to help with error handling. First is the `afterResponse`
 middleware ([docs](#using-http-middleware)), which provides a hook for processing
-all responses from HTTP calls. The other tool is the collection of errors in
+all responses from HTTP calls. Second is `response.throwForStatus()`
+([docs](#http-response-object)), which throws an error if the response status
+indicates an error. Since v10, `afterResponse` defaults to calling `throwForStatus`
+and using `afterResponse` makes you responsible for error response handling
+([docs](#error-response-handling)). The last tool is the collection of errors in
```
FokkeZB, line 1226: Updated the general error handling topic to explain the role of the default `afterResponse` and how using it makes you responsible.

packages/cli/scaffold/create.template.js
```diff
@@ -9,9 +9,8 @@ const perform = async (z, bundle) => {
       name: bundle.inputData.name
     }
   });
-  response.throwForStatus();
   // this should return a single object
-  return z.JSON.parse(response.content);
```
FokkeZB, line 14: While I was here, switched to `response.json` as it does the same thing.

packages/cli/snippets/middleware.js
```diff
@@ -10,6 +10,26 @@ const mustBe200 = (response, z, bundle) => {
   return response;
 };
 
+const handleErrors = (response, z) => {
```
FokkeZB, line 13: Updated the middleware snippet to show:

1. Using `afterResponse` makes you responsible for error response handling.
2. How to throw for statuses that `throwForStatus` won't (correctly).
3. How to prevent `throwForStatus` from throwing.
4. That you should always call `throwForStatus`.

packages/cli/snippets/session-auth.js
```diff
@@ -56,19 +56,9 @@ const includeSessionKeyHeader = (request, z, bundle) => {
...
-    }
-  }
-  return response;
-};
-
 const App = {
   // ...
   authentication: authentication,
-  beforeRequest: [includeSessionKeyHeader],
-  afterResponse: [sessionRefreshIf401]
```
FokkeZB, line 72: Now that we default to call `throwForStatus` and already on the server-side refresh if the status is 401 and the auth type session or OAuth2 w/refresh, there's no need for this anymore.

packages/core/src/tools/create-file-stasher.js
```diff
@@ -151,7 +151,6 @@ const createFileStasher = input => {
             };
 
             if (isStreamed) {
-              maybeResponse.throwForStatus();
```
FokkeZB, line 154: I'm not 💯 on this one, but I _think_ any `maybeResponse` will also have gone through `afterResponse`.

packages/core/test/create-request-client.js
```diff
@@ -270,7 +271,21 @@ describe('request client', () => {
       .catch(done);
   });
 
-  it('should run any afterResponse functions', done => {
+  it('should default afterResponse to [throwForStatus]', () => {
```
FokkeZB, line 274: Test for defaulting `afterResponse`.

packages/core/test/create-request-client.js
```diff
@@ -284,18 +299,10 @@ describe('request client', () => {
       testLogger
     );
     const request = createAppRequestClient(inputWithAfterMiddleware);
-    request({ url: 'https://httpbin.org/get' })
-      .then(responseBefore => {
-        const response = JSON.parse(JSON.stringify(responseBefore));
+    const response = await request({ url: 'https://httpbin.org/status/400' });
```
FokkeZB, line 302: Updated test for not using the default if you define a custom `afterResponse` (with no `throwForStatus`)

packages/core/src/tools/create-app-request-client.js
```diff
@@ -53,17 +53,14 @@ const createAppRequestClient = (input, options) => {
...
-    }
-  }
+  // default `afterResponse` to `[throwForStatus]` to force developers
+  // to define their own if it doesn't throw as expected.
+  const afterResponse =
+    Array.isArray(app.afterResponse) && app.afterResponse.length > 0
+      ? app.afterResponse
+      : [throwForStatus];
```
rebecca-ghazali, line 61: Looks like this is the core change!

FokkeZB: Correct!

packages/legacy-scripting-runner/middleware-factory.js
```diff
@@ -203,18 +203,21 @@ const createBeforeRequest = app => {
 
 const createAfterResponse = app => {
   const authType = _.get(app, 'authentication.type');
+  const autoRefresh =
+    _.get(app, 'legacy.authentication.oauth2Config.autoRefresh') ||
+    _.get(app, 'authentication.oauth2Config.autoRefresh');
```
FokkeZB, line 208: @eliangcs I'm not sure which one I need to look for here?

packages/legacy-scripting-runner/middleware-factory.js
```diff
@@ -203,18 +203,21 @@ const createBeforeRequest = app => {
...
     }
     return response;
   };
 
   let afterResponse;
 
-  if (authType === 'session') {
-    afterResponse = sessionAuthCheckResponse;
+  if (authType === 'session' || (authType === 'oauth2' && autoRefresh)) {
+    afterResponse = throwForStaleAuth;
```
FokkeZB, line 220: For some reason we did simulate `throwForStaleAuth` for session auth, but but OAuth2 w/refresh we relied on `core` taking care of it (see comment in `example-app/oauth2.js`). After this PR `core` won't anymore, so to keep the same behavior for legacy-scripting runs we need to simulate it for OAuth2 as well.

packages/legacy-scripting-runner/test/example-app/api-key-auth.js
```diff
@@ -41,5 +45,8 @@ module.exports = {
       source: beforeRequestSource,
       args: ['request', 'z', 'bundle']
     }
+  ],
+  afterResponse: [
+    { source: afterResponseSource, args: ['response', 'z', 'bundle'] }
```
FokkeZB, line 50: To not default `afterResponse` to `[throwForStatus]` every example app needs to use `legacyScripting.afterResponse` which defaults to a no-op like we will for visual builder apps.

packages/legacy-scripting-runner/test/example-app/oauth2.js
```diff
@@ -63,8 +67,8 @@ module.exports = {
   },
   beforeRequest: [
     { source: beforeRequestSource, args: ['request', 'z', 'bundle'] }
+  ],
+  afterResponse: [
+    { source: afterResponseSource, args: ['response', 'z', 'bundle'] }
   ]
-
-  // We don't need afterResponse to refresh auth as core appends one
-  // automatically when autoRefresh is true
```
FokkeZB, line 69: Not the case anymore - see comment

packages/legacy-scripting-runner/test/example-app/index.js
```diff
@@ -1171,6 +1171,19 @@ const App = {
       source: "return z.legacyScripting.run(bundle, 'hydrate.file');"
     }
   },
+  // This breaks `applyBeforeMiddleware` which it looks like gets the original definition without `findSourceRequireFunctions` applied?!
+  // beforeRequest: [
+  //   {
+  //     source: 'return z.legacyScripting.beforeRequest(request, z, bundle);',
+  //     args: ['request', 'z', 'bundle']
+  //   }
+  // ],
```
FokkeZB, line 1180: @eliangcs it looks like I ran into a bug here. When I enable this, the `hook trigger > `KEY_catch_hook` integration tests fail because the `source` here hasn't been replaced by a function via `findSourceRequireFunctions` when it gets called by `applyBeforeMiddleware`.

Any idea? We don't strictly need `beforeRequest` here (just `afterResponse` - and that works fine), but it may be good to check if this is a test-setup issue or an actual bug in the legacy scripting runner.

packages/legacy-scripting-runner/test/example-app/index.js
```diff
@@ -1171,6 +1171,19 @@ const App = {
...
+  afterResponse: [
+    {
+      source: 'return z.legacyScripting.afterResponse(response, z, bundle);',
+      args: ['response', 'z', 'bundle']
+    }
+  ],
```
FokkeZB, line 1186: This is that the integration tests need to not default `afterResponse` to `[throwForStatus]`.

packages/core/src/execute.js
casshill: when does executeHttpRequest actually get called? I wonder if this is only for a form mode app? Just wondering because I don't think I've seen this get called for code CLI apps or in code mode visual buildler apps?

eliangcs: We call `executeHttpRequest` when a "request shorthand" is used. Namely, when you write something like this for a `perform` or auth `test`:

```
{
  method: 'POST',
  url: '...',
  body: {
    foo: '{{bundle.inputData.foo}}'
  }
}
```

This is also the underlying representation for visual builder's form mode.

eliangcs:
As [discussed](https://zapier.slack.com/archives/C012L990N8L/p1588219938033300), I'm proposing to **always** add the `throwForStatus` middleware, even if an app has its own `afterResponse`. We also add a `req.throwForStatus` boolean as a `z.request` option, true by default. When a developer don't want `z.request` to throw for status, they can set `req.throwForStatus` to false:

```js
z.request({
   // ...
   throwForStatus: false
});
```

So that's the interface we expose to developers. And here are the changes I think we can make to achieve this goal:

1. Add `throwForStatus` middleware right after `app.afterResponse`: https://github.com/zapier/zapier-platform/blob/4878770112684c23d175ecfa56076b1dbe8da9a7/packages/core/src/tools/create-app-request-client.js#L66

2. In [`prepareRequest`](https://github.com/zapier/zapier-platform/blob/4878770112684c23d175ecfa56076b1dbe8da9a7/packages/core/src/http-middlewares/before/prepare-request.js#L105-L114), add `req.throwForStatus` boolean and default it to true

    ```js
    const prepareRequest = function(req) {
     // ...
      req = _.defaults(req, {
        // ...
        throwForStatus: true, // <---
        // ...
      });
    // ...
    ```

3. Modify [`throwForStatus`](https://github.com/zapier/zapier-platform/blob/4878770112684c23d175ecfa56076b1dbe8da9a7/packages/core/src/http-middlewares/after/throw-for-status.js) middleware to honor the `req.throwForStatus` boolean:

    ```js
    const throwForStatus = resp => {
      if (resp.status >= 400 && resp.request.throwForStatus) {
        throw new errors.ResponseError(resp);
      }
      return resp;
    };
    ```

To make sure we don't break VB apps when upgrading them, we'll need to the app definition so they all have a `beforeRequest` that defaults `req.throwForStatus` to false for the requests in code mode. This is a bit tricky but doable.

1. For this new `beforeRequest` to tell if the request is in form mode (called "shorthand" internally), we set a flag `req.shorthand` before executing the request. This is where we execute a shorthand request: https://github.com/zapier/zapier-platform/blob/4878770112684c23d175ecfa56076b1dbe8da9a7/packages/core/src/execute.js#L18-L29

   We can set `req.shorthand` by modifying the [`REQUEST_OBJECT_SHORTHAND_OPTIONS`](https://github.com/zapier/zapier-platform/blob/4878770112684c23d175ecfa56076b1dbe8da9a7/packages/core/src/constants.js#L21) constant into:

    ```js
    const REQUEST_OBJECT_SHORTHAND_OPTIONS = { replace: true, shorthand: true };
    ```
    
2. When upgrading VB apps to core 10, we add this `beforeRequest`:

    ```js
    {
      beforeRequest: [
        // ... existing middlewares
        {
          args: ['req', 'z', 'bundle'],
          source: 'if (!req.shorthand) { req.throwForStatus = false; } return req;'
        }
      ]
    }
    ```
    
    This middleware would disable `throwForStatus` for code-mode requests while enabling it for form-mode requests.

FokkeZB:
Loving this @eliangcs!

I would suggest that we `request({ throwForStatus: false })` in `prepareResponse` gets copied to `response` so that `afterResponse` middleware can also set it to `false` to prevent `throwForStatus` from throwing for a response.

Of course they could set `response.request.throwForStatus` but I don't like such a nested API, also because we already have `response.throwForStatus`. For those reasons, perhaps we should flip it and use `skipThrowForStatus` and in `beforeRequest` middleware set `response.skipThrowForStatus = request.skipThrowForStatus`?

We could then also use `afterResponse` instead of `beforeRequest` to handle the VB app migrations, as that would also allow us to get rid of `throwForStaleAuth` (since we now default to `throwForStatus, which also throws for 401) by using:



I've proposed this option in https://docs.google.com/document/d/1j1O2_88dirRvciAEO05axE_B1J2tCfKM8qsGRshDepc/edit#heading=h.e2nf42jcxd7n

packages/core/src/execute.js
casshill: hmmm @FokkeZB sounds like this could also be used in a CLI app then, not just visual builder? Not sure if we considered that case.

eliangcs:
@FokkeZB yes, `response.skipThrowForStatus` (instead of `response.request.throwForStatus`) looks cleaner! Let's go for your proposal.

packages/core/src/execute.js
FokkeZB: Yes, I never realized before, but preventing `throwForStatus` from throwing or handling unparsable content isn't something you can currently do (without the hacky overwriting of status and content) in CLI apps that have at least one shorthand/form-mode request either, as there as well we always run `throwForStatus` and `JSON.parse(response.content)`.

So this indeed will benefit CLI developers as well.

FokkeZB:
I've made updates for the solution to be conform https://docs.google.com/document/d/1j1O2_88dirRvciAEO05axE_B1J2tCfKM8qsGRshDepc/edit# and for clarity sake pushed this as a new PR: #210

PR #194, opened 2020-04-09
feat(legacy-scripting-runner): support for unbounded curlies

eliangcs:
Addresses [PDE-1467](https://zapierorg.atlassian.net/browse/PDE-1467).

When converting a WB app, we replace "unbounded" curlies `{{var}}` in URLs with "bounded" curlies, which are either:

- `{{bundle.authData.var}}`
- `{{bundle.inputData.var}}`
- `{{process.env.var}}`

Which one to use is decided based on the field definition. That is, if `var` is in auth fields, then we replace `var` with `bundle.authData.var`.

The idea was to be explicit about variable referencing. But unfortunately, being explicit requires some guessing and is not always reliable. This PR adds support for unbounded curlies, so `{{var}}` in URLs will be still `{{var}}`. legacy-scripting-runner will do the implicit variable referencing as WB does.

Existing converted apps that are still using bounded curlies like `{{bundle.inputData.var}}` will continue to work. The change is backward compatible.

This PR works with https://github.com/zapier/zapier/pull/38762.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -5,9 +5,10 @@ const FormData = require('form-data');
 const flatten = require('flat');
 
 const {
-  createBundleBank,
   recurseReplaceBank
 } = require('zapier-platform-core/src/tools/cleaner');
+
+const { flattenPaths } = require('zapier-platform-core/src/tools/data');
 
 const createInternalRequestClient = input => {
   const addQueryParams = require('zapier-platform-core/src/http-middlewares/before/add-query-params');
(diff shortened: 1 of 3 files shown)


packages/legacy-scripting-runner/index.js
```diff
@@ -207,9 +212,44 @@ const parseFinalResult = async (result, event) => {
   return result;
 };
 
+const serializeValueForCurlies = value => {
+  if (typeof value === 'boolean') {
+    return value ? 'True' : 'False';
```
xavdid, line 217: I might do these as lowercase. Who knows how the external server reads them, but since that's how they appear in JSON, lowercase feels safer to me.

eliangcs: This is actually how WB behaves. If you write `https://example.com?active={{is_active}}` for a trigger/action URL and `is_active` is a boolean, the boolean going to be serialized to `True` or `False`.

PR #195, opened 2020-04-10
feat(core,legacy-scripting-runner): allow GET requests to have body

eliangcs:
Adds a hidden option `allowGetBody` to `z.request()`. When `allowGetBody` is true, `z.request` won't delete the body for the GET request. This is to add support for WB's ability to send GET requests with a body.

This has to be done with a nasty hack because node-fetch explicitly [forbids](https://github.com/node-fetch/node-fetch/blob/v2.6.0/src/request.js#L75-L78) us to send a GET request with a body (see also https://github.com/whatwg/fetch/issues/551). Not an ideal solution, but it gets the job done.

### DIFF
--- a/packages/core/src/http-middlewares/before/prepare-request.js
+++ b/packages/core/src/http-middlewares/before/prepare-request.js
@@ -44,7 +44,7 @@ const coerceBody = req => {
   const contentType = getContentType(req.headers || {});
 
   // No need for body on get
-  if (req.method === 'GET') {
+  if (req.method === 'GET' && !req.allowGetBody) {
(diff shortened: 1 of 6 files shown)


eliangcs:
@xavdid I think this "feature" can still be useful for CLI apps. We probably don't want to encourage this usage—that's why I made it a "hidden" operation. But there are APIs like [ElasticSearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html) that requires users to GET with body. So we should at least provide an option for developers to do it.

PR #198, opened 2020-04-10
docs(core): update examples to demonstrate z.errors.Error

FokkeZB:
Updates example apps and snippets to demonstrate (and with that encourage) the use of `z.errors.Error` over `Error`.

**JIRA:** `SXP-1624` [Replace use of Error by z.errors.Error in examples and docs](https://zapierorg.atlassian.net/browse/SXP-1624)

### DIFF
--- a/example-apps/babel/src/authentication.js
+++ b/example-apps/babel/src/authentication.js
@@ -9,7 +9,11 @@ const test = async (z /*, bundle */) => {
   // This method can return any truthy value to indicate the credentials are valid.
   // Raise an error to show
   if (response.status === 401) {
-    throw new Error('The username and/or password you supplied is incorrect');
+    throw new z.errors.Error(
+      'The username and/or password you supplied is incorrect',
+      'AuthenticationError',
+      response.status
+    );
   }
   return response;
 };
--- a/example-apps/onedrive/utils.js
+++ b/example-apps/onedrive/utils.js
@@ -8,7 +8,7 @@ const contentDisposition = require('content-disposition')
 // on the OneDrive API. We intentionally don't set it up as an `afterResposne`
 // handler because not *all* calls need it (i.e. the auth test and file create),
 // so we break it out and share the code this way instead.
-const parseResponse = (type, response) => {
+const parseResponse = (z, type, response) => {
   let results = []
 
   if (response.status >= 200 && response.status < 300) {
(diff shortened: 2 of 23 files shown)


example-apps/babel/src/authentication.js
```diff
@@ -9,7 +9,11 @@ const test = async (z /*, bundle */) => {
   // This method can return any truthy value to indicate the credentials are valid.
   // Raise an error to show
   if (response.status === 401) {
-    throw new Error('The username and/or password you supplied is incorrect');
+    throw new z.errors.Error(
+      'The username and/or password you supplied is incorrect',
+      'AuthenticationError',
```
FokkeZB, line 14: The second argument can be any code and as long as we're the only ones able to create troubleshooting docs, this isn't of much direct use to the developer.

However, the idea is/was that we'll grow a set of standard codes that we have generic troubleshooting docs on and ultimately, we'd to list the most common error codes we see for an app in its dashboard and maybe even allow partners to write their own troubleshooting docs for them.

example-apps/onedrive/utils.js
```diff
@@ -8,7 +8,7 @@ const contentDisposition = require('content-disposition')
 // on the OneDrive API. We intentionally don't set it up as an `afterResposne`
 // handler because not *all* calls need it (i.e. the auth test and file create),
 // so we break it out and share the code this way instead.
-const parseResponse = (type, response) => {
+const parseResponse = (z, type, response) => {
```
FokkeZB, line 11: Notice that the signature of this function has changed. All usage has been updated.

packages/cli/README.md
```diff
@@ -1932,9 +1932,7 @@ const listExample = (z, bundle) => {
   return z
     .request('https://example.com/api/v2/recipes.json', customHttpOptions)
     .then(response => {
-      if (response.status >= 300) {
-        throw new Error(`Unexpected status code ${response.status}`);
-      }
```
FokkeZB, line 1937: We should encourage the use of `throwForStatus` as much as we can (at least until we default its middleware)

FokkeZB: 🤦 I already updated the snippets this was generated from, so I'll revert this and regenerate.

FokkeZB: Oh, that actually results in the same diff - which makes sense as on pre-commit like you said `README.md` is updated with the updated snippets.

PR #200, opened 2020-04-13
docs(core) Use `response.json` where possible

FokkeZB:
I originally included some of these in #193 but figured it was better done in a separate PR.

AFAIK unless you set `raw:true` you can use `response.json` instead of `z.JSON.parse(response.content)` so I guess we want to promote that more simple shortcut as much as we can.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -566,7 +566,7 @@ const getSessionKey = (z, bundle) => {
       throw new Error('The username/password you supplied is invalid');
     }
     return {
-      sessionKey: z.JSON.parse(response.content).sessionKey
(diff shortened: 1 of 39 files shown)


packages/cli/README.md
```diff
@@ -2030,15 +2030,10 @@ const mustBe200 = (response, z, bundle) => {
   return response;
 };
 
-const autoParseJson = (response, z, bundle) => {
-  response.json = z.JSON.parse(response.content);
-  return response;
-};
```
FokkeZB, line 2036: This example middleware was just overriding the already set `response.json` so IMO did not have much value as example and fortunately we have another `afterResponse` example.

PR #202, opened 2020-04-23
improvement(cli): Compress zip files more

xavdid:
Adds some compression when creating `build.zip`. 

Using an app w/ `lodash` and `moment` as dependencies, the zip size went from `1.6M` -> `409K`, so that's promising!

I also tested with an app pushed to prod - everything worked normally. I didn't notice any difference in the app build time. 

Test suite was the exact same speed.


xavdid:
sure, done!

PR #205, opened 2020-04-28
improvement(cli) Simplify and improve auth file generation

xavdid:
Replaces hard-coded auth files w/ generated ones.


packages/cli/src/utils/init.js
```diff
@@ -28,6 +30,70 @@ const initApp = async (path, createFunc) => {
...
+  const suppliedItem =
+    {
+      custom: 'API Key',
+      oauth2: 'access token'
+    }[authType] || 'username and/or password';
+
+  const output = `
+    ${testAuth(authType)}
+
+    const handleBadResponses = (response, z, bundle) => {
```
xavdid, line 65: I don't love that these are stored in strings (there's no syntax highlighting when writing, there's no pretttier), but i'm not sure a better way to do it. We want to build the code in pieces (so being able to drop in conditional strings inline is great), so this is probably the best way to do it. 

The big alternative is to lean on `jscodeshift`, but i'm not sure that'll be much easier to write and iterate with. It's also tougher to read. We could probably reduce code re-use, but it's dense. 

The bottom line is that this'll get prettified before it goes to the user, so it's ok if it's ugly in here as long as its syntactically correct.

packages/cli/src/utils/init.js
```diff
@@ -28,6 +30,70 @@ const initApp = async (path, createFunc) => {
...
+
+        // The test method allows Zapier to verify that the credentials a user provides are valid.
+        // We'll execute this method whenver a user connects their account for the first time.
+        test,
+
+        // this can be a "{{string}}" or a function
+        connectionLabel: '{{bundle.inputData.username}}'
+      },
+
+      befores: [],
```
xavdid, line 89: These middleware being picked up depend on an `index.js` file like this:

```js
const {
  config: authentication,
  befores = [],
  afters = []
} = require('./authentication')

const App = {
  // This is just shorthand to reference the installed dependencies you have. Zapier will
  // need to know these before we can upload
  version: require('./package.json').version,
  platformVersion: require('zapier-platform-core').version,

  authentication,

  beforeRequest: [...befores],

  afterResponse: [...afters],

  resources: {},

  // If you want your trigger to show up, you better include it here!
  triggers: {},

  // If you want your searches to show up, you better include it here!
  searches: {},

  // If you want your creates to show up, you better include it here!
  creates: {}
}

// Finally, export the app.
module.exports = App
```

packages/cli/src/utils/init.js
```diff
@@ -28,6 +30,70 @@ const initApp = async (path, createFunc) => {
...
+const testAuth = authType => {
+  // todo: oauth1 trello url?
+  const testUrl =
+    authType === 'digest'
+      ? 'https://httpbin.org/digest-auth/auth/myuser/mypass'
+      : 'https://auth-json-server.zapier-staging.com/me';
+
+  return `
+      const test = (z, bundle) => {
+        return z.request({
```
xavdid, line 42: the tests don't do the check that they currently do - I've confirmed (both in local unit tests and connecting an auth on the site) that throwing for 401s works just as well in middleware as it does in the `testAuth` function. Not sure why an auth would only want to check for bad creds in the test function (and not always), but it seems fine here. 

If the test function throws, then that error will supersede any middleware.

packages/cli/src/utils/init.js
```diff
@@ -28,6 +30,70 @@ const initApp = async (path, createFunc) => {
   await removeDir(tempAppDir);
 };
 
+const testAuth = authType => {
+  // todo: oauth1 trello url?
```
eliangcs, line 34: Yeah, we can use Trello for OAuth1. Or, we can add OAuth1 support to https://auth-json-server.zapier-staging.com and use that. But I'd just use what we have (Trello) for now.

packages/cli/src/utils/init.js
eliangcs: > I don't love that these are stored in strings

I also feel that. It's a little hard to maintain this way. Can we split these code snippets into template files? It's kind of like what we're already doing with the `scaffold` command, where we keep template files in the [`scaffold`](https://github.com/zapier/zapier-platform/tree/4878770112684c23d175ecfa56076b1dbe8da9a7/packages/cli/scaffold) folder.

I also noticed [previously](https://github.com/zapier/zapier-platform/pull/74) you created an [`examples`](https://github.com/zapier/zapier-platform/tree/4878770112684c23d175ecfa56076b1dbe8da9a7/examples/auth) folder. I forget what was the intention there. Are they going to be used as templates for `init` or are they just "examples"?

packages/cli/src/utils/init.js
```diff
@@ -28,6 +30,70 @@ const initApp = async (path, createFunc) => {
...
+  const testUrl =
+    authType === 'digest'
+      ? 'https://httpbin.org/digest-auth/auth/myuser/mypass'
+      : 'https://auth-json-server.zapier-staging.com/me';
+
+  return `
+      const test = (z, bundle) => {
+        return z.request({
+          url: '${testUrl}',
+          ${authType === 'oauth2' ? inspect({}) : ''}
```
eliangcs, line 44: I'm not sure what it is for. Is it intended or just a WIP thing?

xavdid: yep! that lets us print objects as strings

packages/cli/src/utils/init.js
xavdid: The examples folder was from a PR distilling every `example-apps` folder into a slim set of files. We unfortunately can't delete `example-apps` because older CLI versions use it on master for the `init` command. 

We can have them in template files for lodash. that'll get us syntax highlighting, but not prettier (since the `<%= %>` tags make it invalid again). If we go this route, we'll do a lot of reading of files and treating them as strings. Might be better, i'll play with it today. 

I think either way, if the result is run through `prettier`, it'll ensure all snippets are syntactically correct and formatted nicely. as long as we have test coverage, we'll at least know if something is super broken.

xavdid:
We would still use `yeoman` - it would be responsible for pulling all the files and putting them in the right place (we'd have a single copy of a basic `package.json`, `index.js`, etc). Instead of having a set of auth files that have a lot of overlap, `yeoman` would wall this function to generate one as needed.

xavdid:
so I tried what I did yesterday with both lodash templates (which I still don't love; they don't feel materially easier to maintain than big strings) and something a little out of left field.

Strings are easier to work with here and very flexible for all of the conditionals. I tried something _declarative_, inspired by some of the AST work I had done before: a bunch of tiny functions that return strings and can be easily combined to output valid js. 

With this approach, we could make very reusable chunks that combine into the auth files we want to write. It's built on functions like `ifStatement` and `returnStatement`, but we can also make `zRequest` and `describeBlock` as we get higher up the chain. I find it easy to reason about at a glance without having to worry about the exact syntax. Prettier takes care of everything at the end and then we're all set!

I think of the three (1. `strings`, 2. `templates`, 3. `composable functions`) this is my favorite approach right now. I found it a little tough to reason about the templates since we were going to need multiples and have a lot of conditional logic. 

I'd love to hear what your thoughts on it are though! It definitely might be over-engineered for what we need. 😁

eliangcs:
@xavdid thanks for trying the template route!

> It's built on functions like `ifStatement` and `returnStatement`, but we can also make `zRequest` and `describeBlock` as we get higher up the chain.

This sounds much like React components! So from small building blocks, we can make bigger building blocks, and generate a complete `.js` file eventually. Yeah, I think that's a good idea.

packages/cli/src/utils/files.js
```diff
@@ -36,6 +36,11 @@ const readFile = (fileName, errMsg) => {
   );
 };
 
+const readFileStr = async (fileName, errMsg) => {
```
xavdid, line 39: I didn't end up using this, but long term I'd like to adopt this anyway, so i'll go ahead and leave ¯\\\_(ツ)_/¯

xavdid:
ok! we've got all of our 6 auth types generating complete files. 

assuming this is 👍for you, we can merge it, you can pull it into the `yo` branch, and it should drop right in.

eliangcs:
@xavdid awesome! I've merged this branch to the `yo` branch. We can keep working it on #206. Closing this PR.

PR #209, opened 2020-04-30
fix(cli): Allow for different changelog version formats

codebycaleb:
https://keepachangelog.com/en/1.0.0/ uses a format like so:

but our changelog utils only handled

So this commit just loosens up the restrictions a bit. :)


eliangcs:
@xavdid a more structural changelog format is good, but I think it'd be nice if we keep supporting the markdown format.

PR #211, opened 2020-05-02
BREAKING CHANGE(core): Introduce response.data with support for form-urlencoded and custom parsing

FokkeZB:
# Definitions:

- Code-mode: [`FunctionSchema`](https://github.com/zapier/zapier-platform/blob/master/packages/schema/docs/build/schema.md#functionschema)
- Form-mode: [`RequestSchema`](https://github.com/zapier/zapier-platform/blob/master/packages/schema/docs/build/schema.md#requestschema)

# Commits

I've split the work up in 3 commits:

- Changes: https://github.com/zapier/zapier-platform/pull/211/commits/44f509a763991bfa132da7536229fb0303c0c458
- Tests: https://github.com/zapier/zapier-platform/pull/211/commits/5bb92a242dc7f8493c02cc39b56e12b9492e64f6
- Documentation & Examples: https://github.com/zapier/zapier-platform/pull/211/commits/439a5963983332181cbde31c8cf9606e6370891d

# Changes

- `prepareContentResponse` now parses `.content` to `.data` if its either `application/x-www-form-urlencoded` or can be parsed as JSON.
- `prepareContentResponse` still sets `.json` but it should be considered deprecated.
- `executeHttpRequest` now uses `.data` instead of (again) parsing `.content` as `application/x-www-form-urlencoded` or JSON.
- `executeHttpRequest` now throws a helpful error instead of `SyntaxError` if `.data` is not set and thus not `application/x-www-form-urlencoded`, JSON or set in `afterResponse`.

See the first commit.

# Benefits

- Code-mode no longer need to parse `application/x-www-form-urlencoded`, just like form-mode requests.
- Developers don't need to `JSON.stringify()` their custom-parsed `.content` back to the same property for form-mode to work.
- The type-agnostic `.data` property allows us to later add support for types like XML.

See the below _Examples_.

# Examples

These examples all use `afterResponse`. The code-mode ones could also handle this within the function of course, but this makes it easier to compare between code and form-mode.

## Form-mode request with form response

Already and still works out of the box. It just now relies on `response.data` set by `prepareResponse` instead of parsing `response.content` just before returning it.

This means it may be a breaking change for CLI apps that were doing custom parsing in `afterResponse` as the *Form-mode request with XML response* example.

### Before

```node
const querystring = require('querystring');

const afterResponse = (response) => {
  if (response.headers.get('content-type') === 'application/x-www-form-urlencoded') {
    response.json = querystring.parse(response.content);
  }
}
```

### After

```node
z.request('https://example.com/').then(response => response.data);
```

## Code-mode request with JSON response

You can still use `response.json` but we encourage switching to `response.data`.

### Before

```node
z.request('https://example.com/').then(response => response.json);
```

### After

```node
z.request('https://example.com/').then(response => response.data);
```

## Form-mode request with XML response

Note that both before and after currently are only possible in CLI apps, but once https://github.com/zapier/zapier/pull/39899 makes the Middleware tab public for all, this can also be used in visual builder apps.

### Before

As we defaulted to parse `response.content` as JSON:

```node
const parser = require('fast-xml-parser');

const afterResponse = (response) => {
  if ('xml' in response.headers.get('content-type')) {
    const parsed = parser(response.content);
    response.content = JSON.stringify(parsed);
  }
}
```

### After

As we now use `response.data`:

```node
const parser = require('fast-xml-parser');

const afterResponse = (response) => {
  if ('xml' in response.headers.get('content-type')) {
    response.data = parser(response.content);
  }
}
```

### Before

```node
const parser = require('fast-xml-parser');

const afterResponse = (response) => {
  if ('xml' in response.headers.get('content-type')) {
    response.json = parser(response.content);
  }
}
```

### After:

```node
const parser = require('fast-xml-parser');

const afterResponse = (response) => {
  if ('xml' in response.headers.get('content-type')) {
    response.data = parser(response.content);
  }
}
```

### DIFF
--- a/example-apps/babel/src/resources/recipe.js
+++ b/example-apps/babel/src/resources/recipe.js
@@ -2,19 +2,19 @@ const _sharedBaseUrl = 'https://auth-json-server.zapier-staging.com';
 
 const getRecipe = async (z, bundle) => {
   const response = await z.request({
-    url: `${_sharedBaseUrl}/recipes/${bundle.inputData.id}`
+    url: `${_sharedBaseUrl}/recipes/${bundle.inputData.id}`,
   });
-  return response.json;
+  return response.data;
 };
 
 const listRecipes = async (z, bundle) => {
   const response = await z.request({
     url: _sharedBaseUrl + '/recipes',
     params: {
-      style: bundle.inputData.style
-    }
+      style: bundle.inputData.style,
+    },
   });
-  return response.json;
+  return response.data;
 };
 
 const createRecipe = async (z, bundle) => {
@@ -24,23 +24,23 @@ const createRecipe = async (z, bundle) => {
     body: {
       name: bundle.inputData.name,
       directions: bundle.inputData.directions,
-      authorId: bundle.inputData.authorId
+      authorId: bundle.inputData.authorId,
     },
     headers: {
-      'content-type': 'application/json'
-    }
+      'content-type': 'application/json',
+    },
   });
-  return response.json;
+  return response.data;
 };
 
 const searchRecipe = async (z, bundle) => {
   const response = await z.request({
     url: _sharedBaseUrl + '/recipes',
     params: {
-      nameSearch: bundle.inputData.name
-    }
+      nameSearch: bundle.inputData.name,
+    },
   });
-  const matchingRecipes = response.json;
+  const matchingRecipes = response.data;
 
   // Only return the first matching recipe
   if (matchingRecipes && matchingRecipes.length) {
@@ -56,7 +56,7 @@ const sample = {
   name: 'Best Spagetti Ever',
   authorId: 1,
   directions: '1. Boil Noodles\n2.Serve with sauce',
-  style: 'italian'
+  style: 'italian',
 };
 
 // This file exports a Recipe resource. The definition below contains all of the keys available,
@@ -70,31 +70,31 @@ const Recipe = {
   get: {
     display: {
       label: 'Get Recipe',
-      description: 'Gets a recipe.'
+      description: 'Gets a recipe.',
     },
     operation: {
       inputFields: [{ key: 'id', required: true }],
       perform: getRecipe,
-      sample
-    }
+      sample,
+    },
   },
   // The list method on this resource becomes a Trigger on the app. Zapier will use polling to watch for new records
   list: {
     display: {
       label: 'New Recipe',
-      description: 'Trigger when a new recipe is added.'
+      description: 'Trigger when a new recipe is added.',
     },
     operation: {
       inputFields: [
         {
           key: 'style',
           type: 'string',
-          helpText: 'Explain what style of cuisine this is.'
-        }
+          helpText: 'Explain what style of cuisine this is.',
+        },
       ],
       perform: listRecipes,
-      sample
-    }
+      sample,
+    },
   },
   // If your app supports webhooks, you can define a hook method instead of a list method.
   // Zapier will turn this into a webhook Trigger on the app.
@@ -105,7 +105,7 @@ const Recipe = {
   create: {
     display: {
       label: 'Create Recipe',
-      description: 'Creates a new recipe.'
+      description: 'Creates a new recipe.',
     },
     operation: {
       inputFields: [
@@ -114,36 +114,36 @@ const Recipe = {
           key: 'directions',
           required: true,
           type: 'text',
-          helpText: 'Explain how should one make the recipe, step by step.'
+          helpText: 'Explain how should one make the recipe, step by step.',
         },
         {
           key: 'authorId',
           required: true,
           type: 'integer',
-          label: 'Author ID'
+          label: 'Author ID',
         },
         {
           key: 'style',
           required: false,
           type: 'string',
-          helpText: 'Explain what style of cuisine this is.'
-        }
+          helpText: 'Explain what style of cuisine this is.',
+        },
       ],
       perform: createRecipe,
-      sample
-    }
+      sample,
+    },
   },
 
   search: {
     display: {
       label: 'Find Recipe',
-      description: 'Finds an existing recipe by name.'
+      description: 'Finds an existing recipe by name.',
     },
     operation: {
       inputFields: [{ key: 'name', required: true, type: 'string' }],
       perform: searchRecipe,
-      sample
-    }
+      sample,
+    },
   },
 
   // In cases where Zapier needs to show an example record to the user, but we are unable to get a live example
@@ -161,8 +161,8 @@ const Recipe = {
     { key: 'name', label: 'Name' },
     { key: 'directions', label: 'Directions' },
     { key: 'authorId', label: 'Author ID' },
-    { key: 'style', label: 'Style' }
-  ]
+    { key: 'style', label: 'Style' },
+  ],
 };
 
 export default Recipe;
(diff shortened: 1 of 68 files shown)


FokkeZB:
@szchenghuang a heads-up that I'd like to include this in v10 as well. It's the last piece of the puzzle that makes `afterResponse` work well with shorthand/form-mode request. I'll have it ready for review this week.

xavdid:
@FokkeZB we're getting v10 ready to go out the door - do you have a timeline on this?

FokkeZB:
@xavdid I've got it almost finished. I was prioritizing https://github.com/zapier/zapier/pull/39898 over this one, but I now realize that we won't need to switch visual builder apps immediately, so this should come first as we don't want to have another breaking change soon after.

I'll work on this this afternoon and also prioritize writing docs on both with lots of examples.

eliangcs:
@FokkeZB quick question: looks like `response.json` will continue to work, so is it still a breaking change?

FokkeZB:
@eliangcs it would be breaking for CLI apps that currently change `response.content` in their `afterResponse` middleware, with the expectation for form-mode requests to pick up on that. They'll have to replace `response.content = JSON.stringify(parsedOrTransformed)` with `response.data = parsedOrTransformed`.

packages/core/src/http-middlewares/after/prepare-response.js
```diff
@@ -29,20 +31,23 @@ const prepareContentResponse = (resp, request) => {
...
+      json: undefined,
+      data: undefined,
       content: content,
       request: request,
       skipThrowForStatus: request.skipThrowForStatus,
     };
     const outResp = _.extend(preppedResp, replaceHeaders(resp));
+    try {
+      if (outResp.headers.get('content-type') === FORM_TYPE) {
+        outResp.data = querystring.parse(content);
```
FokkeZB, line 45: This adds form-support for `response.data`, which `executeHttpRequest` now relies on but now can also be used via `z.request`.

packages/core/src/http-middlewares/after/prepare-response.js
```diff
@@ -29,20 +31,23 @@ const prepareContentResponse = (resp, request) => {
...
+      } else {
+        outResp.data = JSON.parse(content);
+        outResp.json = outResp.data; // DEPRECATED
```
FokkeZB, line 48: Keeps `.json` around for now. As v10 is breaking we could also drop it. We also may want to not make it a reference to `.data` to prevent unexpected side-effects, but I don't think it's a big deal.

packages/core/test/create-app.js
```diff
@@ -616,4 +617,66 @@ describe('create-app', () => {
       );
     });
   });
+
+  describe('response.content parsing', () => {
```
FokkeZB, line 621: This adds test to ensure that form-mode/shorthand request still work with both JSON and form-urlencoded. We didn't have any tests for it AFAIK :(

packages/core/test/http-middleware.js
```diff
@@ -529,3 +529,151 @@ describe('http logResponse after middleware', () => {
     });
   });
 });
+
+describe('http prepareResponse', () => {
```
FokkeZB, line 533: This adds tests for `prepareResponse` for which AFAIK we didn't have any.

example-apps/oauth1-trello/authentication.js
```diff
@@ -17,7 +15,7 @@ const getRequestToken = async (z, bundle) => {
       oauth_callback: bundle.inputData.redirect_uri,
     },
   });
-  return querystring.parse(response.content);
```
FokkeZB, line 20: Trello is a great example of an app that can now use `response.data`

packages/legacy-scripting-runner/index.js
```diff
@@ -694,22 +693,14 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
         }
 
         if (options.defaultToResponse && !result) {
-          try {
-            result = zcli.JSON.parse(response.content);
-          } catch {
-            result = {};
-          }
```
FokkeZB, line 701: We could already use `response.json` here, but now `response.data`.

packages/cli/src/tests/utils/build.js
```diff
@@ -15,23 +14,31 @@ describe('build (runs slowly)', () => {
...
+    const appPackageJsonPath = path.join(tmpDir, 'package.json');
+    const appPackageJson = JSON.parse(
+      fs.readFileSync(appPackageJsonPath, { encoding: 'utf8' })
     );
-    fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson));
-
+    appPackageJson.dependencies[PLATFORM_PACKAGE] = path.resolve(
+      coreDir,
+      coreTarball
+    );
```
FokkeZB, line 40: I needed to change this test to not use the currently published version but a tarball of the local version or else changes to the typescript app that didn't match the published definitions would break it.

example-apps/babel/src/resources/recipe.js
```diff
@@ -114,36 +114,36 @@ const Recipe = {
           key: 'directions',
           required: true,
           type: 'text',
-          helpText: 'Explain how should one make the recipe, step by step.'
```
FokkeZB, line 117: I'm not exactly sure how these formatting changes got in?

FokkeZB: This is what `yarn prettier --single-quote --write "**/*.js"` as when run by `lint-staged` gets me. Perhaps these files were previously not run through prettier?

packages/core/src/http-middlewares/after/prepare-response.js
xavdid: Yeah. it's _probably_ fine to have them be a reference to each other, but we also don't lose much from making them a copy. I'd probably replace w/ another call to `JSON.parse`

packages/core/src/http-middlewares/after/prepare-response.js
```diff
@@ -29,20 +31,23 @@ const prepareContentResponse = (resp, request) => {
...
+      }
+    } catch (_) {}
```
xavdid, line 50: this `_` variable shadows the root `lodash` declaration. Should probably rename to `e` or `_e` or something. I would have expected `eslint` to catch this. 🤔

packages/legacy-scripting-runner/index.js
eliangcs: Unfortunately, legacy-scripting-runner has to (or is better to) work with older versions to core. Its `package.json` [states](https://github.com/zapier/zapier-platform/blob/eff0e5c3991e6089d4afe911c1870c06d427da38/packages/legacy-scripting-runner/package.json#L40) that it should work as long as core >= 7.6.0. So can we undo the change here?

packages/legacy-scripting-runner/index.js
```diff
@@ -694,22 +693,14 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
...
       } else {
         if (options.checkResponseStatus) {
           response.throwForStatus();
         }
 
-        try {
-          result = zcli.JSON.parse(response.content);
-        } catch {
-          result = {};
-        }
```
eliangcs, line 712: Ditto.

packages/legacy-scripting-runner/index.js
```diff
@@ -725,7 +716,7 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
       skipThrowForStatus: true,
     });
     response.throwForStatus();
-    return querystring.parse(response.content);
```
eliangcs, line 728: Ditto.

FokkeZB:
Thanks for the reviews @xavdid and @eliangcs - all feedback has been addressed.

PR #214, opened 2020-05-07
fix(core): don't log request body when it's form data

eliangcs:
Fixes `logResponse` middleware so we don't `JSON.stringify` a `FormData` object.

Passing a `FormData` object to request body is valid usage when you want to upload a file:

```js
const FormData = require('form-data')

  const form = new FormData();
  form.append('filename', 'sample.txt');
  form.append('file', getStreamSomewhere());

    url: 'https://...',
    body: form
  });

  // ...
};
```

I notice the `logResponse` middleware would just `JSON.stringify` the `FormData` object and send it to our log server, which results in something like [this](https://cdn.zapier.com/storage/files/2244d52292eeb3cba00d0dbac55f70e9.txt). This does not make sense but fine for most cases, at least it doesn't crash. However, I found for certain stream objects, such as Node.js' built-in `http.request`, it would crash with this error:

```
TypeError: Converting circular structure to JSON
    at JSON.stringify
```

This PR fixes that by not JSON stringifying request body when it's a `FormData` object.

### DIFF
--- a/packages/core/src/http-middlewares/after/log-response.js
+++ b/packages/core/src/http-middlewares/after/log-response.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const stream = require('stream');
+
 // Prepare a request/reponse to be logged to the backend.
 // Generally respects the "Zapier" request and resp object format.
 const prepareRequestLog = (req, resp) => {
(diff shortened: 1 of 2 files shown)


packages/core/src/http-middlewares/after/log-response.js
```diff
@@ -6,15 +6,25 @@ const prepareRequestLog = (req, resp) => {
...
   } else {
-    body = '<probably streaming data>';
+    responseBody = '<probably streaming data>';
+  }
+
+  let requestBody = req.body;
+  if (
+    requestBody &&
+    requestBody.constructor &&
+    requestBody.constructor.name === 'FormData'
```
xavdid, line 24: Does this depend on a specific npm package being used? I imagine there's more than one way to generate form data.

eliangcs: Good point! Took a more generic approach in 2b7edbc.

eliangcs: As a side note, `x instanceof stream` is also how node-fetch tells if the body is a stream: https://github.com/node-fetch/node-fetch/blob/73b7a089742e86132b1b2acc53c6b56169cb57c2/src/body.js#L270

PR #217, opened 2020-05-13
ci: use Travis build stages to speed up tests

eliangcs:
Leverages Travis [Build Stages](https://docs.travis-ci.com/user/build-stages/) to run groups of tests in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,13 @@ before_install:
   - pip install --user awscli
 install:
   - yarn install
+stages:
+  - name: "Test"
+  - name: "Deploy"
+    if: tag IS present AND env(TRAVIS_NODE_VERSION) IS "10.16.3"
 jobs:
   include:
-    - stage: "Tests"
+    - stage: "Test"
       name: "cli: unit tests"
       script: "yarn workspace zapier-platform-cli test"
     - name: "cli: smoke tests"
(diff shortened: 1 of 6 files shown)


packages/cli/package.json
```diff
@@ -28,7 +28,7 @@
     "precommit": "yarn docs && git add docs README.md ../../docs",
     "version": "yarn docs && git add docs/* README.md",
     "postversion": "git push && git push --tags",
-    "lint": "eslint src snippets --fix",
+    "lint": "eslint src snippets",
```
eliangcs, line 31: Remove `--fix` so eslint returns a non-zero value instead of actually fixing it when there's a lint error.

xavdid: Ah, this is interesting! If we're running `yarn lint` during development then we'd probably want `--fix` included. But, I almost never do that (my editor auto-fixes what it can, as does the pre-commit hook). Given that this is used in CI where we _do_ want it to fail, this is probably fine.

.travis.yml
```diff
@@ -6,17 +6,40 @@ before_install:
...
+    - name: "legacy-scripting-runner: integration tests"
+      script: "yarn workspace zapier-platform-legacy-scripting-runner test"
+    - stage: "Deploy"
+      script: skip
+      deploy:
+        provider: script
+        script: ./scripts/publish.sh
+        on:
+          tags: true,
+          node: "10.16.3"
```
xavdid, line 39: This is probably fine as is, but we might consider using `10.x` if it's available (or `12.x`, soon!) since lambda doesn't stick to a single version anymore.

PR #221, opened 2020-05-15
BREAKING CHANGE(cli) raise error if integration depends on cli

xavdid:
something I noticed with a partner app today - sometimes they list CLI as a dep. Never need to, so now we'll error for it. We don't really have a zip size problem anymore, but it doesn't hurt to keep devs from including it by accident.

### DIFF
--- a/packages/cli/src/utils/misc.js
+++ b/packages/cli/src/utils/misc.js
@@ -10,6 +10,7 @@ const semver = require('semver');
 
 const {
   PLATFORM_PACKAGE,
+  PACKAGE_NAME,
   PACKAGE_VERSION,
   LAMBDA_VERSION,
 } = require('../constants');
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/misc.js
```diff
@@ -81,20 +82,28 @@ const isValidAppInstall = () => {
   let packageJson;
   try {
     packageJson = require(path.join(process.cwd(), 'package.json'));
+
+    if (_.get(packageJson, ['dependencies', PACKAGE_NAME])) {
+      return {
+        valid: false,
+        reason: `Your integration should not depend on ${PACKAGE_NAME}. Please remove it from the \`dependencies\` section of your \`package.json\`. If you want to include a local install, it should go in \`devDependencies.\``,
+      };
```
eliangcs, line 90: I think this is a bit strict. Can we make it a warning with `console.warn`?

xavdid: My thought was that there's _never_ a reason to have cli as a prod dep. I agree that we shouldn't spend time chasing every dep that probably shouldn't be included, but this one is never needed as is the most likely one to end up in there, so it makes sense to hard block on it. I feel like warnings mostly get ignored. 😁

eliangcs: I agree they never need to include `zapier-platform-cli` in `dependencies`. But we can also speak the same for `jest`, `mocha`, `prettier`, and other dev tools. If they really want to that, maybe it's because they want to use a function from those dev tools and we shouldn't block them from doing that.

eliangcs: To put more thoughts on this, I'm assuming the developer knows what they're doing while you assume the developer is making a mistake when they do that. Your assumption is probably more likely to happen. It'd be great if there's a way to accommodate both groups of developers. That is:

* for the developers who are making a mistake (a large percentage): don't let them ignore the warning
* for the developers who know what they're doing: don't block them

Maybe make the warning message more visible using colors and space so they don't ignore that easily?

xavdid: > don't let them ignore the warning

I guess we could do that by throwing an error but adding a CLI flag that allows the user to overwrite the error. Feels like overkill for this one thing though. 

Honestly, at this point i'm interested in what the numbers are. I think i'll close this for now as not to block the release. I'll see if I can add analytics to capture app id (if available) and what packages each app depends on. then we'll know how widespread an issue this is. Might help inform other dep-related decisions.

eliangcs: Capturing analytics is a good call, thanks!

PR #222, opened 2020-05-15
docs: clean up docs

eliangcs:
Updates stale info in the docs. Also removes that orange banner from the rendered HTML docs.

Before:

After:


eliangcs:
Only doc changes, nothing major. Merging this without a review.

PR #223, opened 2020-05-15
chore: v10 release prep

eliangcs:
Prep for CLI v10 release:

- Adds v10 changelog
- Adds Node 12 to `.travis.yml`


CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
+## 10.0.0
+
+Another major release! We have some great improvements in this version but also have breaking changes. Please review the following to see if you need to change anything to upgrade `zapier-platform-core` to v10.
+
+First, Zapier integrations that depend on the new Core v10 **will run using Node.js 12**. You can install Node.js 12 using `nvm`.
```
xavdid, line 5: I would list each of these bolded breaking changes in the bullets below. I like these wordier summaries, but it should also be easy to scan a list of bullets for ❗and quickly know what, if anything, they need to do to run their app on v10.

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+
+Second, **`z.request` now always calls `response.throwForStatus`** via a middleware by default. You no longer need to call `response.throwForStatus` after `z.request`, the built-in middleware will do that for you. See [Error Response Handling](https://github.com/zapier/zapier-platform/blob/master/packages/cli/README.md#error-response-handling) for details.
```
xavdid, line 7: awesome link to docs here

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+
+Third, **`response.throwForStatus` now only throws an error if the status code is between 400 and 600**. Before v10, it throws for status >= 300. So if your code rely on that old behavior, you should check `response.status` explicitly instead of using `response.throwForStatus`.
```
xavdid, line 9: I'd note that it's 400-600 (inclusive) to be more explicit about it.

xavdid: also, should change a couple of verb tenses: 

> Before v10, it threw for status >= 300. So if your code relies on that old behavior,

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+
+Fourth, we now **parse JSON and form-encoded response body by default**. The parsed object is available as `response.data` (`response.json` will be still available for JSON body but less preferable). Before v10, we only parse JSON for [manual requests](https://github.com/zapier/zapier-platform/blob/master/packages/cli/README.md#manual-http-requests); parse JSON and form-encoded body for [shorthand requests](https://github.com/zapier/zapier-platform/blob/master/packages/cli/README.md#shorthand-http-requests). This change could be breaking if you have an `afterResponse` that modifies `response.content`, with the expectation for shorthand requests to pick up on that. In which case, you'll have to replace `response.content = JSON.stringify(parsedOrTransformed)` with `response.data = parsedOrTransformed`.
```
xavdid, line 11: > Before v10, we only parse JSON for [manual requests]

Two things about that line: 

1. should be `we only parsed` JSON for manual requests
2. also, it was undocumented, but it's not accurate. We always tried to parse json out and put it on `response.json`, even on "manual" `z.request`. See discussion [here](https://zapier.slack.com/archives/CA6FTB140/p1588787498041100?thread_ts=1588785216.036500&cid=CA6FTB140).

So since the "official story" is that they used to have to parse it themselves and no longer have to, this line is fine (besides the typo)

xavdid: Also, I really like the line:

> This change could be breaking if you have an `afterResponse` that modifies `response.content`, with the expectation for shorthand requests to pick up on that. In which case, you'll have to replace `response.content = JSON.stringify(parsedOrTransformed)` with `response.data = parsedOrTransformed`.

because it clearly says what action they need to take and if it applies to them.

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+* :nail_care: `build` command no longer needs login (#216)
+* :nail_care: `promote` command becomes more receptive about the changelog format (#209)
+* :nail_care: Warn when you put `zapier-platform-cli` in your `dependencies` (#221)
+* :hammer: Mass dependency update and linting (#218, #220)
+* :scroll: Update and clean up docs (#222)
+* :scroll: Add some clarity around what we're sending for analytics (#215)
+
+## core
+
+* :exclamation: `response.throwForStatus` now only throws for 400 <= status <= 600 (#192)
```
xavdid, line 31: I think every breaking change mentioned above should also be down here. So i'd mention that app use node 12

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+
+Fifth, we rewrote the CLI `zapier init` command. The project templates are more up-to-date, with better coding practices. However, **we reduced the number of templates**. Try `zapier init --help` to see what are available.
+
+Finally, `zapier init` no longer uses the `minimal` template by default. If you don't specify a `--template`, the command will prompts you interactively. So if you're using `zapier init` in CI and expect to use `minimal` by default, you should change the command to `zapier init --template minimal`.
+
+See below for a detailed changelog:
+
+### cli
+
+* :exclamation: `init` command becomes more interative and comes with brand new project templates (#210)
```
xavdid, line 21: I'd explicitly mention which templates were removed. Something like

> We've removed the following templates: 

> * `trigger` 
> * `search`
> * `babel`

For trigger/search/action, use the `zapier scaffold` command instead. For babel, look at the `typescript` example and replace the build step with the similar code from https://babeljs.io/setup#installation

packages/cli/src/version-store.js
```diff
@@ -12,4 +12,5 @@ module.exports = [
   { nodeVersion: '8.10.0', npmVersion: '>=5.6.0' }, // 7.x
   { nodeVersion: '8.10.0', npmVersion: '>=5.6.0' }, // 8.x
   { nodeVersion: '10', npmVersion: '>=5.6.0' }, // 9.x; it's no longer a specific version
+  { nodeVersion: '12', npmVersion: '>=5.6.0' }, // 10.x
```
xavdid, line 15: is the `npm` version here actually relevant? was there a bug on older versions of npm that we're avoiding?

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+
+Fifth, we rewrote the CLI `zapier init` command. The project templates are more up-to-date, with better coding practices. However, **we reduced the number of templates**. Try `zapier init --help` to see what are available.
+
+Finally, `zapier init` no longer uses the `minimal` template by default. If you don't specify a `--template`, the command will prompts you interactively. So if you're using `zapier init` in CI and expect to use `minimal` by default, you should change the command to `zapier init --template minimal`.
```
xavdid, line 15: > command will prompt you interactively

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+
+See below for a detailed changelog:
```
xavdid, line 17: Might add that **❗denotes a breaking change** so folks will know to scan for it. ¯\\\_(ツ)_/¯

CHANGELOG.md
xavdid: also, a typo: `interative`

packages/cli/src/version-store.js
eliangcs: Not that I'm aware of, but I think npm 5.6.0 is old and not a strict requirement, so we should keep it there.

PR #210, opened 2020-05-02
BREAKING CHANGE(core): Make throwForStatus an opt-out, also for code/function operations

FokkeZB:
> **NOTE:** Replaces #193

## Description

Make throwForStatus an opt-out, also for code-mode/function operations.

### Why?

1. We up its share in v3 errors from 10 closer to the 60% on v2 so that:
    1. We have more control over the (default) error handling, which recently has been updated to extract messages and codes from error responses like we already did for v1/2.
    2. We lower the 20% "errors" that contain JSON, HTML or Node Runtime errors (pointing at no/poor error handling) on v3 closer to the 1% on v2.
2. We force developers to use `afterResponse` for custom error handling if `throwForStatus` throws when it shouldn't (or doesn't - correctly - when it should).

This requires the visual builder to also support defining a (single) `afterResponse` middleware, which is what we've shipped via https://github.com/zapier/zapier/pull/39765 for staff-only as part of the 2020 (remote) engineering retreat, and can make public via https://github.com/zapier/zapier/pull/39899.

### Shorthand/form-mode Requests

We always run `throwForStatus`, following any `afterResponse`, meaning:

1. You cannot use shorthand/form-mode for API requests that may return a status > 300 **when there is no error**.
2. You cannot use shorthand/form-mode for API requests that may return a status <= 300 **when there is an error**, because `throwForStatus` will be called but not throw, after which we'll try to parse `response.content` as a successful result, which will either throw an unhelpful error if we fail to do so or break the Zap as fields will be different.

In CLI apps and after https://github.com/zapier/zapier/pull/39899 you can handle this via `afterResponse`, but for 1) this is hacky:

```node
SHOULD_NOT_THROW = 444;
SHOULD_THROW = 222;

if (response.status === SHOULD_NOT_THROW) {
  response.status = 200;
}

if (response.status === SHOULD_THROW) {
  throw z.errors.Error(response.json.errorMessage);

  // or hacky:
  response.status = 500;
}

return response;
```

### Function/Code-Mode Request

`throwForStatus` is an (easy to forget) opt-in:

```node
return z.request({ ... })
  .then(response => {
    response.throwForStatus();
    return response.json;
  }); 
```

### Function/Code-Mode Requests

No more need to (remember to) call `throwForStatus`:

```node
return z.request({ ... })
  .then(response => response.json);
```

Ability to opt-out of `throwForStatus` for specific requests:

```node
return z.request({
  ...
  skipThrowForStatus: true
})
  .then(response => {
[...]

PR #226, opened 2020-05-19
refactor(legacy-scripting-runner): some tests aren't compiling function source

eliangcs:
Fixes https://github.com/zapier/zapier-platform/pull/210/files#r418914759, where some tests of legacy-scripting-runner aren't properly compiling function source in the app definition. This only fixes the tests and won't affect the library itself.

### DIFF
--- a/packages/legacy-scripting-runner/test/example-app/api-key-auth.js
+++ b/packages/legacy-scripting-runner/test/example-app/api-key-auth.js
@@ -12,14 +12,6 @@ const testAuthSource = `
   });
 `;
 
-const beforeRequestSource = `
-  return z.legacyScripting.beforeRequest(request, z, bundle);
-`;
-
-const afterResponseSource = `
-  return z.legacyScripting.afterResponse(response, z, bundle);
-`;
-
 module.exports = {
   legacy: {
     authentication: {
@@ -40,13 +32,4 @@ module.exports = {
       },
     ],
   },
-  beforeRequest: [
-    {
-      source: beforeRequestSource,
-      args: ['request', 'z', 'bundle'],
-    },
-  ],
-  afterResponse: [
-    { source: afterResponseSource, args: ['response', 'z', 'bundle'] },
-  ],
 };
--- a/packages/legacy-scripting-runner/test/example-app/index.js
+++ b/packages/legacy-scripting-runner/test/example-app/index.js
@@ -1155,9 +1155,9 @@ const MovieSearch = {
   },
 };
 
-// const beforeRequestSource = `
-//   return z.legacyScripting.beforeRequest(request, z, bundle);
-// `;
+const beforeRequestSource = `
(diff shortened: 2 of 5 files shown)


eliangcs:
cc @FokkeZB

packages/legacy-scripting-runner/test/example-app/index.js
```diff
@@ -1194,12 +1194,12 @@ const App = {
     },
   },
   // This breaks `applyBeforeMiddleware` which it looks like gets the original definition without `findSourceRequireFunctions` applied?!
```
FokkeZB, line 1196: We can remove this.

packages/legacy-scripting-runner/test/example-app/api-key-auth.js
```diff
@@ -40,13 +32,4 @@ module.exports = {
...
   },
-  beforeRequest: [
-    {
-      source: beforeRequestSource,
-      args: ['request', 'z', 'bundle'],
-    },
-  ],
-  afterResponse: [
-    { source: afterResponseSource, args: ['response', 'z', 'bundle'] },
-  ],
```
FokkeZB, line 51: I don't understand why we are and can remove these middlewares for the `custom`, `oauth2` and `session` examples as we do have middleware for them in https://github.com/zapier/zapier-platform/blob/4bae98de1deb3aae3e9c2fa0aede140fad02fde7/packages/legacy-scripting-runner/middleware-factory.js.

packages/legacy-scripting-runner/test/example-app/index.js
eliangcs: 🤦I'll do that in another PR. Thanks!

packages/legacy-scripting-runner/test/example-app/api-key-auth.js
eliangcs: Eventually these auth definitions will be merged to the main app definition. For example, see https://github.com/zapier/zapier-platform/blob/4bae98de1deb3aae3e9c2fa0aede140fad02fde7/packages/legacy-scripting-runner/test/integration-test.js#L14-L16 and https://github.com/zapier/zapier-platform/blob/4bae98de1deb3aae3e9c2fa0aede140fad02fde7/packages/legacy-scripting-runner/test/integration-test.js#L59-L62 

so we only need to write middlewares once in the main app definition.

FokkeZB: Ah, ok. I wonder if developers that look at these examples may get confused and think these middleware are not needed, while this is only due to how we merge in the test app.

eliangcs: > developers that look at these examples may get confused

Hopefully, that doesn't happen. Those "examples" are only for us to run tests, they're never meant to be examples for external developers.

PR #228, opened 2020-05-19
refactor(cli,core,legacy-scripting-runner): use Zapier-hosted httpbin

eliangcs:
Replaces all `httpbin.org` with `httpbin.zapier-tooling.com`. Most of the changes are in tests, so there are no user or developer-facing changes. The only change that affects developers is the `digest-auth` project template.

The biggest difference between httpbin.org and httpbin.zapier-tooling.com is that the Zapier one uses arrays to hold values for headers and query params. So in the test code, we need to change a lot of `.should.eql(value)` to `.should.deepEqual([value])`.

While I was there, I refactored some of promise usages into async/await syntax.

### DIFF
--- a/packages/cli/src/generators/templates/authTests/digest.test.js
+++ b/packages/cli/src/generators/templates/authTests/digest.test.js
@@ -18,7 +18,7 @@ describe('digest auth', () => {
     const response = await appTester(App.authentication.test, bundle);
 
     expect(response.status).toBe(200);
-    expect(response.data.authenticated).toBe(true);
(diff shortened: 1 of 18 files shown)


eliangcs:
> do you think it's worth having the httpbin url in a constant somewhere?

Yes! I made `HTTPBIN_URL` and `AUTH_JSON_SERVER` constants for core and legacy-scripting-runner. I didn't change for cli because those are in project templates and auth codegen, which will require some work to contantize. Let's leave it for another day. Thanks!

PR #229, opened 2020-05-19
chore(cli): regenerate example apps using the new init command

eliangcs:
We haven't decided where to showcase all the new project templates. Let's make it simple for now. I regenerated some of the example apps using the new `zapier init` command and replaced the ones in the `example-apps` directory.

All contents in this PR are auto-generated except for `packages/cli/src/tests/utils/build.js`, which was to fix two failing tests depending on how the `typescript` example is set up.


eliangcs:
> we can run this as a script as part of the release process, so we'll always have fresh examples.

@xavdid I'd let Travis run such a script and push code back to GitHub. We can definitely look into it.

PR #230, opened 2020-05-20
fix(cli): fix npm registry dependency in tests

eliangcs:
Fixes yet another circular dependency in the CLI test code. I was releasing v10 and:

- core depended on schema v10
- schema v10 wasn't on npm registry
- broke cli tests (`build` command tests and smoke tests)

So we need to patch `zapier-platform-core`'s `package.json` to make sure it uses local `zapier-platform-schema`.


eliangcs:
Merging this without review to unblock v10 release.

PR #231, opened 2020-05-22
docs: add v10 notes

eliangcs:
Add v10 notes everywhere in the docs, especially with `response.data` and `response.throwForStatus()`, so developers don't get confused if they're still on v9.


eliangcs:
Only doc changes, so I'm merging without review.

PR #232, opened 2020-05-22
ci: enable cache in .travis.yml

eliangcs:
Hopefully to make Travis run faster.


eliangcs:
Looks like it works:



Merging without review.

PR #241, opened 2020-06-11
fix(cli) properly handle invalid keys during VB conversion

### LINKED ISSUE #234: zapier convert throws an error even though the UI doesn't show any
### Bug Description
I started creating my Zaps in the UI and they just work there (they can authenticate and show 
 no errors when validating) but **now I'm trying out the CLI** and I get the following error when running `zapier convert 103195 . --version=1.0.0`

image

All my zaps have this http header binding syntax where I can't see anything amiss..
image

My UI doesn't think there is anything wrong yes few warnings telling me to add description but no errors.

image

I hope I have given you enough information on this for you to help me out because I really would like to be able to extract what I have done in the UI into to code where I get more power.

### Version Info

- Version info: zapier-platform-cli/10.0.0 win32-x64 node-v12.16.2
- Operating System: Windows 10
- App id: I don't get that far because it fails. But if its the same as the "integration id" its 103195

xavdid:
fixes #234, where an auth key like `X-API-KEY` wasn't handled correctly in the resulting JS.

We could probably refactor this conversion code to use the declarative code we did during the eng retreat, but that's probably fine for now.

### DIFF
--- a/packages/cli/src/utils/convert.js
+++ b/packages/cli/src/utils/convert.js
@@ -213,7 +213,7 @@ const renderAuthData = (appDefinition) => {
   const fieldKeys = getAuthFieldKeys(appDefinition);
   const lines = _.map(fieldKeys, (key) => {
     const upperKey = _.snakeCase(key).toUpperCase();
-    return `"${key}": process.env["${upperKey}"]`;
+    return `"${key}": process.env.${upperKey}`;
   });
   if (_.isEmpty(lines)) {
     return `{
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/convert.js
```diff
@@ -213,7 +213,7 @@ const renderAuthData = (appDefinition) => {
   const fieldKeys = getAuthFieldKeys(appDefinition);
   const lines = _.map(fieldKeys, (key) => {
     const upperKey = _.snakeCase(key).toUpperCase();
-    return `${key}: process.env.${upperKey}`;
+    return `"${key}": process.env["${upperKey}"]`;
```
eliangcs, line 216: `upperKey` is snake-cased. Is the double quoting still necessary? But I guess it won't hurt either. 🤷‍♂️

xavdid: oh, right you are. I'll pull it out. keeps it a little cleaner.

PR #249, opened 2020-07-20
docs: regenerate example apps

eliangcs:
* Regenerate `session-auth` and `oauth2` templates to example-apps
* Add missing the `oauth1-trello` to example-apps
* Fix syntax errors in the `oauth1-trello` template


eliangcs:
Going to merge this and cut a release without a review because there isn't anything interesting to see here.

PR #251, opened 2020-07-22
feat(schema,core): Firehose webhooks

casshill:
Update schema to support firehose webhooks
Add extra checks to core for firehose webhooks
Looks like the linter got a bit happy on some extra files


packages/schema/lib/schemas/AppSchema.js
```diff
@@ -102,6 +103,14 @@ module.exports = makeSchema(
           hide: true,
         },
       },
+      firehoseWebhooks: {
+        description:
+          '**INTERNAL USE ONLY**. Zapier uses this for internal webhook app configurations.',
+        $ref: FirehoseWebhookSchema.id,
+        docAnnotation: {
+          hide: true,
```
casshill, line 111: this is how I make it so this is not discoverable in our docs? We want to keep it internal for now (maybe forever)

packages/core/src/checks/firehose_is_array.js
```diff
@@ -0,0 +1,23 @@
...
+
+const _ = require('lodash');
+const { simpleTruncate } = require('../tools/data');
+
+const isFirehoseWebhook = require('./is-firehose-webhook');
+
+/*
+  The firehoseWebhook performSubscriptionKeyList function should always return an array of objects.
+*/
+const firehoseWebhookIsArray = {
```
casshill, line 11: I tried testing this locally, but didn't get this to error. I also tried the trigger is array and it also didn't error (when I deliberately tried to make it). Not sure if there's a particular way to test this locally?

packages/schema/lib/schemas/AppSchema.js
xavdid: yep! you can't hide it from the code itself, but this should keep it out of the docs. You can run `npm run docs` to verify.

packages/core/src/checks/firehose_is_array.js
xavdid: You registered the check correctly, so i'm not sure. It should get run whenever your `perform` is called, so nothing special should be needed. Though, your local app needs to be using this version of core, so that could be it.

We've also got unit tests for these checks, so that should be able to verify the behavior.

packages/schema/docs/build/schema.md
```diff
@@ -700,6 +701,26 @@ An array or collection of fields.
 
 -----
 
+## /FirehoseWebhookSchema
```
xavdid, line 704: I think the schema itself should also be excluded from the doc? if you're trying to hide this

packages/core/src/checks/is-firehose-webhook.js
```diff
@@ -0,0 +1,6 @@
+module.exports = (method) => {
+  return (
+    method.startsWith('firehoseWebhooks.') &&
```
xavdid, line 3: Should `firehoseWebhooks.whatever.performSubscriptionKeyList` be a valid name? It doesn't seem like it, so this could just check if the method is exactly `firehoseWebhooks.performSubscriptionKeyList`

casshill: haha oops. I was just flowing with the style of the other checks 😅

packages/schema/lib/schemas/FirehoseWebhookSchema.js
```diff
@@ -0,0 +1,25 @@
...
+const makeSchema = require('../utils/makeSchema');
+
+const FunctionSchema = require('./FunctionSchema');
+
+module.exports = makeSchema(
+  {
+    id: '/FirehoseWebhookSchema',
+    description:
+      'To idenitfy firehose webhooks are their corresponding Zaps. ' +
+      'For Zapier internal webhooks - not publically available.',
```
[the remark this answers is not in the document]
casshill: At the moment the answer to that is no. But it is possible this might change, although not any time soon. If we did open it up to devs, we would need to make some changes and add a few more safety checks and docs etc

packages/schema/docs/build/schema.md
casshill: hmmm I'm not able to get this working. I tried the docAnnotation hide in both the app schema and the Firehose schema, and it seems to be working at the app level but not the individual schema level. npm run docs still adds FirehoseWebhookSchema to this doc. I don't see any examples of hiding an individual schema, is that feature available?

packages/schema/lib/utils/buildDocs.js
```diff
@@ -27,7 +27,9 @@ const walkSchemas = (InitSchema, callback) => {
 const collectSchemas = (InitSchema) => {
   const schemas = {};
   walkSchemas(InitSchema, (Schema) => {
-    schemas[Schema.id] = Schema;
+    if (!_.get(Schema, 'schema.docAnnotation.hide')) {
+      schemas[Schema.id] = Schema;
+    }
```
casshill, line 32: @xavdid I feel like hiding entire schemas from the docs isn't supported, how do you feel about adding something like this?

xavdid: Nice! yeah, this would work.  

Your other (simpler) option is to not type the `firehoseWebhook` property at all, just make it a generic object like we do for `legacy`: 

https://github.com/zapier/zapier-platform/blob/6d4603b71fb3333a77d063729b8d7fe35583d20e/packages/schema/lib/schemas/AppSchema.js#L102-L109

That might simplify the schema portion of this PR in a significant way.

casshill: ah interesting. So that means essentially you can put whatever you like in there, it's essentially the same as allowing additional properties in a schema? I think either way looks good, if you'd prefer all schemas are documented then I can switch it over to object to be consistent with the hidden legacy schema

casshill: makes sense, we'd want to switch it over to a real documented schema if we ever decide to release to developers, but they may not ever happen - we may keep it private, or "on request"

xavdid: yeah, we'd document if if we released it. it's simple enough that hiding and simplifying it is ok for now. 

🚢  !

PR #253, opened 2020-07-31
fix(legacy-scripting-runner): make response headers case-insensitive

eliangcs:
Fixes [PDE-1561](https://zapierorg.atlassian.net/browse/PDE-1561).

A `KEY_post` method may access the response header like this:

```js
function KEY_post_poll(bundle) {
  var name = bundle.response.headers['My-Name'];
  // ...
}
```

However, in CLI/VB world, our HTTP client is `node-fetch` and it lowercases response headers for us, so the above header accessing will break. It needs to be changed to `bundle.response.headers['my-name']`.

This PR adds an `afterResponse` middleware to make header keys case-insensitive, so both `my-name` and `My-Name`, or even `mY-NaMe` are going to work.

Stole from @codebycaleb's commit https://github.com/zapier/zapier-platform/commit/f0575398887a980f57207b0cff8b4ace3284de6c.

### DIFF
--- a/packages/legacy-scripting-runner/middleware-factory.js
+++ b/packages/legacy-scripting-runner/middleware-factory.js
@@ -222,6 +222,27 @@ const createBeforeRequest = (app) => {
   };
 };
 
+const proxyHeaders = (headers) => {
+  const proxy = {
+    get: (target, prop) => {
+      const original = Reflect.get(target, prop);
+      if (typeof original === 'function' || typeof original === 'symbol') {
+        // defaults to defined functions on the Headers class; the symbol type
+        // is used for accessing an internal map
+        return original;
+      }
+      try {
+        // try to retrieve the header via the get() function
+        return target.get(prop); // try to retrieve the header via the get() function
+      } catch {
+        // otherwise, default to original target[prop] value
+        return original;
+      }
+    },
+  };
+  return new Proxy(headers, proxy);
+};
+
 const createAfterResponse = (app) => {
   const authType = _.get(app, 'authentication.type');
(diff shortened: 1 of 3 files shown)


packages/legacy-scripting-runner/middleware-factory.js
```diff
@@ -222,6 +222,27 @@ const createBeforeRequest = (app) => {
...
+    get: (target, prop) => {
+      const original = Reflect.get(target, prop);
+      if (typeof original === 'function' || typeof original === 'symbol') {
+        // defaults to defined functions on the Headers class; the symbol type
+        // is used for accessing an internal map
+        return original;
+      }
+      try {
+        // try to retrieve the header via the get() function
+        return target.get(prop); // try to retrieve the header via the get() function
```
codebycaleb, line 236: very very inconsequential nitpick here, but it looks like the comment above got copied twice

eliangcs:
@xavdid thanks for the further clarification! But there's one more thing I'd like to clarify:

> In CLI, it's a `Headers` class provided by node-fetch. Since it's not a regular object, no square bracket property access works

This is true _if_ we were to use the response object returned by `node-fetch` directly, we won't able to use square bracket accessors like `response.headers['key']`. But in `prepareResponse` middleware we [transform](https://github.com/zapier/zapier-platform/blob/ccea44af190dfd8c52c00e3681e6fff968e281eb/packages/core/src/http-middlewares/after/middleware-utils.js#L3-L7) the `node-fetch`'s `Headers` object to a regular object, so square bracket accessors actually work for the response object returned by `z.request`.

PR #254, opened 2020-08-04
feat(core): Add check to make sure input fields do not have duplicate keys.

stevemolitor:
Add check to make sure input fields do not have duplicate keys.

When a dev app defines input fields with duplicate keys, weird things can happen both in the editor and when running the zap. It's happened a few times with child (line item) fields, where the parent key of a line items group conflicts with the key of a top level field. It tends to happen when fields are defined dynamically. 

You can see an example of where this happened recently with the MailChimpAPI in [this issue](https://github.com/zapier/zapier/issues/42961). _Looong_ slack convo [here](https://zapier.slack.com/archives/C406QGE15/p1596286348390200). It took a while to figure out what was happening. Basically, a line items group with the key of `email` was added dynamically [here](https://github.com/zapier/zapier-platform-app-mailchimp/blob/f40cb8c8be7be2908b73306631c8a72ef735f58f/src/resources/member.js#L148-L159), which conflicted with a statically defined `email` field, and much weirdness ensued. The fix is [here](https://github.com/zapier/zapier-platform-app-mailchimp/commit/97aed73a48e23c61791c2cd9f6ab6634f7ca8371#diff-bb368e48445c9a3931f82ca68b552eee).

This PR adds a check that will throw an error if we detect duplicate input fields. Maybe just logging something would be better? Even if we just logged a breadcrumb to GL, I think bugs like this would be much quicker to find.

To test this PR, we should probably try it out with an app that we know does not have duplicate input field keys, but that has line item fields. And then test it with a dummy app that intentionally does have duplicate field keys, ideally the scenario where a line item field is added dynamically whose key stomps on another field key.

This approach may be wrongheaded. In that case, consider this PR a conversation starter! For example, maybe this would be better handled or logged in the monolith runner, or .....? 

cc @cvanderkolk @thomir

### DIFF
--- a/packages/core/integration-test/integration-test.js
+++ b/packages/core/integration-test/integration-test.js
@@ -578,6 +578,10 @@ const doTest = (runner) => {
         'resources.failerfuncpromise.list.operation.perform',
         'Failer on promise function!'
       );
+      testError(
+        'resources.duplicateinputfields.list.operation.inputFields',
+        'Invalid API Response:\n  - Duplicate field keys: key 1'
+      );
     });
   });
 };
--- a/packages/core/src/checks/index.js
+++ b/packages/core/src/checks/index.js
@@ -7,4 +7,6 @@ module.exports = {
   triggerIsObject: require('./trigger-is-object'),
   triggerHasUniqueIds: require('./trigger-has-unique-ids'),
   triggerHasId: require('./trigger-has-id'),
+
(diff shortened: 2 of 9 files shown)


stevemolitor:
Closing - [gonna solve on the server side](https://zapier.slack.com/archives/C5Z9BP4U9/p1596584265007600).

PR #255, opened 2020-08-06
docs(cli): remove legacy reference to bundle.meta.zap

casshill:
We should not rely on the bundle.meta.zap object in a CLI App, so we should remove the reference in the documentation. Slack discussion https://zapier.slack.com/archives/CH0SN5V0V/p1596691855293400

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -773,34 +773,7 @@ This object holds the user's auth details and the data for the API requests.
 
 > Before v8.0.0, the information in `bundle.meta` was different. See [the old docs](https://github.com/zapier/zapier-platform-cli/blob/a058e6d538a75d215d2e0c52b9f49a97218640c4/README.md#bundlemeta) for the previous values and [the wiki](https://github.com/zapier/zapier-platform/wiki/bundle.meta-changes) for a mapping of old values to new.
 
-There's also `bundle.meta.zap.id`, which is only available in the `performSubscribe` and `performUnsubscribe` methods.
 
-The user's Zap ID is available during the [subscribe and unsubscribe](https://github.com/zapier/zapier-platform/blob/master/packages/schema/docs/build/schema.md#basichookoperationschema) methods.
-
-For example - you could do:
-
-```js
-const subscribeHook = async (z, bundle) => {
-
-  const options = {
-    url: 'https://57b20fb546b57d1100a3c405.mockapi.io/api/hooks',
-    method: 'POST',
-    body: {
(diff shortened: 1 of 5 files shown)


packages/cli/README.md
```diff
@@ -1679,34 +1679,7 @@ This object holds the user's auth details and the data for the API requests.
...
-  const response = await z.request(options);
-  return response.data; // or response.json if you're using core v9 or older
-};
-
-module.exports = {
-  // ... see our rest hook example for additional details: https://github.com/zapier/zapier-platform/blob/master/example-apps/rest-hooks/triggers/recipe.js
-  performSubscribe: subscribeHook,
-  // ...
-};
-```
```
eliangcs, line 1709: I think we should still keep this code example but only remove the `bundle.meta.zap.id` part. Apart from that, the change looks good!

casshill: I thought that at first, but the example is specifically for the bundle.meta.zap.id
If we have another example using some other part of the bundle.meta that would be good - otherwise I think the example isn't helpful? Maybe we can dig up something else from an existing CLI app - I'll check if someone in the team has something good

eliangcs: It doesn't look like we have an example for `performSubscribe` anywhere else in the docs, which is why I suggest keeping it. It can be still helpful without `bundle.meta.zap.id`. At least readers would know the function signature and what this function should look like at the bare minimum.

packages/cli/README-source.md
```diff
@@ -773,35 +773,30 @@ This object holds the user's auth details and the data for the API requests.
...
+    url: `${API_BASE_URL}/teams`,
+    params,
+  });
 
-module.exports = {
-  // ... see our rest hook example for additional details: https://github.com/zapier/zapier-platform/blob/master/example-apps/rest-hooks/triggers/recipe.js
-  performSubscribe: subscribeHook,
-  // ...
+  return response.json;
 };
```
casshill, line 794: switched the zap.id example (that isn't relevant) to an example that uses bundle.meta

packages/cli/README-source.md
```diff
@@ -849,6 +844,30 @@ module.exports = {
 
 This the URL to which you should send hook data. It'll look something like `https://hooks.zapier.com/1234/abcd`. We provide it so you can make a POST request to your server. Your server should store this URL and use is as a destination when there's new data to report.
 
+For example:
```
casshill, line 847: added the subscription function example here, since this is the relevant section

PR #260, opened 2020-08-20
feat(core): Create Typescript defs from json schema

### LINKED ISSUE #8: Typescript bindings

xavdid:
edit: So the first 90% of this was straightforward and the rest was _not_. Namely, the functions mentioned below. I had to get very into the typescript generics functionality and couldn't get it there in a satisfactory way. 

I think there's a lot of promise here, but it'll be hard to get right for the developers. I think we should circle back at some point. 

---

Fixes #8, our oldest open FR!

I'm still doing some polishing, but the brunt of the schemas look great! I'll also update the TS example a bit.


xavdid:
The one big bug I've found is this:

```ts
export type FunctionSchema = string | FunctionRequireSchema | FunctionSourceSchema;
```

which is correct for the output JSON, but wrong for the function that generates that JSON. Namely, `string` should be something like `(z: ZObject, bundle: Bundle)`, plus the other variations for middleware and stuff. I'll have to think on that.

xavdid:
I also want to re-title everything from `FunctionSchema` to `Function` (or maybe `ZFunction`).

xavdid:
closing for now since I don't have the time to work on this, but it was a good endeavor! got closer than before.

BrettBedarf:
FYI I don't believe this is possible using json schema since  javascript functions are not trivial (and likely not useful represented as objects anyways) in JSON. Correct me if I'm wrong, but reviewing the code base json schema is solely used for documentation? Either way theres a problem is trying to use a json definition to describe javascript objects...might be worth rethinking 🤔

For example, you'd get a lot more mileage especially considering maintenance & dx using typescript throughout as your source of truth, get the onslaught of benefits there, and generate type safe up-to-date docs with tsdoc. You can even go the extra mile and use something like [zod](https://github.com/colinhacks/zod) to validate the schemas of said ts types with a [a helper](https://www.npmjs.com/package/tozod) (there are some caveats but I've had good success overall keeping ts as the source of truth). Then validate as part of create/push and your users don't have to worry about trying to ship misshaped apps whether they use typescript or vanilla js. 

Food for thought!

PR #261, opened 2020-08-20
fix(core): Change typescript HTTPResponse types to use `any` instead of `object`

casshill:
linter is on the fritz again :/

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -104,8 +104,8 @@ interface BaseHttpResponse {
 
 export interface HttpResponse extends BaseHttpResponse {
   content: string;
-  data?: object;
-  json?: object;
+  data?: any;
+  json?: any;
 }
 
 export interface RawHttpResponse extends BaseHttpResponse {
--- a/packages/schema/docs/build/schema.md
+++ b/packages/schema/docs/build/schema.md
@@ -337,13 +337,11 @@ Represents user information for a trigger, search, or create.
 
 * `{ label: 'New Thing', description: 'Gets a new thing for you.' }`
 * ```
-  {
-    label: 'New Thing',
+  { label: 'New Thing',
     description: 'Gets a new thing for you.',
     directions: 'This is how you use the thing.',
     hidden: false,
-    important: true
-  }
+    important: true }
   ```
 
 #### Anti-Examples
@@ -453,48 +451,38 @@ How will Zapier create a new object?
 #### Examples
 
 * ```
-  {
-    key: 'recipe',
+  { key: 'recipe',
     noun: 'Recipe',
     display: { label: 'Create Recipe', description: 'Creates a new recipe.' },
-    operation: { perform: '$func$2$f$', sample: { id: 1 } }
-  }
+    operation: { perform: '$func$2$f$', sample: { id: 1 } } }
   ```
 * ```
-  {
-    key: 'recipe',
+  { key: 'recipe',
     noun: 'Recipe',
     display: { label: 'Create Recipe', description: 'Creates a new recipe.' },
-    operation: { perform: '$func$2$f$', sample: { id: 1 }, shouldLock: true }
-  }
+    operation: { perform: '$func$2$f$', sample: { id: 1 }, shouldLock: true } }
   ```
 * ```
-  {
-    key: 'recipe',
+  { key: 'recipe',
     noun: 'Recipe',
     display: { label: 'Create Recipe', description: 'Creates a new recipe.', hidden: true },
-    operation: { perform: '$func$2$f$' }
-  }
+    operation: { perform: '$func$2$f$' } }
   ```
 
 #### Anti-Examples
 
 * `'abc'`
 * ```
-  {
-    key: 'recipe',
+  { key: 'recipe',
     noun: 'Recipe',
     display: { label: 'Create Recipe', description: 'Creates a new recipe.' },
-    operation: { perform: '$func$2$f$', shouldLock: 'yes' }
-  }
+    operation: { perform: '$func$2$f$', shouldLock: 'yes' } }
   ```
 * ```
-  {
-    key: 'recipe',
+  { key: 'recipe',
     noun: 'Recipe',
     display: { label: 'Create Recipe', description: 'Creates a new recipe.' },
-    operation: { perform: '$func$2$f$' }
-  }
+    operation: { perform: '$func$2$f$' } }
   ```
 
 #### Properties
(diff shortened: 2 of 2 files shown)


packages/core/index.d.ts
```diff
@@ -104,8 +104,8 @@ interface BaseHttpResponse {
 
 export interface HttpResponse extends BaseHttpResponse {
   content: string;
-  data?: object;
-  json?: object;
+  data?: any;
+  json?: any;
```
casshill, line 108: No idea what data is - the same as json?
I think json needs to be any, as object doesn't let you access the json properties in the response

xavdid: Data is the parsed response (but plays nicely with multiple data types). See here: 

https://github.com/zapier/zapier-platform/blob/56f1e53629a99297e6eeea9c3b06a1dc47cbc5c6/packages/core/src/http-middlewares/after/prepare-response.js#L44-L48

as for `any` vs `object`, we know that `data` will be some sort of object type and we want to classify it as such. with `any`, we lose all protection. 

A quick example: [Typescript Playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgVzMGBeGAKEAuGIBGAVgKbBQA0MAhrlWAJ4CUaAfDAN4BQMPMA8kVJQAdAGti9CFmYB6GTAgALEAgA2AExgB3EACdR3XgJJkxEqVVnzQazTv2UwIGAEt1xKgEJOhniGEAtlQADhhWMMB0AOSwQcF4AG7EuniCZBC+1IEhYTByME4wusTQui5kLuAZmc4APDAAzHnySioaMMm6eplUMPVN+UH0+MSeMAD1ADoA+hiAIwyM0zLjPrwRVbDqVFC96M5UEBwQCMDAJRC4+CAgqh5gjlQBxLilLmAA5gC+maCQN8TCVQgd4YLY7YTHU7nRg-Kr-QHA0HbKjCAAe4Vatg6ui6ulhf1uCJBKPRzWoqlUMBGUEOVGKeDgcBg2AwnE+QA)

cyberwitch: https://www.bennadel.com/blog/3126-object-access-bracket-notation-vs-dot-notation-with-typescript-in-angular-2-rc-4.htm

Declaring these types as `object` means the dot notation that is currently used heavily breaks.  I wonder if it would make sense to provide a generic for the expected response type?

xavdid: > Declaring these types as object means the dot notation that is currently used heavily breaks

I don't think that's true, seems fine [in this playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgJIBE5jsg3gKGSOQG0BrALmQGcwpQBzAXStvpAfwF998AbCGGQATLHCoYxyALx5CxGAHtFVAEQAjOFFUAaeUQAWEPnxXJVAd0VQ+w1dwDcvBIpDVFAgHSmGAClHYnkqKAJT4Lm4eEN6KfgFwnkYmoc6u7l4+-mIkAOTBOUxhEenRmfG5SaYFYUA).

I think it's important to communicate that `response.data` definitely isn't a `number` or a `string`. It's _something_ wrapped in curly braces, but before you can safely use it, you have to narrow it with the shape of your data. This is all intended behavior.

cyberwitch: Try with type `object` though - https://www.typescriptlang.org/play?#code/DYUwLgBAJghmMC4IHsBGArEBjSBeCA3gFASkQBmyySARKjAE40A0JZAFiMMNRDQO7IGwKDSIBfANxEiWZADsAzslAA6HgHMAFLHirKyAJSyFytZp1wYqztyMy5SlSHXJtumAG0A5Ae8BdY0czFwsPH1seAOMgA

I think a generic would achieve that same goal of requiring the user to narrow the shape of their data, but it's a little more streamlined than `const data = response.data as ExpectedResponseType`.

xavdid: Sure, but that playground is sort of backwards - `object` is supposed to be a more vague type (and is uncommonly used if you know anything else about the shape of the data already).

I agree that having a generic for the response instead of of having to re-cast the response would be a great improvement!

casshill: I had the same problem - I am not able to use dot notation to access any elements in the json of a http response if I use the HttpResponse type - where json is type "object". The object type seems really strict, and seems only appropriate for when you know exactly what the object will look like, but a http response json could be anything (pretty much)

casshill:
hmm I think you are using the wrong example though (It's confusing, as the json response is different types in different areas). 
JSON.parse is fine, because that returns 'any'
it's response.json I'd really like to use (we previously used JSON.parse, but have been updating apps to use response.json), ie. HttpResponse has a property called data and a property called json. I should be able to use response.json.some_json_key, but can't. Here's an example (from js) of what we do now https://cdn.zappy.app/a0626ccf19f11985ee735966f8c24ada.png which we can't do with ts (note the response.json.items). If it is possible and I'm just using it wrong let me know, this is all new  to me!

lpellegr:
I was experiencing the same issue. If we are using TypeScript it's mainly for type safety. Enforcing a return type that ensures the developer thinks about the return type looks a good idea.

Looking around, I found an official example for the very same use case as discussed here:
https://www.typescriptlang.org/play#example/unknown-and-never

It seems the recommended approach is to use _unknown_  as return type.

The question we might ask is why does the official `JSON.parse` signature makes use _any_: most probably for backward compatibility reasons.

lpellegr:
One additional note about why _unknown_ the correct return type to use and not _object_:

`object` is a type that represents the non-primitive type, i.e. anything that is not `number`, `string`, `boolean`, `symbol`, `null`, or `undefined`:
https://www.typescriptlang.org/docs/handbook/basic-types.html#object

but a JSON value is an object, array, number, or string, or one of the three literal names: false, null or true.

xavdid:
@casshill you mentioned that 

> I should be able to use response.json.some_json_key, but can't

which, from a type safety perspective is correct. You first have to tell TS that `response.json` _has_ `some_json_key` before it'll let you access it. It's not a great experience, but it is enforcing the "safest" behavior - you can't access any properties you haven't explicitly said are on the response.

All that said, @lpellegr asks a great question:

> why does the official JSON.parse signature makes use of any

turns out that's intended: https://github.com/Microsoft/TypeScript/issues/15225

I forgot when writing this that `JSON.parse` doesn't actually have to return an object - lots of other things are valid too. 

Long term it would be great to be able to use generics to let devs easily type responses, but in the meantime I think swapping to `any` is the right call. 

Thanks for bringing this up!

PR #262, opened 2020-08-21
feat(cli) Add custom build hook to zapier commands

cyberwitch:
This PR adds support for a custom build hook to zapier CLI commands.

By specifying a `zapier-build` script in `package.json`, apps can ensure that script will run during `zapier build`, `zapier push`, etc.  This is useful if someone wishes to write apps in TypeScript, or using Babel, since they can ensure they are never pushing a stale build that is out of sync with their sources.

If there is no script named `zapier-build`, there will be no change in functionality.


casshill:
Nice! I really like the idea of using the users defined npm scripts instead of prescribing behaviour. Random aside but how good would it be if the zapier push command checked for a `zapier-push` script and ran that?! If not it just runs the current zapier push flow. A lot of our apps we want to pass the skip npm install flag and it's a pain to have to run `zapier push`, get an error, then run `npm run zapier-push`

packages/cli/src/utils/build.js
```diff
@@ -290,6 +290,19 @@ const maybeNotifyAboutOutdated = () => {
   }
 };
 
+const maybeRunBuildScript = async (options = {}) => {
+  const npmScripts = _.get(
+    require(path.resolve(options.cwd || '.', './package.json')),
```
eliangcs, line 295: I'd suggest changing the `path.resolve` call to:

```js
path.resolve(options.cwd || process.cwd(), 'package.json')
```

packages/cli/src/utils/build.js
```diff
@@ -290,6 +290,19 @@ const maybeNotifyAboutOutdated = () => {
...
+    'scripts'
+  );
+
+  if ('zapier-build' in npmScripts) {
```
eliangcs, line 299: We need to make sure `npmScripts` is not undefined here, otherwise we'll get a `TypeError` if `scripts` isn't defined in `package.json`.

Suggested change:

```js
if (npmScripts && 'zapier_build' in npmScripts) {
  ...
}
```

PR #265, opened 2020-09-01
fix(cli): _zapier-build should be optional

eliangcs:
CLI 10.1.0 incorrectly requires `_zapier-build` script to defined. To reproduce:

```
$ zapier -v
zapier-platform-cli/10.1.0 darwin-x64 node-v12.16.3

$ zapier init myapp --template minimal
...

$ cd myapp
$ yarn
...

$ zapier build
✖ Running _zapier-build script
 ›   Error: npm ERR! missing script: _zapier-build
 ›
 ›   npm ERR! A complete log of this run can be found in:
 ›   npm ERR!     /Users/eliang/.npm/_logs/2020-09-01T03_28_19_771Z-debug.log
 ›
 ›
```

This is because of a misuse of `_.get()`. The diff should be self-explanatory.

Found this bug because the smoke tests were [failing](https://travis-ci.com/github/zapier/zapier-platform/jobs/379667535#L268). I'm not sure why the tests didn't catch that in https://github.com/zapier/zapier-platform/pull/262...

### DIFF
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -300,7 +300,7 @@ const maybeRunBuildScript = async (options = {}) => {
       'package.json'
     ));
 
-    if (_.get(pJson, 'scripts', ZAPIER_BUILD_KEY)) {
+    if (_.get(pJson, ['scripts', ZAPIER_BUILD_KEY])) {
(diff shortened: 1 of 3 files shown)


eliangcs:
@xavdid looking more closely, I think the tests didn't catch that because we were installing CLI from npm instead of from the local repo. I pushed 7808f1b to fix that.

PR #266, opened 2020-09-01
fix(core): don't censor safe URLs in logs

eliangcs:
Fixes https://github.com/zapier/zapier/issues/30341.

We're overzealously censoring everything in `bundle.authData` and `process.env`. A common usage is putting a `BASE_API_URL` in `process.env`. When a developer does that, they're going to see this in the logs:

To make HTTP logs easier to access, we should disclose URL-like values. But when the URL may contain credentials (basic auth username/password or query parameters), we still want to censor. The newly-added test case reflects that.


xavdid (approved):
looks great!

PR #272, opened 2020-09-14
chore(cli): move docs script

xavdid:
This started as a PR to reduce our usage of `lodash` for size reasons, but I don't think it's worth the time right now. 

In any case, we don't actually have to ship this file as part of the CLI, so I moved it!


eliangcs (approved):
Makes sense!

PR #273, opened 2020-09-16
docs(cli): document the `$HOIST$` property

codebycaleb:
We have recommended using the `$HOIST$` keyword when appropriate, but it hasn't been present in any documentation yet. Here's an attempt at succinctly describing the usage of the keyword. 🤞

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -324,6 +324,7 @@ a code {
 </ul>
 </li>
 <li><a href="#dehydration">Dehydration</a><ul>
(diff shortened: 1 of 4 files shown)


eliangcs (approved):
I've been wanting to document this but it's never been my priority. Thanks!

One suggestion though: [Dehydration](https://github.com/zapier/zapier-platform/blob/63d5bf2a7cdc4dd7bf9def37d8f252e31c1f1827/packages/cli/README-source.md#dehydration) section is probably a better place to document this.

PR #275, opened 2020-09-18
feat(cli): A flag to skip validation during the build process

tayhalla:
This PR adds a flag (`skip-validation`) to the build command to allow the user to skip validation during the building and pushing of an app. 

During the course of `push`, the app's schema is validated three separate times:
- Once locally against the bundled version of the schema found [here](https://github.com/zapier/zapier-platform/blob/taylorhalliday/add-skip-validation-flag/packages/core/package.json/#L49-L50)
- Once during the call to `/api/platform/cli/check` on the remote server
- Once during the PUT to `/api/platform/cli/apps/<app_id>/versions/<version>` on the remote server

This flag changes the behavior to:
- Bypass the local validation check
- Bypass the call to the remote `/check`
- Sends an additional flag in the body to the server during the call to `/api/platform/cli/apps/<app_id>/versions/<version>` that instructs the remote to bypass validation. 

It should be noted that this flag will only work for superusers on the remote.


packages/cli/src/oclif/commands/build.js
```diff
@@ -35,6 +36,11 @@ BuildCommand.flags = buildFlags({
         'Skips installing a fresh copy of npm dependencies on build. Helpful for using `yarn` or local copies of dependencies.',
       hidden: true,
     }),
+    'skip-validation': flags.boolean({
```
tayhalla, line 39: After writing this change, realized this is the same CLI our dev community uses... (I think). Might be good to suppress the description here... Would love some feedback

packages/core/package-lock.json
```diff
@@ -0,0 +1,272 @@
+{
```
xavdid, line 1: we use yarn, so there shouldn't be a `package-lock.json` file

packages/cli/src/oclif/commands/validate.js
```diff
@@ -9,10 +9,10 @@ const { validateApp } = require('../../utils/api');
 
 class ValidateCommand extends BaseCommand {
   async perform() {
-    this.log('Validating project locally.');
+    this.log('Validating project locally here.');
 
     const errors = await localAppCommand({ command: 'validate' });
-
+    this.log('Validating project locally: .', errors);
```
xavdid, line 15: this looks like a test string? since errors won't always be present and there's a trailing `.`? if so, you can remove.

packages/cli/src/oclif/commands/build.js
xavdid: Same CLI and all the code/discussion (including this one!) is open source! 😁

packages/cli/src/oclif/commands/validate.js
```diff
@@ -9,10 +9,10 @@ const { validateApp } = require('../../utils/api');
 
 class ValidateCommand extends BaseCommand {
   async perform() {
-    this.log('Validating project locally.');
+    this.log('Validating project locally here.');
```
xavdid, line 12: is this change important?

packages/cli/src/oclif/commands/build.js
xavdid: Since you've `hidden: true`, this flag shouldn't show up in docs or help (`zapier help build` shouldn't include it).

packages/cli/src/oclif/commands/build.js
```diff
@@ -35,6 +36,11 @@ BuildCommand.flags = buildFlags({
...
+      description:
+        "Skips a local and remote validation check of the CLI app's schema and AppVersion integrity.",
```
xavdid, line 41: In case anyone _is_ reading through the code, you might mention that this is skipping pre-push checks

xavdid (approved):
Nice! Looks like there's some extra logging you were using during dev. Plus, you should remove the `package-lock.json`. Otherwise, it's looking good!

packages/cli/src/oclif/commands/build.js
tayhalla: Oh ya, public repo... hi world 👋 😄 

> Since you've hidden: true

Oh ya, I guess I did think about this when I made the change. The comment just occurred to me as I was taking a fast last pass through my changes. All good then.

packages/core/package-lock.json
tayhalla: keep forgetting `yarn` is the tool of choice. Will nuke it

packages/cli/src/oclif/commands/validate.js
tayhalla: This was actually intentional sans the `.` and the `errors`. I'll clean it up.

packages/cli/src/oclif/commands/validate.js
tayhalla: nope! was badely placed a logging/debug add. Will revert.

packages/cli/src/oclif/commands/validate.js
tayhalla: nvm nuked this

PR #278, opened 2020-10-07
docs(core, schema): [CTO-11] Add docs for under documented callback URL feature.

bryanhelmig:
Adding some more documentation to cover the `z.generateCallbackUrl()` usage.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -733,31 +733,35 @@ For example, in your `perform` you might do:
 ```js
 const perform = async (z, bundle) => {
   // something like this url:
-  // https://zapier.com/hooks/callback/123/deadbeef-dead-dead-dead-deaddeafbeef/abcdef0123456789abcdef0123456789abcdef01/
+  // https://zapier.com/hooks/callback/123/abcdef01-2345-6789-abcd-ef0123456789/abcdef0123456789abcdef0123456789abcdef01/
   const callbackUrl = z.generateCallbackUrl();
   await z.request({
     url: 'https://example.com/api/slow-job',
     method: 'POST',
     body: {
-      // ... whatever your app needs
+      // ... whatever your integration needs
       url: callbackUrl,
     },
   });
-  return {"hello": "world"};
+  return {"hello": "world"}; // available later in bundle.outputData
 };
 ```
 
-And in your own `/api/slow-job` view (or more likely, an async job) you'd make this request to Zapier:
+And in your own `/api/slow-job` view (or more likely, an async job) you'd make this request to Zapier when the long-running job completes to populate `bundle.cleanedRequest`:
 
 ```http
-POST /hooks/callback/123/deadbeef-dead-dead-dead-deaddeafbeef/abcdef0123456789abcdef0123456789abcdef01/ HTTP/1.1
+POST /hooks/callback/123/abcdef01-2345-6789-abcd-ef0123456789/abcdef0123456789abcdef0123456789abcdef01/ HTTP/1.1
 Host: zapier.com
 Content-Type: application/json
 
 {"foo":"bar"}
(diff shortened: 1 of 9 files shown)


bryanhelmig:
I think I'll need to include something on how to use `performResume` to handle the `cleanedRequest`/`rawRequest`/`outputData` call to complete a callback.

bryanhelmig:
This just improves the docs, mind giving it a review @xavdid?

packages/cli/README-source.md
```diff
@@ -724,6 +724,49 @@ The `z.cursor` object exposes two methods:
...
+### `z.generateCallbackUrl()`
+
+The `z.generateCallbackUrl()` will return a callback URL your app can `POST` to later for handling long running tasks (like transcription or encoding jobs). In the meantime, the Zap and Task will wait for your response and the user will see the Task marked as waiting.
+
+For example, in your `perform` you might do:
+
+```js
+const perform = async (z, bundle) => {
+  // something like this url:
+  // https://zapier.com/hooks/callback/123/deadbeef-dead-dead-dead-deaddeafbeef/abcdef0123456789abcdef0123456789abcdef01/
```
xavdid, line 736: It's probably random, but maybe random up something that doesn't include `dead-dead-dead` 😁

packages/cli/README-source.md
```diff
@@ -724,6 +724,49 @@ The `z.cursor` object exposes two methods:
...
+  const callbackUrl = z.generateCallbackUrl();
+  await z.request({
+    url: 'https://example.com/api/slow-job',
+    method: 'POST',
+    body: {
+      // ... whatever your app needs
```
xavdid, line 742: Officially, we call apps "integrations", so you should replace that here

packages/cli/README-source.md
```diff
@@ -724,6 +724,49 @@ The `z.cursor` object exposes two methods:
...
+      url: callbackUrl,
+    },
+  });
+  return {"hello": "world"};
```
xavdid, line 746: I'd add a comment to clarify that this object is what populates `bundle.outputData` later (right?). I know it's clarified below, but this is the snippet most people will read

packages/cli/README-source.md
```diff
@@ -724,6 +724,49 @@ The `z.cursor` object exposes two methods:
...
+};
+```
+
+And in your own `/api/slow-job` view (or more likely, an async job) you'd make this request to Zapier:
```
xavdid, line 750: I'd clarify with: 

> you'd make this request to Zapier when the long-running job completes:

packages/cli/README-source.md
```diff
@@ -724,6 +724,49 @@ The `z.cursor` object exposes two methods:
...
+
+```http
+POST /hooks/callback/123/deadbeef-dead-dead-dead-deaddeafbeef/abcdef0123456789abcdef0123456789abcdef01/ HTTP/1.1
+Host: zapier.com
+Content-Type: application/json
+
+{"foo":"bar"}
+```
+
+And finally, in a `performResume` to handle the final step (which will receive `bundle.cleanedRequest`, `bundle.rawRequest`, and `bundle.outputData`):
```
xavdid, line 760: I'd clarify which `bundle.X` properties come from where. It looks like

* `bundle.cleanedRequest` is `{foo: "bar"}`, the payload for the callback url
* `bundle.rawRequest` is the request object for `bundle.cleanedRequest`
* `bundle.outputData` is the data returned from the initial `perform`

xavdid (approved):
Nice! Left some minor doc clarity things, but it looks good.

packages/cli/README-source.md
bryanhelmig: 😄 Fair enough!

PR #279, opened 2020-10-07
feat(core): Add bulkReads as a qualified user of the cursor storage

tayhalla:
While trying to leverage the paging cursor in the new `bulkReads` actions, I noticed the system was claiming the tokens being passed up were null. Found this check and modified to accommodate for `bulkReads`

I looked into adding tests, but it doesn't seem this restriction is tested anywhere. Not sure where a good place would be to add one, or if we're cool w this point change.


xavdid (approved):
This will get you rolling for CLI app unit tests, but you'll need to make a change on the server to get it working for real. I've sent the link in slack.

PR #280, opened 2020-10-14
fix(core): Add callback_url fallback for testing

codebycaleb:
`z.generateCallbackUrl()` presently returns `undefined` when called via `AppTester`. This commit adds a placeholder value to be returned by `z.generateCallbackUrl()` - `'auth-json-server.zapier-staging.com/echo'`.


xavdid (approved):
oh nice! great catch.

PR #281, opened 2020-10-15
feat(cli): Add example-app for callbacks

codebycaleb:
The example app uses the [/magic](https://github.com/zapier/auth-json-server/pull/25) endpoint from auth-json-server to demonstrate `z.generateCallbackUrl()` functionality.


eliangcs (commented):
Haven't read the code yet. But maybe we should make this example part of `zapier init` templates?

bryanhelmig:
Super cool @codebycaleb! Is `https://auth-json-server.zapier-staging.com` private to our VPN or can users actually run this in production?

codebycaleb:
@eliangcs I thought that our `zapier init` sourced from this directory. 😅  I definitely want to make it part of `zapier init`! I'll dig into that some more this afternoon - I just found https://github.com/zapier/zapier-platform/tree/master/packages/cli/src/generators through some sleuthing so hopefully I'll be able to figure it out. I'll ping you if I run into a wall!

@bryanhelmig Users can run this in production! 😁

xavdid:
Yeah Chang-Hung (as always) is right! All the init templates are generated dynamically at runtime now. Our long-term plan was to have the `example-apps` directory just be the result of `init`ing each template (so folks could browse without having to actually run any code locally). So if you can, I'd use the codegen/generator tools to make your new example. Then run it, and drop the built example into `example-apps`. The generators are a little confusing at first, but we've got a bunch of functions that return strings that are valid JS. so you'd build a JS file out of a ton of those functions. 

​Let me know if you've got any other questions!

example-apps/callback/test/creates.js
```diff
@@ -0,0 +1,63 @@
+/* globals describe, it */
+
+require('should');
```
xavdid, line 3: we're writing new tests w/ `jest` and it's assertion functions now, see [the typescript test](https://github.com/zapier/zapier-platform/blob/79cec9d851a0123b70e66ca8ede7b9e9c2ea4521/example-apps/typescript/src/test/creates.test.ts)

codebycaleb:
Okay cool! I've got the generator version of this up and running now. 😁

xavdid (approved):
awesome!

PR #282, opened 2020-10-16
fix(cli): improve logging for debugging info

xavdid:
This is a minor detail that's bugged me for a while - when we converted to `oclif`, we lost the ability to have custom info when `zapier --version` was run (which is explicitly asked for when people file issues). This is [what we used to do](https://github.com/zapier/zapier-platform/blob/3cbaa2f10bb45f5c2f88a7afb229c13b394a1483/packages/cli/src/utils/misc.js#L147-L200).

This restores availability for some of that info, plus allows us to add more later.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -94,6 +94,7 @@
     ],
     "hooks": {
       "init": [
+        "./src/oclif/hooks/versionInfo",
         "./src/oclif/hooks/deprecated",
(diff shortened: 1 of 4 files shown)


eliangcs (approved):
Nice improvement! But I found `zapier -v` still prints the same text, unlike `zapier --version`. Feel free to merge once you fix that.

xavdid:
Ah, typo! Good eyes as always.

PR #285, opened 2020-10-22
docs(cli): Add "fallback sample" info

owarij:
Added some information about Fallback sample

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -516,6 +516,16 @@ Each of the 3 types of function expects a certain type of object. As of core v1.
 | Search  | Array       | 0 or more objects. If len > 0, put the best match first                                                              |
 | Action  | Object      | Return values are evaluated by [`isPlainObject`](https://lodash.com/docs#isPlainObject)                              |
 
+### Fallback Sample
+In cases where Zapier needs to show an example record to the user, but we are unable to get a live example from the API, Zapier will fallback to this hard-coded sample. This should reflect the data structure of the Trigger's perform method, and have dummy values that we can show to any user.
+
+```js
+,sample: {
+  dummydata_field1: 'This will be compared against your perform method output'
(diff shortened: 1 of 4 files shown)


packages/cli/README-source.md
```diff
@@ -516,6 +516,16 @@ Each of the 3 types of function expects a certain type of object. As of core v1.
 | Search  | Array       | 0 or more objects. If len > 0, put the best match first                                                              |
 | Action  | Object      | Return values are evaluated by [`isPlainObject`](https://lodash.com/docs#isPlainObject)                              |
 
+### Fallback Sample
+In cases where Zapier needs to show an example record to the user, but we are unable to get a live example from the API, Zapier will fallback to this hard-coded sample. This should reflect the data structure of the Trigger's perform method, and have dummy values that we can show to any user.
+
+```js
+,sample: {
```
xavdid, line 523: the leading comma here is a little confusing. maybe change it to something like

```js
// ...
sample: {
  field1: 'this is a field',
  style: 'mediterranean',
}
// ...
```

packages/cli/README-source.md
```diff
@@ -516,6 +516,16 @@ Each of the 3 types of function expects a certain type of object. As of core v1.
...
+  dummydata_field1: 'This will be compared against your perform method output'
```
xavdid, line 524: Having `dummydata` in the key is misleading - the keys should be the real ones, but the value should be faked. See my suggestion above

xavdid (approved):
Nice! The other thing you need to do **before you merge** is run `yarn` and `yarn docs` so your changes get copied into the actual readme. 

I'm also happy to do that for you if you don't have `yarn` set up and don't want to bother.

xavdid:
I went ahead and rebuilt the docs so your changes are included in our upcoming release. Thanks!

PR #293, opened 2020-12-03
[SUPPORTESC-101] docs(cli): Change 90-day limit for callbacks to 30-day

armchairlinguist:
We believe that due to tasks no longer being saved for 90 days, a callback posting after 30 days has a risk of failing, so the documentation should state 30 days as the max.


eliangcs:
@codebycaleb is this related to your previous work? If yes, I'll let you do the review. Thanks!

codebycaleb (approved):
Looks 💯 to me! It should definitely be a safer bet than 90 days.

PR #294, opened 2020-12-04
PDE-1816 fix(legacy-scripting-runner): bundle.request.data should be an object on the second attempt of pre_oauthv2_refresh

eliangcs:
`pre_oauthv2_refresh` can be called twice if the first attempt failed. We're already ensuring `bundle.request.data` is an object on the first attempt but NOT on the second attempt.

More context in [PDE-1816](https://zapierorg.atlassian.net/browse/PDE-1816).


xavdid (approved):
nice catch!

PR #304, opened 2021-01-13
PDE-1915: fix(legacy-scripting-runner): shouldn't double encode when pre method also encodes the URL

eliangcs:
Fixes an issue where a `pre` method (including `KEY_pre_poll`, `KEY_pre_write`, `pre_subscribe`, etc.) encodes the URL and causes the URL to be double encoded. The test should demonstrate how to reproduce.


xavdid (approved):
These encoding ones are tricky! Katie had a similar bug [in slack](https://zapier.slack.com/archives/CTK4JHZAR/p1605904703187100). The layers of encoding we can do is definitely tricky. 

Also I tagged you in slack, but this PR may have triggered the bot that releases the changes (or you did that manually, i'm not sure). In either case, something to keep an eye on or fix.

PR #305, opened 2021-01-15
PDE-1918: fix(legacy-scripting-runner): better choose between ResponseError and script error

eliangcs:
This handles the case where:

1) we get an error response (status code is 4xx or 5xx), and
2) the `KEY_post_(poll|write|search)` method also throws an error

Which one should we throw, the response error (1) or script error (2)?

WB chooses the error based on the script error type. If the script error is one of the Zapier-defined exception classes, throw the script error; throw the response error otherwise.

More details can be found in [PDE-1918](https://zapierorg.atlassian.net/browse/PDE-1918).

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -12,6 +12,16 @@ const {
 
 const { flattenPaths } = require('zapier-platform-core/src/tools/data');
 
+const {
+  ErrorException,
+  HaltedException,
+  StopRequestException,
+  ExpiredAuthException,
+  RefreshTokenException,
+  InvalidSessionException,
+  DEFINED_ERROR_NAMES,
+} = require('./exceptions');
+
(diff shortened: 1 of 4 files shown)


packages/legacy-scripting-runner/index.js
```diff
@@ -680,6 +691,7 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
 
       if (!options.parseResponse) {
         if (options.checkResponseStatus) {
+          response.skipThrowForStatus = false;
```
eliangcs, line 694: Setting `skipThrowForStatus` to false before every `throwForStatus()` is only necessary for core 10+. Otherwise, `throwForStatus()` won't throw anything on core 10+.

core 9.4.0 doesn't have this `skipThrowForStatus` feature and will ignore.

xavdid (approved):

PR #308, opened 2021-01-27
docs(cli): add warning if user counts are still being calculated

xavdid:
Fixes [PDE-1944](https://zapierorg.atlassian.net/browse/PDE-1944)

### DIFF
--- a/packages/cli/src/oclif/commands/versions.js
+++ b/packages/cli/src/oclif/commands/versions.js
@@ -22,6 +22,12 @@ class VersionCommand extends BaseCommand {
       emptyMessage:
         'No versions to show. Try adding one with the `zapier push` command',
     });
+
+    if (versions[0].user_count === null) {
+      this.warn(
+        'User counts are still being calculated - run this command again in ~10 seconds (or longer if your integration has lots of users).'
+      );
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


packages/cli/src/oclif/commands/versions.js
```diff
@@ -22,6 +22,12 @@ class VersionCommand extends BaseCommand {
       emptyMessage:
         'No versions to show. Try adding one with the `zapier push` command',
     });
+
+    if (versions[0].user_count === null) {
+      this.warn(
+        'User counts are still being calculated - run this command again in ~10 seconds (or longer if your integration has lots of users).'
```
eliangcs, line 28: The user count for each version is calculated independently, so it's possible that some versions have the count while some others don't.

So instead of only checking the first version, we should check all the versions. If there's any version having a null `user_count`. We should show this warning.

Also, I'd probably say "**Some** user counts are still being calculated" to avoid any confusion.

xavdid: Ah, fair point! Got lucky on an app having an exact test case:



Fixed in 9940fa68

eliangcs (approved):
Looks great, thanks!

PR #310, opened 2021-02-01
fix(cli): Required value must be boolean for authentication field (PDE-1953)

iKlsR:
This will prevent building as the process halts on validation errors.

--- 

_edit by David_

[PDE-1953](https://zapierorg.atlassian.net/browse/PDE-1953)


xavdid (approved):
great catch, thank you!

PR #313, opened 2021-02-04
PDE-1966: fix(core): allow redirecting from https to http while disabling SSL checks

eliangcs:
## Steps to reproduce

1. Disable SSL certificate checks in your [advance settings](https://zapier.com/app/settings/advanced).
2. Create a CLI/VB app with a trigger that redirects a request from an `https` URL to `http`. You can use a URL like `https://httpbin.zapier-tooling.com/redirect-to?url=http://example.com` to do that.

## Expected behavior

Successfully redirected to the `http` URL.

## Actual behavior

Error: `Protocol "http:" not supported. Expected "https:"`

### DIFF
--- a/packages/core/src/http-middlewares/before/disable-ssl-cert-check.js
+++ b/packages/core/src/http-middlewares/before/disable-ssl-cert-check.js
@@ -1,12 +1,20 @@
 'use strict';
 
+const http = require('http');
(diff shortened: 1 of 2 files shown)


xavdid (approved):
Great work!

PR #326, opened 2021-02-11
SUPPORTESC-121: docs(cli): Make file stashing snippets copy-paste-ible

codebycaleb:
The old version of these snippets would return an "undefined" result. This change makes copying and pasting more accessible.


xavdid (approved):
Ah, awesome. The context in which our snippets are used is a great thing to keep in mind!

PR #327, opened 2021-02-12
fix(cli): ensure test files can be run out of the box with jest

xavdid:
Test files were still using a `mocha` setup instead of `jest`. Now newly-scaffolded tests will run correctly.

[PDE-1907](https://zapierorg.atlassian.net/browse/PDE-1907)


eliangcs (approved):
Looks good! I didn't test as I assume you've tested.

xavdid:
sure did! fixed the failing smoke test (updated file name) too.

eliangcs (approved):

PR #329, opened 2021-02-16
PDE-2012: fix(legacy-scripting-runner): pre_oauthv2_token and pre_oauthv2_refresh discrepancies

eliangcs:
Fixes OAuth2 scripting discrepancies:

- If `pre_oauthv2_token` or `pre_oauthv2_refresh` is defined, we should NOT retry to fetch or refresh the access token.
- We should not retry either if the error is not thrown by `response.throwForStatus()`. That is, retry only when it's a `ResponseError`.
- The OAuth payload for `pre_oauthv2_token` should be in `bundle.request.params`. Previously, we were setting `bundle.request.data` on the first attempt. And only if it fails, we retried it with `bundle.request.params`.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -846,13 +846,15 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
     return urlObj.href;
   };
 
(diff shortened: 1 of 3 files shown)


xavdid (approved):
good catch! so many little edge cases here.

PR #331, opened 2021-02-18
PDE-2012: fix(legacy-scripting-runner): yet-to-save auth fields should be in bundle.auth_fields

eliangcs:
In `pre_oauthv2_token`, `bundle.auth_fields` should contain yet-to-save auth fields. `bundle.auth_fields` was empty because:

- yet-to-save auth fields are in `bundle.inputData` in the CLI world
- we were copying only from `bundle.authData`, not from `bundle.inputData`, to `bundle.auth_fields`

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -868,9 +868,9 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
     );
     authFieldKeys.push('_zapier_account_id');
 
-    // In CLI, it's bundle.inputData instead of bundle.authData has yet-to-save
-    // (auth hasn't been saved) auth fields. In WB, bundle.auth_fields always
-    // has yet-to-save auth fields.
+    // In CLI, it's bundle.inputData instead of bundle.authData that has
+    // yet-to-save (auth hasn't been saved) auth fields. In WB,
+    // bundle.auth_fields always has yet-to-save auth fields.
     bundle.authData = {
       ..._.pick(bundle.inputData, authFieldKeys),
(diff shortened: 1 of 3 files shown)


packages/legacy-scripting-runner/index.js
```diff
@@ -863,6 +863,19 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
       templateContext
     );
 
+    const authFieldKeys = (_.get(app, 'authentication.fields') || []).map(
+      (f) => f.key
+    );
+    authFieldKeys.push('_zapier_account_id');
+
+    // In CLI, it's bundle.inputData instead of bundle.authData has yet-to-save
```
szchenghuang, line 871:
```suggestion
    // In CLI, bundle.inputData instead of bundle.authData has yet-to-save
```

packages/legacy-scripting-runner/index.js
```diff
@@ -863,6 +863,19 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
...
+    // (auth hasn't been saved) auth fields. In WB, bundle.auth_fields always
```
szchenghuang, line 872:
```suggestion
    // (auth hasn't been saved) auth fields. In WB, bundle.authFields always
```

szchenghuang (dismissed):

packages/legacy-scripting-runner/index.js
eliangcs: I actually wanted to keep the "it's". Fixed the grammar in 30cc5ef.

szchenghuang (approved):

PR #337, opened 2021-03-02
[SUPPORTESC-124] docs(cli): Add note that auth is required for integrations looking to go public

armchairlinguist:
This was noted internally, but didn't seem to be highlighted externally. As auth can be omitted in CLI development, this seemed like a good place to note it.


kgodtfredsen (approved):

xavdid (approved):
strictly speaking there are always edge cases, but this is true 99% of the time

armchairlinguist:
Thanks for the reviews, y'all! After discussions with other teams it isn't clear we want to document this so firmly, so I'm going to close this for now, and maybe we'll revisit in the future.

PR #347, opened 2021-03-15
BREAKING CHANGE (schema) add functional constraint for globally unique input fields

xavdid:
Input fields' `key` must be unique across all fields in a given trigger. Nesting fields under `children` doesn't help here.

### DIFF
--- a/packages/schema/lib/functional-constraints/uniqueInputFieldKeys.js
+++ b/packages/schema/lib/functional-constraints/uniqueInputFieldKeys.js
@@ -1,0 +1,62 @@
+'use strict';
+
+const _ = require('lodash');
+const jsonschema = require('jsonschema');
+
+const actionTypes = ['triggers', 'searches', 'creates'];
+
+const uniqueInputFieldKeys = (definition) => {
+  const errors = [];
+  for (const actionType of actionTypes) {
+    const group = definition[actionType] || {};
+    _.each(group, (action, key) => {
+      const inputFields = _.get(action, ['operation', 'inputFields'], []);
+
+      const existingKeys = {}; // map of key to where it already lives
+
+      inputFields.forEach((inputField, index) => {
+        if (existingKeys[inputField.key]) {
+          errors.push(
+            new jsonschema.ValidationError(
+              `inputField keys must be unique for each action. The key "${
+                inputField.key
+              }" is already in use at ${actionType}.${key}.operation.${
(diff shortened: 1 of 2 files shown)


packages/schema/package.json
```diff
@@ -34,6 +34,7 @@
     "fs-extra": "^9.0.0",
     "istanbul": "0.4.5",
     "markdown-toc": "^1",
+    "mocha": "^7",
```
xavdid, line 37: This was apparently needed to run schema-only tests? Since it looks in `schema/node_modules/.bin` instead of `root/node_modules`

packages/schema/lib/functional-constraints/uniqueInputFieldKeys.js
```diff
@@ -0,0 +1,62 @@
...
+          );
+        } else {
+          existingKeys[inputField.key] = `inputFields[${index}]`;
+        }
+
+        (inputField.children || []).forEach((subField, subFieldIndex) => {
+          if (existingKeys[subField.key]) {
+            errors.push(
+              new jsonschema.ValidationError(
+                `inputField keys must be unique for each action, even if they're children. The key "${
```
xavdid, line 39: This is a little wordy, but it works ¯\\\_(ツ)_/¯

packages/schema/package.json
eliangcs: I don't think this is necessary. When we run `npm test` or `yarn test` under `packages/schema`. `npm` or `yarn` is smart enough to find `mocha` at the root level.

eliangcs (commented):
I tested and it works like a charm! My only suggestion is to remove `mocha` from `package.json`.

packages/schema/package.json
xavdid: Ah, `npm test` was actually my issue:

```
% npm test

> zapier-platform-schema@10.2.0 test /Users/david/projects/zapier/platform/packages/schema
> mocha -t 10000 --recursive test

sh: mocha: command not found
npm ERR! Test failed.  See above for more details.
```

`yarn test` works as expected though. I'll remove!

eliangcs (approved):

packages/schema/package.json
eliangcs: Oh yeah, back to think about it, it only works on `yarn test` works because of yarn workspaces.

Issue #334, opened 2021-02-19
'Files can only be stashed within a create' error thrown for resource based creates


owarij:
### Bug Description

If you use z.stashFile in your CLI integration and it's a resource-based/generated create app - You'll get an error thrown -

 “Files can only be stashed within a create or hydration function/method”

![image](https://user-images.githubusercontent.com/24947557/108499513-c9b0f880-7284-11eb-8254-0711d4e3cb44.png)

It looks like the check here may need some updating/love - https://github.com/zapier/zapier-platform/blob/fb60ef9ba9/packages/core/src/tools/create-file-stasher.js#L94

zapzap:
This issue has been copied into our private issue tracker  (as [PDE-2030](https://zapierorg.atlassian.net/browse/PDE-2030)). Thanks for the report! We'll update this as we learn more.

xavdid:
Hey @owarij I think you're right that check needs updating. I'm having a little trouble reproducing the bug in a test project. do you mind sharing/pointing me towards your resource & test? Thanks!

PR #349, opened 2021-03-17
fix(core): handle stashing files in resource create methods

xavdid:
Fixes https://github.com/zapier/zapier-platform/issues/334 | [PDE-2030](https://zapierorg.atlassian.net/browse/PDE-2030)

Also tweaked a schema example because I couldn't remember if dashes were allowed in keys (they're not!)

### DIFF
--- a/packages/core/src/tools/create-file-stasher.js
+++ b/packages/core/src/tools/create-file-stasher.js
@@ -86,12 +86,14 @@ const createFileStasher = (input) => {
       return ZapierPromise.reject(new Error('rpc is not available'));
     }
 
-    const isRunningOnHydrator =
-      _.get(input, '_zapier.event.method', '').indexOf('hydrators.') === 0;
-    const isRunningOnCreate =
-      _.get(input, '_zapier.event.method', '').indexOf('creates.') === 0;
+    const method = _.get(input, ['_zapier', 'event', 'method'], '');
+    const isValidSource =
+      method.startsWith('hydrators.') ||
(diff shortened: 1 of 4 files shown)


packages/core/src/tools/create-file-stasher.js
```diff
@@ -86,12 +86,14 @@ const createFileStasher = (input) => {
...
-    const isRunningOnCreate =
-      _.get(input, '_zapier.event.method', '').indexOf('creates.') === 0;
-
-    if (!isRunningOnHydrator && !isRunningOnCreate) {
+    const method = _.get(input, ['_zapier', 'event', 'method'], '');
+    const isValidSource =
+      method.startsWith('hydrators.') ||
+      method.startsWith('creates.') ||
+      // key regex from KeySchema
+      method.match(/^resources\.[a-zA-Z]+[a-zA-Z0-9_]*\.create\./);
```
eliangcs, line 94: Seems the `+` is unnecessary?

```suggestion
      method.match(/^resources\.[a-zA-Z][a-zA-Z0-9_]*\.create\./);
```

eliangcs (approved):
The regex pattern could be one-char simplified. Apart from that, all looks good!

packages/core/src/tools/create-file-stasher.js
xavdid: sure! I pulled it from the `KeySchema`. it's been that way for ages. 😅 

Updated!

eliangcs (approved):

PR #350, opened 2021-03-18
BREAKING CHANGE(core) run apps using node 14

xavdid:
I figured I'd do a dep update in a separate PR. 

Also, I _think_ we can take advantage of the travis [build matrix](https://docs.travis-ci.com/user/build-matrix/) feature to more cleanly declare our tests. On each of 12 and 14, run each of `test` and `smoke test`. I know you mentioned travis being slow;  i'm not sure this will help a lot, but it'll at least clean up that file.

### DIFF
--- a/boilerplate/scripts/build.sh
+++ b/boilerplate/scripts/build.sh
@@ -113,8 +113,8 @@ update_deps "$BOILERPLATE_DIR/package.json" $CORE_VERSION $LEGACY_VERSION
 pushd $BOILERPLATE_DIR > /dev/null
 
 # Build the zip!
-# the node-X segment in the next line should match all supported node versions
-zip -R $TARGET_FILE '*.js' '*.json' '*/linux-x64-node-12/*.node' '*/linux-x64-node-14/*.node'
+# the node-X segment in the next line should match all node versions where a boilerplate built with this script might run on
+zip -R $TARGET_FILE '*.js' '*.json' '*/linux-x64-node-10/*.node' '*/linux-x64-node-12/*.node' '*/linux-x64-node-14/*.node'
 
 # Remove generated files
 rm -f zapierwrapper.js definition.json core-*.tgz legacy-*.tgz
--- a/boilerplate/scripts/build.sh
+++ b/boilerplate/scripts/build.sh
@@ -113,8 +113,8 @@ update_deps "$BOILERPLATE_DIR/package.json" $CORE_VERSION $LEGACY_VERSION
 pushd $BOILERPLATE_DIR > /dev/null
 
 # Build the zip!
-# the node-X segment in the next line should match the latest major version
-zip -R $TARGET_FILE '*.js' '*.json' '*/linux-x64-node-10/*.node' '*/linux-x64-node-12/*.node'
+# the node-X segment in the next line should match all supported node versions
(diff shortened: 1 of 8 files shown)


packages/cli/src/utils/build.js
```diff
@@ -159,7 +159,7 @@ const forceIncludeDumbPath = (appConfig, filePath) => {
   return (
     filePath.endsWith('package.json') ||
     filePath.endsWith('definition.json') ||
-    filePath.endsWith(path.join('bin', 'linux-x64-node-10', 'deasync.node')) ||
+    filePath.endsWith(path.join('bin', 'linux-x64-node-12', 'deasync.node')) ||
```
eliangcs, line 162: Not sure if I understand [AWS Lambda runtime support policy](https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html) correctly. It seems Node.js 10 _could_ still be running after it is deprecated on May 28. Does that mean we should keep both Node 10 and 12 here?

eliangcs (commented):
- I left a question about Node 10 support.
- GitHub Actions also provides Build Matrix, of which I think we can take advantage.

boilerplate/scripts/build.sh
```diff
@@ -113,8 +113,8 @@ update_deps "$BOILERPLATE_DIR/package.json" $CORE_VERSION $LEGACY_VERSION
 pushd $BOILERPLATE_DIR > /dev/null
 
 # Build the zip!
-# the node-X segment in the next line should match the latest major version
-zip -R $TARGET_FILE '*.js' '*.json' '*/linux-x64-node-10/*.node' '*/linux-x64-node-12/*.node'
+# the node-X segment in the next line should match all supported node versions
+zip -R $TARGET_FILE '*.js' '*.json' '*/linux-x64-node-12/*.node' '*/linux-x64-node-14/*.node'
```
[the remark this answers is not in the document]
xavdid: yeah, I'll grab that in my big update everything

packages/cli/src/utils/build.js
xavdid: Good question! so my understanding of how this works is that functions will keep running, but at some point we won't be able to update them anymore. This has mostly been fine for us, since a lot of our functions are long-lived.

But, that does mean if we update boilerplates on godzilla apps anytime after may, we'll need to also bump their node versions (and by extension, their platform versions). though, I suppose we could be building a new boilerplate between merging this and upgrading node. It's probably safest to include as many node binaries as we can, I can't imagine they're that big. I can tweak this and the other spot.

xavdid:
I've updated! This binary and version stuff is getting a little confusing, even for me 😅

eliangcs (approved):

packages/cli/README.md
```diff
@@ -2952,7 +2952,7 @@ There are a lot of details left out - check out the full example app for a worki
 
 ### Why doesn't Zapier support newer versions of Node.js?
 
-We run your code on AWS Lambda, which only supports a few [versions](https://docs.aws.amazon.com/lambda/latest/dg/programming-model.html) of Node (the latest of which is `v12`. As that updates, so too will we.
+We run your code on AWS Lambda, which only supports a few [versions](https://docs.aws.amazon.com/lambda/latest/dg/programming-model.html) of Node (the latest of which is `v14`. As that updates, so too will we.
```
rnegron, line 2955: Same comment as above, we are not currently in parity with Lambda Node releases.

PR #351, opened 2021-03-19
chore(cli, core, schema) update deps

xavdid:
This is going to look like a fairly noisy PR, but it should be pretty reviewable commit-by-commit. Most updates were drop-in replacements without any substantive changes. The notable commits ones are as follows:

1. 9389f58: `eslint` update - a few code updates, like `var` -> `let` and `.map` -> `.forEach` when we're not actually building an array
2. db961cb: `yeoman` update - method rename, plus specifying `gen.destinationPath` in a couple of places where we weren't. I'm not sure why it worked before, but now it definitely does!
3. c319b9e: `husky` (our git hook controller)` organizes itself a little differently again. I ran their [auto-update script](https://typicode.github.io/husky/#/?id=husky-4-to-5-cli), plus did some minor tweaks
    * thing I didn't know: `jest some-installed-thing` (like `eslint` or `prettier`) will "just work", even if you're somewhere in a subtree. I knew `yarn test` would run scripts, but didn't know about the install binaries. 
    * that's why the husky hook changed from `lerna run...` to `yarn lerna run ...`, because the command was no longer in the `package.json`
4. a225025: updated a few non-critical `legacy-scripting-runner` deps. Also updated `deasync`, which I _think_ is a forward compatible change. I ran unit tests locally using Node 14 and they worked, so ¯\\\_(ツ)_/¯


eliangcs (approved):
Looks good! I tested the `zapier init` command to make sure it still works after the Yeoman upgrade. Other dep upgrades should be compatible if the tests pass, IMO.

PR #357, opened 2021-04-11
fix(core): inputData for hydration function should be of type T as well (PDE-2198)

kaznovac:
this might be a breaking change, depending on what objects developers have passed to `dehydrate` methods - wrong calls (with extra fields in `inputData`) will be marked so one can consider it a fix rather than a breaking change.

if extra `inputFields` fields should be tolerated a union with generic index type (i.e. `T & { [extra: string]: any }`) will describe the `inputData` more loosely.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -117,7 +117,7 @@ export interface RawHttpResponse extends BaseHttpResponse {
 
 type DehydrateFunc = <T>(
   func: (z: ZObject, bundle: Bundle<T>) => any,
-  inputData: object
+  inputData: T
 ) => string;
 
 export interface ZObject {


xavdid (approved):
Very cool! Feels like a good change. I'll go ahead and mark as breaking - since it _is_ more correct, but could break TS apps. we have a major version coming out soon, so it shouldn't delay the release. 

---

For other readers, see the impact of this change in [the playground](https://www.typescriptlang.org/play?strict=true#code/PTAEGcEsFsAcBsCmoBuiBOUD2A7UWAzCLaZQ-AV3VABcBPWRcAWACg3IcaMCBDAY2QAhCjgAmSADwBJHLAo0AIrxq9QAXlABvUAG0AHgC4INdJwDmAXWO8cdUAF8AfNrahQneUpW9jsr8qqANxsDmxs9IygiogAFnRi6CqIAGKi-BqgkgAqTgAUbqAE6cZ5AF7GAFoA8gBGAFaI-DQANKC1ohKIxiLiUrkAlBoutnQthZ4Kgb6g2aAgoBTgZA1NNB444Ny8YvhEczRYEIhk8Lu1cbwokFjobEPqLltmOOYh7KyRyDWrzZlahTEcToSW4xhi8USyTSOH441Y7nQiAAjhQmDRSlR4MZnhYBo9QAAFdAkSDLSSjJyhcKsfi4LagCqgH6NP6aAEI0BI1HozHobEmF7mfFOYmk5YAOiR4Cw8DQeVMaIG8PcQPioO6oDyBDasF4SWg4BFHPciMQNCoeAABnTZVbCmFWI62AsgShQHSgTS6Zt1pCNQBZLDXJiZcpVOqs1rtTpIHqxxCSLTQYOQRDSMTGHAUaAXdDOY2FBYAARo4AAtIh9IxmpX0CTqOXQDheKRQGIsEwcABydZVsk0Qo+mVICXwLDmPIdPqICWTbyqCUt0gDQpIi3oPDTrpzuRTHwSlMhjPzMCQdYAaxwWAA7uBQEe0yf+1sWE6aWUJWqQck8v7kkGIbgG0yapummagAAzA4q6sCWZaVtWax1g2oBNnScDwGm1D8Co-CxKGkRYGwn7fhqf4JIGYHAdoKZiMeEEAOwwS6YBVqYahfPe+rILw8DjjeiBiCRX7AuR-7cIBaY0VotoCoqiBtI+4HGNBAxAA)

kaznovac:
@xavdid  you made me think about this xD

default ts behaviour is not to allow extra properties... went through the example and here's what's going on:

```typescript
// simple version of some of our types

interface Bundle<InputData = { [x: string]: any }> {
  inputData: InputData;
}

type DehydrateFunc = <T>(
  func: (z: ZObject, bundle: Bundle<T>) => any,
  inputData: T // use object instead of T to see old behavior
) => string;

type ZObject = {
  dehyrate: DehydrateFunc, // <-- this will leave the generic T unknown - to be determined later by inference
  request: (url: string)=> Promise<any>
}

const z: ZObject = {
  request: (url: string)=>Promise.resolve(true),
  dehyrate: (f, params)=>{
    return `cool`
  }
}

// dev code

const hydrateMovies = (z: ZObject, bundle: Bundle<{movieId: number}>)=>{
  // @ts-expect-error - name doesn't exist
  console.log(bundle.inputData.name)
  return bundle.inputData.movieId // it knows movieId exists
}

z.dehyrate(hydrateMovies, {movieId: 3})
// @ts-expect-error - complier catches typo
z.dehyrate(hydrateMovies, {modvieId: 7})
// extra types are allowed
z.dehyrate(hydrateMovies, {cool: true, movieId: 3}) // <-- type T here is inferred to the second parameter's type i.e. `{cool: true, movieId: 3}` [hover the dehydrate method in playground]
```

i'm not aware how to tell the ts to use the first instance for inference without explicit `T` stating...
this would be explicit typesafe call that marks `cool` as an unknown property
```typescript
// @ts-expect-error - complier catches unknown properties, T is explicitly stated
z.dehyrate<{movieId: number}>(hydrateMovies, {cool: true, movieId: 3})
```

good thing is that i think current behavour doesn't need to be changed (extra fields shouldn't break the hydrate code; although some typos might slip when T has similarly named properties), i was just curious why it works :)

xavdid:
That's so interesting! because `hydrateMovies` _does_ type its bundle with `movieId`, typos are caught. 

But, extra properties are also allowed 🤔  I'll look into it and find out why.

In any case, I think this PR is correctly marked as not being breaking.

kaznovac:
got the answer on the stack overflow: https://stackoverflow.com/questions/67108349/is-there-a-way-to-hint-ts-which-occurrence-of-generic-parameter-to-use-for-gener/67110530#67110530

i'm guessing the typescript is doing the 'safe thing' and inferring the type as the union

i agree, this is not a breaking brehaviour

xavdid:
Oh nice! that's good to know. Thanks for looking into it.

PR #358, opened 2021-04-15
PDE-2205: build(core): typescript target es2019 for node 12

kaznovac:
as per https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

node 12 is es2019 compatible


eliangcs (approved):
👍

Issue #169, opened 2020-03-04
`source.on is not a function` errors when using async / await for z.stashFile


codebycaleb:
### Bug Description

The following hydration code will raise a `source.on is not a function` error:

```
const x = await z.request(...);
return z.stashFile(x);
```

This stems from the `form-data` NPM package expecting a Stream (or Stream-like) object to be passed in for files, but we're passing in a Response. We already have some handling built in for when `z.stashFile` is passed a Promise (https://github.com/zapier/zapier-platform/blob/f7c42348057ae09323d27247800ccc5e5779de8a/packages/core/src/tools/create-file-stasher.js#L107-L162). We could extract this handling and apply it for non-Promise calls as well.

### Reproduction Steps

1. `zapier init example-files --template files`
2. `cd example-files`
3. `npm i`
4. `zapier test` (everything should work)
5. Edit the `hydrators.js` file to the following (adding `async` and `await`):

```
const hydrators = {
  downloadFile: async (z, bundle) => {
    // use standard auth to request the file
    const filePromise = await z.request({
      url: `https://wt-d9eeb64793d8836c8641adb2acda6ed3-0.run.webtask.io/download-file?id=${bundle.inputData.fileId}`,
      raw: true
    });

    // and swap it for a stashed URL
    return z.stashFile(filePromise).then(url => {
      z.console.log(`Stashed URL = ${url}`);
      return url;
    });
  }
};

module.exports = hydrators;
```

6. `zapier test` raises the error in the title

### Version Info

- Version info: `zapier-platform-cli/9.1.0 darwin-x64 node-v10.16.3`
- Operating System: `macOS 10.15.3`
- App id: N/A

PR #361, opened 2021-04-20
SUPPORTESC-148 fix(core): stashFile no longer throws 'source.on' error when a request that uses await is passed in

nicolemichellethomas:
This Pull Request fixes [this issue](https://github.com/zapier/zapier-platform/issues/169)

Previously, when await was used on a request that was passed to z.stashFile like so:

```
const x = await z.request(...);
return z.stashFile(x);
```

formData would throw the `source.on is not a function`. This PR fixes that issue. 

Jira: https://zapierorg.atlassian.net/browse/SUPPORTESC-148

### DIFF
--- a/packages/core/src/tools/create-file-stasher.js
+++ b/packages/core/src/tools/create-file-stasher.js
@@ -106,7 +106,7 @@ const createFileStasher = (input) => {
     return rpc('get_presigned_upload_post_data', fileContentType).then(
       (result) => {
         const parseFinalResponse = (stream) => {
-          var newBufferStringStream = stream;
+          let newBufferStringStream = stream;
 
           // if stream is string, don't update headers, just return as is
           if (_.isString(stream)) {
--- a/packages/core/src/tools/create-file-stasher.js
+++ b/packages/core/src/tools/create-file-stasher.js
@@ -105,70 +105,69 @@ const createFileStasher = (input) => {
 
     return rpc('get_presigned_upload_post_data', fileContentType).then(
       (result) => {
-        if (isPromise(bufferStringStream)) {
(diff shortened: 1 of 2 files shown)


xavdid:
@nicolemichellethomas thanks for this PR! I'll dig into this in earnest tomorrow (unless someone else beats me to it). In the meantime, can you link your jira card and provide a little context in the PR description? End-users don't have jira access, so being able to read the PR helps them understand changes for the code they use. 

Sounds like this one fixing an issue where an error is thrown, but shouldn't be?

nicolemichellethomas:
@xavdid You bet, sorry about that! Wasn't sure about the protocol 😅 Let me know if you have any questions or need anything else!

packages/core/src/tools/create-file-stasher.js
```diff
@@ -105,70 +105,69 @@ const createFileStasher = (input) => {
...
-                      contentDisposition.parse(cd).parameters.filename;
-                  }
-                }
-              } else {
-                throw new Error(
-                  'Cannot stash a Promise wrapped file of unknown type.'
-                );
-              }
+        const parseFinalResponse = (stream) => {
+          var newBufferStringStream = stream;
```
eliangcs, line 109:
```suggestion
          let newBufferStringStream = stream;
```

eliangcs (approved):
I confirmed it fixes the issue. I only have a suggested change on the `var` vs. `let` lining error.

The file stasher code is one of the spots that always take me a long time to understand. So really appreciate you took the time to read the code and fix it!

nicolemichellethomas:
@eliangcs Awesome! I've committed the let -> var change you suggested :)

eliangcs (approved):
💯

PR #365, opened 2021-05-06
PDE-2333: Fix duplicate API Key headers part 2

eliangcs:
Turns out #364 only fixes part of the issue. This PR fixes another edge case that can lead to duplicate API Key headers.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -307,7 +307,7 @@ const dedupeHeaders = (headers) => {
     if (lowerToFirstKeys[lowerKey] === undefined) {
       lowerToFirstKeys[lowerKey] = k;
     }
-    const firstKey = lowerToFirstKeys[lowerKey] || k;
+    const firstKey = lowerToFirstKeys[lowerKey];
     result[firstKey] = v;
     return result;
   }, {});
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -295,22 +295,22 @@ const cleanHeaders = (headers) => {
 };
 
 // Header keys are considered case insensitive. This function removes duplicate
-// headers. A former value is replaced by latterly appearing value with the same
-// case-insensitive key.
+// headers.
+// dedupeHeaders({'x-key': 'one', 'X-KEY': 'two'}) returns {'x-key': 'two'}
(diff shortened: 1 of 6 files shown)


packages/legacy-scripting-runner/test/example-app/index.js
```diff
@@ -317,6 +317,15 @@ const legacyScriptingSource = `
         return bundle.request;
       },
 
+      movie_pre_poll_double_headers: function(bundle) {
```
eliangcs, line 320: Before calling `movie_pre_poll_double_headers`, `bundle.request.headers` has a `X-Api-Key: one` header.

packages/legacy-scripting-runner/test/example-app/index.js
```diff
@@ -317,6 +317,15 @@ const legacyScriptingSource = `
...
+        bundle.request.url = '${HTTPBIN_URL}/get';
+        bundle.request.headers = {
+          'x-api-key': 'two',
+          'X-API-KEY': 'three'
+        };
+        return bundle.request;
```
eliangcs, line 326: By the time `movie_pre_poll_double_headers` returns, `bundle.request.headers` only has two entries: `x-api-key: two` and `X-API-KEY: three`.

We expect the final headers should only contain one entry `x-api-key: three`.

packages/legacy-scripting-runner/index.js
```diff
@@ -295,22 +295,22 @@ const cleanHeaders = (headers) => {
...
   const lowerToFirstKeys = {};
 
-  const newHeaders = {};
-  Object.entries(headers).forEach(([k, v], index) => {
+  const newHeaders = Object.entries(headers).reduce((result, [k, v]) => {
     const lowerKey = k.toLowerCase();
     if (lowerToFirstKeys[lowerKey] === undefined) {
       lowerToFirstKeys[lowerKey] = k;
     }
     const firstKey = lowerToFirstKeys[lowerKey] || k;
```
szchenghuang, line 310: @eliangcs It seems `const firstKey = lowerToFirstKeys[lowerKey]` is enought?

szchenghuang (approved):

packages/legacy-scripting-runner/index.js
eliangcs: @szchenghuang you're right. Removed that in ad2b37a. Thanks!

PR #374, opened 2021-05-18
fix(cli): handle missing versions better in env command

xavdid:
Fixes [PDE-2356](https://zapierorg.atlassian.net/browse/PDE-2356). 

1. First, explicitly check that the version exists before we try to make an API call (the server error isn't super helpful here)
2. Also, `this.error` expects a string, so if we get an list of errors back, we should only send the first one. That's the `TypeError: first argument must be a string or instance of Error` in the jira ticket.

### DIFF
--- a/packages/cli/src/oclif/commands/env/set.js
+++ b/packages/cli/src/oclif/commands/env/set.js
@@ -37,6 +37,11 @@ class SetEnvCommand extends BaseCommand {
     }, {});
 
     const app = await this.getWritableApp();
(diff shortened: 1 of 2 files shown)


eliangcs (approved):
Looks good! Would be better if it has a test but I'm fine without one!

xavdid:
> Would be better if it has a test

You and me both 😅 I don't think we have any test coverage on the commands themselves right now. We have decent coverage on the underlying API code though! Maybe this is my next hackathon project.

Issue #375, opened 2021-05-18
Functions schema validation failure in v11.0.0


EtienneLem:
### Bug Description

Trying to upgrade to v11.0.0, we’re now getting schema validation errors.

Here’s an example of the error messages we’re getting:

```
App.creates.draft.operation.inputFields[17].key
inputField keys must be unique for each action. The key "undefined" is already in use at creates.draft.operation.inputFields[16].key

App.creates.draft.operation.inputFields[18].key
inputField keys must be unique for each action. The key "undefined" is already in use at creates.draft.operation.inputFields[16].key
```

Looking at the schema docs and the validation implementation of #347, it feels like it might not be handling the cases of functions? Validation takes for granted that `key` is required, but if my understanding of the schema if correct, `FunctionSchema` does not require a `key`.

### Reproduction Steps

These key-less fields are functions. Both async and sync fails the validation. Our `inputFields` are typically structured like this:

```js
const getField = (z, bundle) => {
  return {
    key: 'c',
    children: [{ key: 'c1' }, { key: 'c2' }],
  }
}

const getDynamicField = async (z, bundle) => {
  if (!bundle.inputData.someData) {
    return []
  }

  const fields = await fetchFields()
  return [
    {
      key: 'd',
      fields
    }
  ]
}

const inputFields = [
  { key: 'a' },
  { key: 'b' },
  getField,
  getDynamicField,
]
```

Here it would probably fail when validating `getDynamicField` because "undefined" is already in use at `inputFields[2]`.

### Version Info

- Version info: 11.0.0
- Operating System: macOS (arm64)
- App id: 129003 (not publicly available yet)

zapzap:
This issue has been copied into our private issue tracker  (as [PDE-2410](https://zapierorg.atlassian.net/browse/PDE-2410)). Thanks for the report! We'll update this as we learn more.

xavdid:
Great find! I'll get that shored up. In the meantime, you can downgrade `zapier-platform-core` to the `10.x` line or skip validation until it's fixed.

xavdid:
The fix is merged and it'll go out in the next release sometime in the next few weeks. Thanks for the report!

EtienneLem:
Great! Thanks for the super quick turnaround 🤘🏼

xavdid:
@EtienneLem this is released with `core@11.0.1`, available now.

EtienneLem:
Awesome, thanks!

PR #376, opened 2021-05-19
fix(schema): skip checking keys on fields without key properties

xavdid:
Fixes https://github.com/zapier/zapier-platform/issues/375 (and [PDE-2410](https://zapierorg.atlassian.net/browse/PDE-2410))


eliangcs (approved):
Thanks for fixing!

eliangcs:
@xavdid we'd want to cut a release for this because the issue is blocking developers from using CLI 11.x. I'll do that next week if you don't have time this week.

xavdid:
Yeah good memory! I want to see if there's any other little bug fixes we can get in there. I'll look through issues tomorrow.

Issue #16, opened 2019-06-25
Feature Request: Allow calling the `z.request` method within tests


andreynering:
In our Zapier App, we' touch a real API instance, instead of relying on mocking, because that makes we more confident about our tests.

Since doing that is slow, we're using the Sepia library to cache HTTP requests on tests (a HTTP request is only made by the first time for a given endpoint): https://github.com/linkedin/sepia

Well, it'd be really useful to do HTTP request directly on tests using the `request` method (currently only available inside the Zapier App as `z.request`), for example, to fetch a given record and pass as parameter to a given function. I'd expect that to execute middlewares, like the authentication one.

xavdid:
@karser @andreynering @Lardpower hi folks! 

I took a swing at implementing this today in #385. It still needs some polish and tests, but I believe the core idea should work. If one of y'all has some time in the next week, I'd appreciate you verifying that it covers your use cases. There's testing instructions in that PR and I can field any other questions you may have.

Lardpower:
@xavdid, thank you!

standielpls:
Cleaning up some old issues, this has been released as of zapier-platform-core v11.1.0 ([changelog](https://github.com/zapier/zapier-platform/blob/main/changelog/v10-v13.md#1110))

PR #385, opened 2021-06-04
feat(core): allow running adhoc functions in tests

xavdid:
Addresses https://github.com/zapier/zapier-platform/issues/16.

see also: [PDE-2539](https://zapierorg.atlassian.net/browse/PDE-2539)

## Todo

- [x] verify that TS works nicely
- [x] write unit tests

## Testing

1. With this PR cloned locally, run `yarn link` in the `core` directory
2. elsewhere, run `zapier init ad-hoc-test -t custom-auth`
3. run `cd ad-hoc-test && yarn && yarn link zapier-platform-core`

Now, `appTester` can be fed ad-hoc functions. Try the following test:

```js
/* globals describe, it, expect */

const zapier = require("zapier-platform-core");

const App = require("../index");
const appTester = zapier.createAppTester(App);

describe("custom auth", () => {
  it("passes authentication and returns json", async () => {
    const bundle = {
      authData: {
        apiKey: "secret",
      },
    };

    const response = await appTester(
      (z, bundle) =>
        z.request({ url: "https://auth-json-server.zapier-staging.com/me" }),
      bundle    
    );
    expect(response.data).toHaveProperty("username");
  });
});
```

It's a simple test, but it proves a point: a test function can use `z.request` and take advantage of middleware.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -2666,7 +2666,7 @@ describe('triggers', () => {
 
 ### Using the `z` Object in Tests
 
-Introduced in `core@11.1.0`, `appTester` can now run arbitrary functions:
+Introduced in `core@11.1.0`, `appTester` can now run arbitrary functions that are not in your app definition:
 
 ```js
 /* globals describe, expect, test */
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -117,6 +117,7 @@ This doc describes the latest CLI version (**11.0.1**), as of this writing. If y
     + [v10 Breaking Change: Auth Refresh](#v10-breaking-change-auth-refresh)
 - [Testing](#testing)
   * [Writing Unit Tests](#writing-unit-tests)
+  * [Using the `z` Object in Tests](#using-the-z-object-in-tests)
   * [Mocking Requests](#mocking-requests)
   * [Running Unit Tests](#running-unit-tests)
   * [Testing & Environment Variables](#testing--environment-variables)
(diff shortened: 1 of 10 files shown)


xavdid:
@eliangcs this isn't ready for an actual review yet, but I figured it would be worth getting some early eyes just in case there's a reason we shouldn't pursue this.

eliangcs:
@xavdid this is cool! But I wonder what prevents developers from doing this injection inside their code. In #16, @Lardpower mentioned they could inject the function in the test code:

```javascript
let App = require('../../index');

// include any custom functions in any custom App field like "lib"
App.lib = {
    user: require('./lib/user.js')
}

...

// use it later during the test
appTester(App.lib.user.get).then(response => {});
```

I'm probably missing something. Does the above code have any problems?

xavdid:
That's a valid workaround, but it's sort of an antipattern to tell users to add invalid properties to their app. If it's behavior we want to support, then being able to call functions inline (pre-defined or otherwise) feels better. Feels more "above the board".

Does that make sense? Should I go ahead with finishing this PR?

eliangcs:
@xavdid the feature request makes sense to me, but I don't think the `adHoc` flag is necessary. Could we make `appTester` automatically inject the function if it doesn't exist in the app definition? Something like:

```javascript
    let method;
    try {
      method = resolveMethodPath(appRaw, methodOrFunc);
    } catch (err) {
      // We may need to let resolveMethodPath throw different error types to do this
      if (e instanceof MethodNotfound) {
        appRaw._testRequest = (z, bundle) => methodOrFunc(z, bundle);
        method = resolveMethodPath(appRaw, methodOrFunc);
      } else {
        throw err;
      }
    }
```

What do you think?

xavdid:
Sure, makes sense. I did it without a property in 67b48f66. We'll run any function we get, but it correctly throws errors if you give an object (or other allowed type) that isn't found.

eliangcs:
@xavdid looking good! Let me know once this PR is ready for review.

xavdid:
Ok, this is ready! add tests, docs, and cleaned up `resolveMethodPath` while I Was in there.

packages/cli/README.md
```diff
@@ -2663,6 +2664,60 @@ describe('triggers', () => {
 
 ```
 
+### Using the `z` Object in Tests
+
+Introduced in `core@11.1.0`, `appTester` can now run arbitrary functions:
```
eliangcs, line 2669:
```suggestion
Introduced in `core@11.1.0`, `appTester` can now run arbitrary functions that are not in your app definition:
```

eliangcs (approved):
Thanks for the good test coverage! I only have a suggestion on the docs.

eliangcs (approved):

PR #388, opened 2021-06-21
[PDE-2554] fix(cli): Remove string literal from boolean reference in session-auth example

kgodtfredsen:
https://github.com/zapier/zapier/issues/50045


eliangcs (approved):
I tested and it works! Thanks!

PR #390, opened 2021-06-22
PDE-2564: core(feat): allow using await in inline function source (9.x backport)

eliangcs:
This is a 9.x backport merging into the `release-9.x` branch. I created this new branch `release-9.x` from tag `zapier-platform-cli@9.4.2` so we can release more `9.x` versions from that branch. Once this PR is approved, I'll merge and cut a release **9.5.0**.

This PR addresses two Jira tickets:
- [PDE-2564](https://zapierorg.atlassian.net/browse/PDE-2564): Add `async/await` support for Visual Builder (hackathon project)
- [PDE-2465](https://zapierorg.atlassian.net/browse/PDE-2465): Backport #387 to 9.x to make sure calling `z.request()` in the legacy script is logged

For [PDE-2564](https://zapierorg.atlassian.net/browse/PDE-2564), this patch allows developers to use the `await` operator in Visual Builder's Code Mode. For example, Visual Builder may generate this code for auth testing:

```javascript
const options = {
  url: 'https://auth-json-server.zapier-staging.com/me',
  method: 'GET',
  headers: {
    'X-Api-Key': bundle.authData.api_key
  }
};
return z.request(options)
  .then((response) => {
    response.throwForStatus();
    return response.json;
  });
```

Now with the support for the `await` operator, developers can write this:

```javascript
const options = {
  url: 'https://auth-json-server.zapier-staging.com/me',
  method: 'GET',
  headers: {
    'X-Api-Key': bundle.authData.api_key
  }
};
const response = await z.request(options);
response.throwForStatus();
return response.json;
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '10.16.3' # lambda version as of 2019-11-21
+  - '14'
 before_install:
   - npm install -g yarn
   - pip install --user awscli
--- a/packages/core/src/create-command-handler.js
+++ b/packages/core/src/create-command-handler.js
@@ -18,7 +18,7 @@ const commandHandlers = {
   commands like 'execute', 'validate', 'definition', 'request'.
 */
 const createCommandHandler = compiledApp => {
-  return input => {
+  return async input => {
     const command = input._zapier.event.command || 'execute'; // validate || definition || request
     const handler = commandHandlers[command];
     if (!handler) {
@@ -26,7 +26,7 @@ const createCommandHandler = compiledApp => {
     }
 
     try {
-      return handler(compiledApp, input);
+      return await handler(compiledApp, input);
     } catch (err) {
       return handleError(err);
     }
--- a/packages/core/src/tools/create-lambda-handler.js
+++ b/packages/core/src/tools/create-lambda-handler.js
@@ -204,7 +204,8 @@ const createLambdaHandler = appRawOrPath => {
           // Adds logging for _all_ kinds of http(s) requests, no matter the library
           if (!skipHttpPatch) {
             const httpPatch = createHttpPatch(event);
-            httpPatch(require('http')); // 'https' uses 'http' under the hood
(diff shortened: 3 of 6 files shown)


xavdid (approved):
Nice! My only thought here is if it being async changes our error-handling story at all. I don't think it does (and I see you've got a test), but I could see a situation where wherever this created function is called (the `execute` command handler, I think?) we do a `catch` but don't expect a rejected promise. So, if you haven't, that bears looking into.

The only other consideration isn't related to the content of this PR exactly, but about the backporting. Apps that have been running continuously (most public apps, which probably don't have their lambda functions pruned) could still be running on Node 8. It's possible that they don't work on later node versions, but we don't know it yet. We could be poking the bear by refreshing apps that have been humming along nicely. It's _probably_ fine, but that was an area of concern that popped up as I was thinking through this. 

Also, my guess is that we'll take godzilla apps from 9.x right to 11.x, so we probably don't need a 10.x release, FWIW.

Great job!

eliangcs:
@xavdid thanks for the great feedback! I didn't think that much so I took a second look.

I think we can safely assume the core always calls the created function asynchronously. The command handler named `app` is created here: https://github.com/zapier/zapier-platform/blob/717bbedacf123d2a8e3b72eb67cb0cdda95ca6ab/packages/core/src/create-app.js#L43-L45

In `applyMiddleware`, there's a `resolve` wrapping around the `app()` call: https://github.com/zapier/zapier-platform/blob/717bbedacf123d2a8e3b72eb67cb0cdda95ca6ab/packages/core/src/middleware.js#L88-L92

This means even if `app` doesn't return a Promise, that `resolve` will make it a Promise. So I think we're safe replacing `new Fucntion()` with `new AsyncFunction()`.

Good call on the Node 8 concern! I looked into the database. No Visual Builder apps are running on Node.js 8. I think they all got upgraded in the previously 9.x releases.

xavdid:
Sweet! Carry on 🚢

PR #391, opened 2021-06-28
[CTO-81] Allow for caller to opt out of replace in execute-request.js.

bryanhelmig:
As part of hackathon, this option would be valuable to have when calling into execute request! The reason is that user defined requests could try to exfiltrate `{{bundle.authData.accessToken}}` or `{{process.env.SOME_SECRET}}`...

Ref https://github.com/zapier/zapier/pull/50081.

### DIFF
--- a/packages/core/src/execute-request.js
+++ b/packages/core/src/execute-request.js
@@ -10,7 +10,7 @@ const executeRequest = (input) => {
   if (!options.url) {
     throw new Error('Missing url for request');
   }
-  options.replace = true;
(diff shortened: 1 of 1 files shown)


eliangcs (approved):
The change looks good! Maybe we want to document it in [Shorthand HTTP Requests](https://github.com/zapier/zapier-platform/blob/548e33223715b04f8926bc178cccb14aedbe5cca/packages/cli/README-source.md#shorthand-http-requests) too?

PR #392, opened 2021-06-28
[CTO-81] Document raw request flags.

bryanhelmig:
The sibling to https://github.com/zapier/zapier/pull/50081 which adds the flags for supporting this.

Also explicitly updates the nvmrc Node version to make what Lambda and Travis are expecting.


bryanhelmig:
Pairs well with https://github.com/zapier/zapier-platform/pull/391.

PR #393, opened 2021-06-29
feat(core): incorporate secret-scrubber into core

xavdid:
Incorporates my hackathon project, the secret scrubber, into the platform!

It was mostly a project of moving and improving existing scrubbing code. Now, others can take advantage of it too!

[PDE-2555](https://zapierorg.atlassian.net/browse/PDE-2555)

### DIFF
--- a/packages/core/src/constants.js
+++ b/packages/core/src/constants.js
@@ -23,17 +23,6 @@ const REQUEST_OBJECT_SHORTHAND_OPTIONS = { isShorthand: true, replace: true };
 const DEFAULT_LOGGING_HTTP_ENDPOINT = 'https://httplogger.zapier.com/input';
 const DEFAULT_LOGGING_HTTP_API_KEY = 'R24hzu86v3jntwtX2DtYECeWAB'; // It's ok, this isn't PROD
 
-const SENSITIVE_KEYS = [
-  'api_key',
-  'apikey',
-  'auth',
-  'passwd',
-  'password',
-  'secret',
-  'signature',
-  'token',
-];
-
 const SAFE_LOG_KEYS = [
   'account_id',
   'api_title',
@@ -76,6 +65,5 @@ module.exports = {
   REQUEST_OBJECT_SHORTHAND_OPTIONS,
   RESPONSE_SIZE_LIMIT,
   SAFE_LOG_KEYS,
-  SENSITIVE_KEYS,
   STATUSES,
 };
--- a/packages/core/src/tools/create-logger.js
+++ b/packages/core/src/tools/create-logger.js
@@ -3,19 +3,29 @@
 const _ = require('lodash');
 
 const request = require('./request-client-internal');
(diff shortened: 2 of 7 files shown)


packages/core/src/constants.js
```diff
@@ -23,17 +23,6 @@ const REQUEST_OBJECT_SHORTHAND_OPTIONS = { isShorthand: true, replace: true };
 const DEFAULT_LOGGING_HTTP_ENDPOINT = 'https://httplogger.zapier.com/input';
 const DEFAULT_LOGGING_HTTP_API_KEY = 'R24hzu86v3jntwtX2DtYECeWAB'; // It's ok, this isn't PROD
 
-const SENSITIVE_KEYS = [
```
xavdid, line 26: this logic was moved to the package

packages/core/src/tools/create-logger.js
```diff
@@ -89,66 +99,19 @@ const toStdout = (event, msg, data) => {
...
-          // ignore; Buffer is semi-selective about what types it takes
-        }
-      }
-      return bank;
-    },
-    {}
+  const authData = Object.values(bundle.authData || {});
+  // for the most part, we should censor all the values from authData
+  // the exception is safe urls, which should be filtered out - we want those to be logged
+  const sensitiveAuthData = authData.filter(
```
xavdid, line 107: This logic keeps "safe urls" around. there's code in scrubber to "intelligently" pull sensitive data out of urls which we could use here too, but we're also fine to censor all non-safe urls. 

If we did that, it would pull `my-secret` out of `https://www.example.com/whatever?token=my-secret&foo=bar`. Then instead of censoring the whole url, it would be `https://www.example.com/whatever?token=:censored:3:asdf:&foo=bar`. But, it's also fine as is.

packages/core/test/logger.js
```diff
@@ -72,9 +72,9 @@ describe('logger', () => {
         token: options.token,
         message: 'test',
         data: {
-          password: ':censored:6:a5023f748d:',
```
xavdid, line 75: The tests are all the same, but I changed the default salt for the `censor` function, so the hashes had to be updated.

xavdid:
The CLI tests are still broken, but should auto-resolve once [secret-scrubber-js!2](https://gitlab.com/zapier/secret-scrubber-js/-/merge_requests/2) is released

eliangcs (commented):
Cool work! I'll take another look once you fix the failing test.

I'd love to see we make the secret-scrubber repo public. Is that something you plan to do?

xavdid:
ok, tests pass! I'm glad they caught that node versioning issue. 

And yes, I do want them open sourced! The JS package is already publicly available, so hopefully it's not too much work. We went with Gitlab because we're not sure to what extent we're moving away from Github; that's a thing to circle back to.

eliangcs (approved):
👍

PR #394, opened 2021-06-29
PDE-2556: fix(legacy-scripting-runner): don't make multipart body if `request.files` is empty

eliangcs:
Fixes https://github.com/zapier/zapier/issues/50077.

Say there's an action of which key is `upload`. It has a file field `file` with the following `upload_pre_write` method defined:

```javascript
var Zap = {
  upload_pre_write: function(bundle) {
    bundle.request.files = {};
    return bundle.request;
  }
};
```

The `bundle.request.files = {}` line should make legacy-scripting-runner "cancel" encoding request body in `multipart/form-data` and go back to encode the body in `application/json`.

### DIFF
--- a/packages/legacy-scripting-runner/test/example-app/index.js
+++ b/packages/legacy-scripting-runner/test/example-app/index.js
@@ -851,7 +851,7 @@ const legacyScriptingSource = `
         if (_.isEmpty(bundle.request.files)) {
           // Reach here when the optional file field is empty
           bundle.request.headers['Content-Type'] = 'application/json';
-          bundle.request.url = 'https://httpbin.zapier-tooling.com/post';
+          bundle.request.url = '${HTTPBIN_URL}/post';
           return bundle.request;
         } else {
           // Reach here when the optional file field is filled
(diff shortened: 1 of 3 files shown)


packages/legacy-scripting-runner/test/example-app/index.js
```diff
@@ -861,6 +861,21 @@ const legacyScriptingSource = `
...
+
+        bundle.request.url = '${HTTPBIN_URL}/post';
+
+        // This should make legacy-scripting-runner switch from multipart to
+        // JSON body
+        bundle.request.files = {};
+        bundle.request.data = z.JSON.stringify(data);
+
+        return bundle.request;
+      },
```
eliangcs, line 877: This is the only meaningful change in this file. Others are done by prettier.

eliangcs, line 877: This is the only meaningful change in this file. Others are done by prettier.

packages/legacy-scripting-runner/test/integration-test.js
```diff
@@ -3361,6 +3419,43 @@ describe('Integration Test', () => {
...
+        const json = response.json;
+        should.deepEqual(json.file, [
+          'png',
+          `${HTTPBIN_URL}/image/png`,
+          'image/png',
+        ]);
+        should.equal(json.filename, 'pig.png');
+        should.equal(json.yes, true);
+      });
+    });
```
eliangcs, line 3457: This is the only meaningful change in `integration-test.js`. Others are only auto-formatting done by prettier.

xavdid (approved):

PR #399, opened 2021-07-21
feat(schema) TN-169 Adds canPaginate to BasicHookOperationSchema

cyberwitch:
This is required for hook triggers to paginate with cursors in performList - either for editor dropdowns, or Transfer support


packages/schema/lib/schemas/BasicHookOperationSchema.js
```diff
@@ -51,6 +51,10 @@ BasicHookOperationSchema.properties = {
       },
     },
   },
+  canPaginate: {
+    description: 'Does this endpoint support a page offset?',
```
xavdid, line 55: I realize this is the existing text, but this is a good opportunity to improve it. `canPaginate` also controls whether the `cursorStore` works. So maybe something like:

> Does this trigger support pagination? Either via a numeric "page" offset or via temporary cursor storage.

that's a little wordy, but you get the idea.

xavdid (commented):
Nice! You'll want to fix that readme thing, but it's otherwise good.

packages/schema/lib/schemas/BasicHookOperationSchema.js
cyberwitch: Hmm...now that I think about it, is this description actually correct?  I think `canPaginate` is *only* required for the cursor store, and https://github.com/zapier/zapier/pull/16646 makes it sound like there could be a performance hit if triggers set `canPaginate` but don't use the cursor store.  Maybe we should actually flip it?

> Does this endpoint support pagination via temporary cursor storage?

cyberwitch: I know I've seen CLI triggers that use page offset without setting `canPaginate`, like https://github.com/zapier/zapier-platform-cli-apps/blob/f7c6523b357a792d052f021199ff96cc3612cfeb/apps/mixer/triggers/dynamic_dropdowns/list_follows.js

xavdid: yep, you're correct! I wonder why we even added this flag then - I think it predates the cursor store 🤔

xavdid: ha, In https://github.com/zapier/zapier/pull/16646, I said (in 2018):

> I notice the BasicPollingOperationSchema mentions canPaginate, but that string doesn't show up anywhere in our codebase

So maybe this is all just a happy accident.

packages/cli/README.md
```diff
@@ -2666,7 +2666,7 @@ describe('triggers', () => {
 
 ### Using the `z` Object in Tests
 
-Introduced in `core@11.1.0`, `appTester` can now run arbitrary functions that are not in your app definition:
```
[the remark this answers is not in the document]
cyberwitch: The line's being shortened, not removed.  Not sure why, but it looks like the short version of the line was added to `README-source.md` in https://github.com/zapier/zapier-platform/pull/385/files#diff-3dde54695dbc62f7d2d2212cc889a7aeb3fbffee64355037ceaaba6755b616af and `README.md` is out of sync, maybe from some other branch?  Building my changes included that one.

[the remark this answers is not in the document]
xavdid: huh! I guess something didn't get run when I added it originally. ¯\\\_(ツ)_/¯ sorry about that!

xavdid (approved):
nice update!

cyberwitch:
@xavdid Thanks!

I merged, and then ran `yarn version --minor` as directed in `packages/schema/README.md`.  This didn't publish anything to npm though - it just created a local commit updating the version in `packages/schema/docs/build/schema.md`, `packages/schema/exported-schema.json`, and `packages/schema/package.json` to `11.1.0`.

Do I need to push this commit to `master`, and then create another PR to bump `core/package.json`'s dependency on `zapier-platform-schema` to `11.1.0`, if I want to have a CLI app start using this new schema?

Issue #290, opened 2020-11-17
Wrong Header names in webhook based trigger 


the-csaba:
### Bug Description

Same headers are renamed inside the `bundle.rawRequest.headers` , have an unexpected `Http-` prefix, which makes using it hard. 

Also contrary to the documentation the `bundle.cleanedRequest` only contains the content, no header or method access.

<details>
 <summary>Example headers</summary>

```js
{
  'Http-Host': 'hooks.zapier.com',
  'Http-X-Request-Id': '26433038c459bbb396e829c60757bab1',
  'Http-X-Real-Ip': ';redacted;',
  'Http-X-Forwarded-For': ';redacted;',
  'Http-X-Forwarded-Host': 'hooks.zapier.com',
  'Http-X-Forwarded-Port': '443',
  'Http-X-Forwarded-Proto': 'https',
  'Http-X-Scheme': 'https',
  'Http-X-Original-Forwarded-For': '35.189.17.248',
  'Content-Length': '2972',
  'Http-Accept': '*/*',
  'Http-Accept-Encoding': 'deflate, gzip',
  'Content-Type': 'application/json',
  'Http-Referer': 'https://hooks.zapier.com/hooks/standard/4208012/bfea20d2a2c1451bbccd9a5dcfb3388d/',
  'Http-User-Agent': 'WooCommerce/4.7.0 Hookshot (WordPress/5.5.3)',
  'Http-X-Wc-Webhook-Delivery-Id': '2bc1695b2762fe98193866742c246b9b',
  'Http-X-Wc-Webhook-Event': 'updated',
  'Http-X-Wc-Webhook-Id': '2622',
  'Http-X-Wc-Webhook-Resource': 'order',
  'Http-X-Wc-Webhook-Signature': 'E8QTnfh3/VrBQMQQfE71fbtWI8nQZG47WPwN2bUygT8=',
  'Http-X-Wc-Webhook-Source': ':redacted:',
  'Http-X-Wc-Webhook-Topic': 'order.updated',
  'Http-X-Wordpress-Gmt-Offset': '8'
}
```
</details>

### Reproduction Steps

1. Setup a wenhook based trigger for example WooCommerce
2. send data to the trigger
3. inspect the `bundle`

### Version Info

* CLI version: 10.1.1
* Node.js version: v12.19.0
* OS info: darwin-x64
* `zapier-platform-core` dependency: 10.1.1

xavdid:
[...]

PR #401, opened 2021-07-23
[SUPPORTESC-179] docs(cli): Add information about header format in bundle.rawRequest

armchairlinguist:
Adds a note about how headers are (re)formatted in the bundle, so that devs no longer need to guess/test.
Relevant to #290 - we may improve this in the future, in which case this addition would need to be removed.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -3059,8 +3059,8 @@ new records in your system (add a recipe to the catalog).</p><p>The definition f
   <div class="row-height">
     <div class="col-md-5 col-sm-12 col-height  docs-primary">
       <blockquote>
-<p><code>bundle.rawRequest</code> is only available in the <code>perform</code> for web hooks, <code>getAccessToken</code> for oauth authentication methods, and <code>performResume</code> in a callback action.</p>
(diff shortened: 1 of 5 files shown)


armchairlinguist:
My testing on this indicated that headers are not typically available in `bundle.cleanedRequest`, even though [Line 888](https://github.com/zapier/zapier-platform/pull/401/files#diff-3dde54695dbc62f7d2d2212cc889a7aeb3fbffee64355037ceaaba6755b616afR888) suggests that they may be. Is my testing missing something, or can we say that they are not present?

codebycaleb (changes requested):
The content looks great! Can you run `yarn run docs` from inside `packages/cli/` to generate the rest of the README docs? Then I'll be happy to give the 👍

armchairlinguist:
Oh of course, sorry! I got lazy and did this in the UI, and forgot that step.

codebycaleb (dismissed):

armchairlinguist:
Just to add context, the situation with cleanedRequest is more complex per recent discussions, so I'm going to merge this PR as-is, and any changes there can be separate.

codebycaleb (approved):

PR #402, opened 2021-07-23
feat(core): TN-169 Updates the app tester to support hook with canPaginate for performList

cyberwitch:
In order to make a CLI hook trigger compatible with Transfer, it needs to be able to set `canPaginate` for its `performList` method.  This PR updates the app tester to validate such a schema.

I wasn't sure whether `create-app-tester` is something third party devs might import, or if it's only used internally by us, but to be safe I refactored `shouldPaginate` into its own file rather than refactor `create-app-tester.js`'s module.exports.

### DIFF
--- a/packages/core/src/tools/should-paginate.js
+++ b/packages/core/src/tools/should-paginate.js
@@ -16,7 +16,11 @@ const shouldPaginate = (appRaw, method) => {
   const operation = get(appRaw, methodParts);
 
   if (methodParts[0] === 'triggers') {
-    if (operation.type === 'poll' && methodName === 'perform') {
+    // Polling operations may not specify type
+    if (
+      ['polling', undefined].includes(operation.type) &&
+      methodName === 'perform'
+    ) {
       return !!operation.canPaginate;
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/tools/should-paginate.js
```diff
@@ -0,0 +1,43 @@
...
+// resources.contact.list.operation.perform if it's a resource
+// resources.contact.hook.operation.performList if it's a resource
+
+const shouldPaginate = (appRaw, method) => {
+  const methodParts = method.split('.');
+  const methodName = methodParts.pop();
+  const operation = get(appRaw, methodParts);
+
+  if (methodParts[0] === 'triggers') {
+    if (operation.type === 'poll' && methodName === 'perform') {
```
xavdid, line 19: 🚨  the operation type is `polling`, not `poll` ([docs](https://github.com/zapier/zapier-platform/blob/master/packages/schema/docs/build/schema.md#properties-11))

xavdid: also, the `type` is optional on polling triggers, but (mostly) required on hook triggers (unless they're part of a resource). I don't believe this block (or the tests) cover the missing `type` for non-resource polls.

xavdid (commented):
You're correct that this also needs an update. My comment is correct - it's annoying that this is duplicated, but if I remember correctly, there's not an easy way around it.

> I wasn't sure whether create-app-tester is something third party devs might import

Public devs definitely use it! It's the way to write unit tests, which _hopefully_ everyone is doing 👀 

I left one 🚨, which should be addressed before moving.j

The other thing is `type` not always being required. so there may be more tests needed.

xavdid (approved):
nice! tests look great!

PR #408, opened 2021-08-10
feat(core): TN-320 Allows apps to pass along throttled API errors

cyberwitch:
This PR adds the ability to raise a `ThrottledError`, which requests that the engine not retry this step for a specified number of seconds. It also passes up a header that can be used to auto-throttle 429s for the hinted amount of time.

This PR will work in conjunction with https://github.com/zapier/zapier/pull/52412, which makes the feature previously only available to Python apps available to CLI as well.


packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "zapier-platform-cli",
-  "version": "11.1.1",
+  "version": "11.2.0",
```
xavdid, line 3: 🚨 don't increment this, we'll do it when we cut a release

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "zapier-platform-core",
-  "version": "11.1.1",
+  "version": "11.2.0",
```
xavdid, line 3: same here

packages/cli/README-source.md
```diff
@@ -1475,6 +1476,28 @@ const yourAfterResponse = (resp) => {
 };
 ```
 
+### Handling Throttled Requests
+
+Since v11.2.0, there are two types of errors that can cause Zapier to throttle an operation and retry at a later time.
```
xavdid, line 1481: this part is fine, since this will likely be released in 11.2.0

xavdid (commented):
looks awesome! if you pull out the changes to the `version` we're good to merge here.

cyberwitch:
@xavdid Thanks for the feedback! I think this is ready for re-review.

xavdid (approved):

PR #410, opened 2021-08-13
fix(legacy-scripting-runner): properly call afterResponse middlewares with z and bundle

xavdid:
We were seeing an issue where legacy-scrpiting-runner was throwing an error during auth refresh because `z` was undefined. See [slack](https://zapier.slack.com/archives/C5Z9BP4U9/p1628883793294600?thread_ts=1628871527.280000&cid=C5Z9BP4U9). The fix is to call the middleware with all the args we expect; simple enough.

I verified the fix in the affected app, but didn't write a test (see below).

[PDE-2635](https://zapierorg.atlassian.net/browse/PDE-2635)

---

I would love a test here, but for some reason, I was having trouble writing it. I had something like:

```js
it('middleware', () => {
      const appDef = _.cloneDeep(appDefinition);
      appDep.afterResponse.push((response, z, bundle)=>{ throw new z.errors.RefreshAuthError('Authentication needs refreshing'); })
      const _compiledApp = schemaTools.prepareApp(appDef);
      const _app = createApp(appDef);
      // ...
```

but my custom function wasn't running. I debugged for like an hour and then figured it wasn't worth the time. If you know how to do it offhand, you're welcome to add a commit.


packages/cli/README.md
```diff
@@ -2666,7 +2666,7 @@ describe('triggers', () => {
 
 ### Using the `z` Object in Tests
 
-Introduced in `core@11.1.0`, `appTester` can now run arbitrary functions that are not in your app definition:
```
xavdid, line 2669: not sure why this changed (I was on a fresh master), but it now correctly reflects `readme-source` ¯\\\_(ツ)_/¯

eliangcs (dismissed):
Thanks for fixing! I added a test to make sure `RefreshAuthError` is correctly thrown.

eliangcs (approved):

PR #411, opened 2021-08-16
PDE-2636 fix(tests): failing smoke tests on package size

eliangcs:
NPM registry seems to no longer provide a `content-length` header for `HEAD` requests. This broke some of our smoke tests where we send a `HEAD` request to NPM registry to get the package size.

Here's an example failed build on Travis: https://app.travis-ci.com/github/zapier/zapier-platform/builds/235320597

This PR fixes the issue by replacing the `HEAD` request with a `GET` request.


xavdid (approved):
dang! good catch

PR #412, opened 2021-08-16
PDE-2658 fix(ci): faster CI builds

eliangcs:
We're on the "One Concurrent Job" plan (meaning: no concurrency) on Travis so the tests would be faster if we don't have too many jobs.

This PR reduces the number of Travis jobs from 10 to 5, shortening the build time from ~23 mins to [~18 mins](https://app.travis-ci.com/github/zapier/zapier-platform/builds/235541078).

While I was there, I bumped secret-scrubber from 1.0.1 to 1.0.2 to fix a compatibility issue on Node.js 12.


eliangcs:
I'm testing a new Travis configuration to see if tests can be faster. But I'm running into this error on [Travis](https://app.travis-ci.com/github/zapier/zapier-platform/jobs/531561397#L414-L415) on Node.js 12:

```
/home/travis/build/zapier/zapier-platform/node_modules/@zapier/secret-scrubber/lib/utils.js:25
    const salted = val + (process.env.ZAPIER_SCRUBBER_SALT ?? 'averysecretsalt');
```

So `??` is only supported by Node.js 14+. But we didn't have this error with previous builds. I wonder why this PR breaks. @xavdid

xavdid:
I'm not sure why we didn't see it until now, but version 1.0.2 of secret scrubber should have fixed this: [CHANGELOG](https://gitlab.com/zapier/team-developer-platform/secret-scrubber-js/-/blob/main/CHANGELOG.md).

I see that the core's `package.json` depends on `^1.0.1`. Maybe manually bump it to 1.0.2? I would guess 1.0.1 is stuck in the `yarn.lock`.

xavdid (approved):
Nice! Any improvement is something.

I would think long-term we'd want to move off travis, probably either to gitlab or github actions. Really depends on where this project will live and what gitlab stuff we can open source.

PR #414, opened 2021-08-16
feat(legacy-scripting-runner): INT-15611 - Add isBulkRead to convertBundleMeta keys

kreddlear:
Integrations is enabling some WB > CLI converted apps to work with Transfer. Transfer passes along the `bundle.meta` flag `isBulkRead`, so ideally it would be helpful to use this flag to add Transfer-specific logic to existing legacy scripting. Since that flag isn't currently defined in the `convertBundleMeta` function, it gets wiped from the bundle.

This PR adds it in so that ideally it will stick around and be usable in legacy scripting.

I also added it to the tests, and to one part of the integration test. Please let me know if I need to add it to anywhere else.

Slack conversation here: https://zapier.slack.com/archives/C5Z9BP4U9/p1629145497338300


packages/schema/docs/build/schema.md
```diff
@@ -256,19 +252,15 @@ Key | Required | Type | Description
...
 * ```
-  {
-    authorizeUrl: { require: 'some/path/to/file.js' },
+  { authorizeUrl: { require: 'some/path/to/file.js' },
     getAccessToken: { require: 'some/path/to/file2.js' },
     refreshAccessToken: { require: 'some/path/to/file3.js' },
     scope: 'read/write',
-    autoRefresh: true
-  }
+    autoRefresh: true }
```
xavdid, line 263: this big formatting change is usually the result of an out-of-date `prettier` installation.

packages/legacy-scripting-runner/bundle.js
```diff
@@ -224,6 +224,7 @@ const convertBundleMeta = (meta) => {
       meta.first_poll === undefined ? meta.isPopulatingDedupe : meta.first_poll,
     limit: meta.limit,
     page: meta.page,
+    isBulkRead: meta.isBulkRead,
```
xavdid, line 227: you might do `meta.isBulkRead || false` so that it's always present, but reacts accordingly if `isBulkRead` isn't in the incoming meta. IIRC we're only supplying that to certain apps to start.

packages/legacy-scripting-runner/test/bundle.js
```diff
@@ -52,6 +53,7 @@ describe('bundleConverter', () => {
       standard_poll: true,
       page: 1,
       limit: 100,
+      isBulkRead: false,
```
xavdid, line 56: if there's a default in the code itself, you may not need all these additions

xavdid (dismissed):

packages/schema/docs/build/schema.md
kreddlear: yikes, yeah I didn't realize the pre-commit would prettier that automatically! I'll make sure that's updated, thanks

packages/legacy-scripting-runner/test/bundle.js
kreddlear: I was hoping that would be the case, but without these additions, the tests failed. I only added this in places where the tests were failing -- if there's a better way to do it I'd definitely be for that instead!

packages/schema/docs/build/schema.md
kreddlear: possibly silly question but -- where would the relevant prettier installation be, or alternately how do the schema docs get rebuilt? I updated prettier locally and `prettier -v` now returns the most recent version, but the pre-commit doesn't seem to be making further changes to the schema 🤔

packages/legacy-scripting-runner/test/bundle.js
xavdid: ok, no worries! Not sure what's up there, but it's probably fine.

packages/schema/docs/build/schema.md
xavdid: it would be in `node_modules` at the repo root. it's honestly not a big deal if this gets reformatted in weird ways, but I figured it was worth mentioning

xavdid:
There's a fix for the tests failing in `master`, so you can merge back.

packages/schema/docs/build/schema.md
kreddlear: figured it out! I was using an old node version from working on an older app. should be all good now.

kreddlear:
Merged master in and figured out the issue that was causing the reformatting -- waiting to make sure tests now pass and this should be good to go.

xavdid (approved):
nice! it's always something like that.

PR #418, opened 2021-09-09
PDE-2704 fix(legacy-scripting-runner): add support for `z.request({ json: true })`

eliangcs:
In Web Builder scripting, developers could pass `json: true` to `z.request()`:

```js
var Zap = {
  KEY_write: function(bundle) {
       var data = z.JSON.parse(bundle.request.data);
       data.hello = 'world';
       var response = z.request({
           method: 'POST',
           url: 'https://httpbin.zapier-tooling.com/post',
           json: true,
           body: data  // <- an object
       });
       return response.content;  // <- an object
  }
};
```

The returned result should be an object containing `{"json": {"hello": "world"}}`. This PR fixes legacy-scripting-runner to recreate this behavior.

### DIFF
--- a/packages/legacy-scripting-runner/zfactory.js
+++ b/packages/legacy-scripting-runner/zfactory.js
@@ -24,8 +24,12 @@ const convertBundleRequest = (bundleOrBundleRequest) => {
     };
   }
 
-  bundleRequest.qs = bundleRequest.params || {};
-  bundleRequest.body = bundleRequest.data || '';
+  if (!bundleRequest.qs && bundleRequest.params) {
+    bundleRequest.qs = bundleRequest.params;
+  }
+  if (!bundleRequest.body && bundleRequest.data) {
+    bundleRequest.body = bundleRequest.data;
+  }
 
   delete bundleRequest.params;
   delete bundleRequest.data;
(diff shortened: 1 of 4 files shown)


packages/legacy-scripting-runner/zfactory.js
```diff
@@ -24,8 +24,12 @@ const convertBundleRequest = (bundleOrBundleRequest) => {
...
   }
 
-  bundleRequest.qs = bundleRequest.params || {};
-  bundleRequest.body = bundleRequest.data || '';
+  if (!bundleRequest.qs && bundleRequest.params) {
+    bundleRequest.qs = bundleRequest.params;
+  }
+  if (!bundleRequest.body && bundleRequest.data) {
+    bundleRequest.body = bundleRequest.data;
+  }
```
eliangcs, line 32: Borrowed this from the [code](https://github.com/zapier/lambda-wrapper/blob/aeb3fc53fe07ad0990fe1fb6b3daea7461c623a3/lambda-handlers/evalbox-javascript/utils.js#L236-L240.) we used for WB.

xavdid (dismissed):

szchenghuang (approved):
Thanks for picking this up and enhancing it.

PR #419, opened 2021-09-10
[SUPPORTESC-192] docs(cli): A bit more detail around inputData and inputDataRaw

armchairlinguist:
I had noticed the phrase "is kind of like" in comparing these two and wanted to give a bit more context on differences - so added another type of processing done on raw input and related examples.

Very open to changes if there's more detail I should add, or a better way to present this!

It looks like the pre-commit processing also generated a lot of schema changes, which aren't part of the the work I did, but seem to be mostly formatting details, so hopefully not a problem. I tried making sure I was on the latest version of everything, but the changes persisted.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -801,7 +801,7 @@ This object holds the user's auth details and the data for the API requests.
 
 ### `bundle.inputData`
 
-`bundle.inputData` is user-provided data for this particular run of the trigger/search/action, as defined by the `inputFields`. For example:
+`bundle.inputData` is user-provided data for this particular run of the trigger/search/create, as defined by the `inputFields`. For example:
 
 ```js
 {
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -801,23 +801,25 @@ This object holds the user's auth details and the data for the API requests.
 
 ### `bundle.inputData`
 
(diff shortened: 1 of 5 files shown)


packages/cli/README-source.md
```diff
@@ -801,27 +801,31 @@ This object holds the user's auth details and the data for the API requests.
 
 ### `bundle.inputData`
 
-`bundle.inputData` is user-provided data for this particular run of the trigger/search/create, as defined by the inputFields. For example:
+`bundle.inputData` is user-provided data for this particular run of the trigger/search/action, as defined by the `inputFields`. For example:
```
eliangcs, line 804: I suggest we change `action` back to `create`. `create` is a more specific term in the CLI while `action` has a broader meaning and sometimes includes `search` and `create`.

eliangcs (commented):
Thanks for improving the docs! The PR looks good to me except for the `action` thing. Once that's fixed, I can re-approve.

Don't worry about the schema changes. I believe they were caused by the different Node versions we're using. We didn't spend too much investigating because that won't break anything.

packages/cli/README-source.md
armchairlinguist: Makes sense, sorry about that! I'll get it fixed up.

xavdid: ^ @megan-driscoll

eliangcs (approved):
Thanks for the improvement! Feel free to merge.

PR #423, opened 2021-09-15
PDE-2751 fix(legacy-scripting-runner): Empty `request.data` should send an empty request body instead of an empty object

eliangcs:
## Legacy scripting code to reproduce the issue

```js
var Zap = {
  KEY_pre_write: function(bundle) {
    bundle.request.url = 'https://auth-json-server.zapier-staging.com/echo';
    bundle.request.data = '';
    // or bundle.request.data = null;
    return bundle.request;
  }
};
```

## Expected behavior

The request body should be an empty string.

## Current behavior

The request body is JSON-encoded empty object, i.e., `{}`.

### DIFF
--- a/packages/legacy-scripting-runner/test/example-app/index.js
+++ b/packages/legacy-scripting-runner/test/example-app/index.js
@@ -642,6 +642,12 @@ const legacyScriptingSource = `
         return bundle.request;
       },
 
+      movie_pre_write_request_data_empty_string: function(bundle) {
+        bundle.request.url = '${AUTH_JSON_SERVER_URL}/echo';
+        bundle.request.data = '';
+        return bundle.request;
+      },
+
       movie_write_default_headers: function(bundle) {
         bundle.request.url = '${HTTPBIN_URL}/post';
         bundle.request.data = z.JSON.stringify({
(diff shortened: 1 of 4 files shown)


xavdid (dismissed):

packages/legacy-scripting-runner/test/example-app/index.js
```diff
@@ -313,7 +313,7 @@ const legacyScriptingSource = `
       },
 
       movie_pre_poll_env_var: function(bundle) {
-        bundle.request.url = 'https://{{process.env.SECRET_HTTPBIN_URL}}/get';
+        bundle.request.url = '{{process.env.SECRET_HTTPBIN_URL}}/get';
```
eliangcs, line 316: I did this so we can spin up httpbin and auth-json-server locally and run the tests really fast:

```
$ AUTH_JSON_SERVER=http://localhost:3000 HTTPBIN_URL=http://localhost:3001 yarn test
```

szchenghuang: Given this `HTTPBIN_URL=http://localhost:3001,` it seems a need to rename the variable to something more generic.

szchenghuang (approved):

eliangcs:
> What if we say `if (!result.data)`?

@szchenghuang good question! I've tested and found WB behaves this way if you set `request.data` to different values in a `KEY_pre_(poll|write|search)` method:

1. `request.data === undefined` -> fall back to encode the input field values in the request body. For example, `request.body` would be `JSON.stringify(bundle.action_fields)` for `KEY_pre_write`.
2. `request.data === null || request.data === ''` -> `request.body` should be an empty string.
3. `request.data` is an object -> `request.body` should be `JSON.parse(request.data)`.

So `if (!result.data)` won't give us the desired behavior because of 1.

PR #424, opened 2021-09-16
PDE-2737 fix(legacy-scripting-runner): duplicate headers with session auth

eliangcs:
## Legacy Web Builder app to reproduce the issue

Session auth in headers with the following auth mapping:

```json
{
  "x-api-key": "{{api_key}}",
  "content-type": "application/json"
}
```

Then run an action with **no scripting**.

## Expected behavior

Only one `content-type: application/json` header should be sent.

## Current behavior

Two duplicate `content-type: application/json` headers are sent.

## Related PRs

Related to the previous fixes with API Key auth:

- https://github.com/zapier/zapier-platform/pull/364
- https://github.com/zapier/zapier-platform/pull/365

### DIFF
--- a/packages/legacy-scripting-runner/middleware-factory.js
+++ b/packages/legacy-scripting-runner/middleware-factory.js
@@ -20,6 +20,28 @@ const renderTemplate = (templateString, context) => {
   return _.template(templateString, options)(finalContext);
 };
 
+const getLowerHeaders = (headers) =>
+  Object.entries(headers).reduce((result, [k, v]) => {
+    result[k.toLowerCase()] = v;
+    return result;
+  }, {});
+
+const applyAuthMappingInHeaders = (authMapping, req, authData) => {
+  const lowerHeaders = getLowerHeaders(req.headers);
+  Object.entries(authMapping).forEach(([k, v]) => {
+    const lowerKey = k.toLowerCase();
+    if (!lowerHeaders[lowerKey]) {
+      req.headers[k] = renderTemplate(v, authData);
+    }
+  });
+};
+
+const applyAuthMappingInQuerystring = (authMapping, req, authData) => {
+  Object.entries(authMapping).forEach(([k, v]) => {
+    req.params[k] = req.params[k] || renderTemplate(v, authData);
+  });
+};
+
 const createBeforeRequest = (app) => {
   const authType = _.get(app, 'authentication.type');
   const authMapping = _.get(app, 'legacy.authentication.mapping');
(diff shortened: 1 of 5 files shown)


packages/legacy-scripting-runner/middleware-factory.js
```diff
@@ -20,35 +20,51 @@ const renderTemplate = (templateString, context) => {
...
+    result[k.toLowerCase()] = v;
+    return result;
+  }, {});
+
+const applyAuthMappingInHeaders = (authMapping, req, authData) => {
+  const lowerHeaders = getLowerHeaders(req.headers);
+  Object.entries(authMapping).forEach(([k, v]) => {
+    const lowerKey = k.toLowerCase();
+    if (!lowerHeaders[lowerKey]) {
+      req.headers[k] = renderTemplate(v, authData);
```
benzapier, line 34: Seems like we should always use a header/value from `authData`, and discard/replace an existing header that's already present within the request?  

```js
req.headers.secret_key = 'some default value';
usersAuthData = { secret_key: "the user's stored/saved auth key" };
applyAuthMappingInHeaders(mapping, req, usersAuthData);
```

packages/legacy-scripting-runner/middleware-factory.js
```diff
@@ -20,35 +20,51 @@ const renderTemplate = (templateString, context) => {
...
+    }
+  });
+};
+
+const applyAuthMappingInQuerystring = (authMapping, req, authData) => {
+  Object.entries(authMapping).forEach(([k, v]) => {
+    req.params[k] = req.params[k] || renderTemplate(v, authData);
```
benzapier, line 41: Should the two items on the right-hand be switched, so that a user's authData value replaces anything already present in the request?  

`req.params[k] = renderTemplate(v, authData) || req.params[k];`

But maybe that breaks something else, since this is the current/existing behavior

xavdid (dismissed):
Looks good!

Re: Ben's concerns- I don't think users will typically have standard headers in their auth, so IMO there's not much concern for clobbering there.

packages/legacy-scripting-runner/middleware-factory.js
eliangcs: Great callout! I did another test on WB. WB seems to always override the request headers with `Content-Type: application/json; charset=utf-8` and `Accept: application/json`. So regardless what `Content-Type` the `authData` may have, the header is always `Content-Type: application/json; charset=utf-8` before the bundle goes into a `pre_(poll|write|search)` method.

packages/legacy-scripting-runner/middleware-factory.js
eliangcs: > But maybe that breaks something else, since this is the current/existing behavior

Yep, let's keep the current behavior unless we found WB has a different behavior.

szchenghuang:
Approving per requested by @eliangcs for a review, but it'd be nice to have @benzapier take another look for further feedback.

szchenghuang (approved):

PR #426, opened 2021-09-17
[SUPPORTESC-198] docs(cli): Update return types for current platform and add a few details

armchairlinguist:
* Removed an outdated link to v2 documentation
* Clarify return type behavior for search
* Add note about the function of empty arrays for hook triggers


xavdid (approved):
Love these docs updates!

PR #427, opened 2021-09-17
fix(core): handle nullish values in auth data

xavdid:
fixes [PDE-2771](https://zapierorg.atlassian.net/browse/PDE-2771)

Secret scrubber only takes a `(string | number)[]`, so we have to filter out nullish values before they go in. Secret scrubber's `recruseExtract` does this, but only if we use it!

Tests won't pass until [zapier/team-developer-platform/secret-scrubber-js!3](https://gitlab.com/zapier/team-developer-platform/secret-scrubber-js/-/merge_requests/3) is merged.

### DIFF
--- a/packages/core/src/tools/create-logger.js
+++ b/packages/core/src/tools/create-logger.js
@@ -11,7 +11,11 @@ const {
   SAFE_LOG_KEYS,
 } = require('../constants');
 const { unheader } = require('./http');
-const { scrub, findSensitiveValues } = require('@zapier/secret-scrubber');
+const {
+  scrub,
+  findSensitiveValues,
+  recurseExtract,
+} = require('@zapier/secret-scrubber');
 // not really a public function, but it came from here originally
 const { isUrlWithSecrets } = require('@zapier/secret-scrubber/lib/convenience');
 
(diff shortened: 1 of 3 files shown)


packages/core/src/tools/create-logger.js
```diff
@@ -101,12 +105,15 @@ const toStdout = (event, msg, data) => {
...
   const bundle = event.bundle || {};
-  const authData = Object.values(bundle.authData || {});
+  const authData = bundle.authData || {};
   // for the most part, we should censor all the values from authData
   // the exception is safe urls, which should be filtered out - we want those to be logged
-  const sensitiveAuthData = authData.filter(
-    (item) => !isUrl(item) || isUrlWithSecrets(item)
-  );
+  const sensitiveAuthData = recurseExtract(authData, (key, value) => {
+    if (isUrl(value) && !isUrlWithSecrets(value)) {
```
xavdid, line 112: logic is the same, but this felt more readable.

packages/core/src/tools/data.js
```diff
@@ -153,7 +153,7 @@ const flattenPaths = (data, { preserve = {} } = {}) => {
 
 // A simpler, and memory-friendlier version of _.truncate()
 const simpleTruncate = (string, length, suffix) => {
-  if (string === undefined) {
+  if (string == null) {
```
xavdid, line 156: this now returns for `null` and `undefined`; before we were converting `null` to `''`, which was weird.

packages/core/test/logger.js
```diff
@@ -288,7 +288,7 @@ describe('logger', () => {
 
   // this test fails because the function that creates the sensitive bank doesn't
   // recurse to find all sensitive values
-  it.skip('should replace sensitive data that nested', () => {
```
xavdid, line 291: 🎉

xavdid (commented):
ok! Deployed secret scrubber and this is looking good. Now we don't explode when someone tries to log `null`.

eliangcs (approved):

PR #428, opened 2021-09-22
SUPPORTESC-214 perf(core): Speed up cleaner.js

codebycaleb:
Given an object with M keys and a bank with N keys, we call replacer M * N times. The actual replacer is pretty light-weight, except for its usage of Regex. With a bank size of ~4000 keys and a object size of ~500 keys, we can see 2,000,000 compilations of Regex. It can cause quite a slowdown!

This change will still result in N * M calls to replacer, but only N compilations of Regex.


xavdid (approved):
awesome! great investigation

PR #432, opened 2021-10-11
feat(schema) support hook to poll triggers INT-15776

casshill:
Add support for hook_to_poll style triggers (new schema)
* soft launch by hiding the doc references ie. this pr should not generate changes to the docs
* require performList and exclude perform. 
  * Perform is a no-op for hook-to-poll triggers, the live Zap run calls performList (since it's essentially polling)
  * Should performSubscribe and performUnsubscribe be required? I think they are only optional for legacy hook support

### DIFF
--- a/packages/schema/exported-schema.json
+++ b/packages/schema/exported-schema.json
@@ -699,7 +699,7 @@
         "type": {
           "description": "Must be explicitly set to `\"hook\"` unless this hook is defined as part of a resource, in which case it's optional.",
           "type": "string",
-          "enum": ["hook", "hook_to_poll"],
+          "enum": ["hook"],
           "docAnnotation": {
             "required": {
               "type": "replace",
@@ -1091,6 +1091,9 @@
             },
             {
               "$ref": "/BasicHookOperationSchema"
+            },
+            {
+              "$ref": "/BasicHookToPollOperationSchema"
             }
           ]
         }
--- a/packages/schema/exported-schema.json
+++ b/packages/schema/exported-schema.json
@@ -1063,6 +1063,87 @@
       },
       "additionalProperties": false
     },
+    "BasicHookToPollOperationSchema": {
+      "id": "/BasicHookToPollOperationSchema",
+      "description": "Represents the inbound mechanics of hook to poll style triggers. Defers to list for fields.",
+      "type": "object",
+      "required": ["performList"],
+      "properties": {
+        "type": {
+          "description": "Must be explicitly set to `\"hook\"` unless this hook is defined as part of a resource, in which case it's optional.",
+          "type": "string",
+          "enum": ["hook_to_poll"],
+          "required": {
+            "type": "replace",
+            "value": "**yes** (with exceptions, see description)"
+          }
+        },
+        "resource": {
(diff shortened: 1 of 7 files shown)


xavdid (commented):
@casshill is this ready for review? It's still marked as draft, so I wasn't sure. 

In any case, looks good! I'm not sure how much you care about hiding the new value from the docs, but we have _some_ code for that in here: 

https://github.com/zapier/zapier-platform/blob/e4cacfaa0b2e2bd8ede621905fc291fce71e0eae/packages/schema/lib/utils/buildDocs.js#L121-L136

We don't have anything for hiding a specific enum, but we could add it.

casshill:
> @casshill is this ready for review? It's still marked as draft, so I wasn't sure.
> 
> In any case, looks good! I'm not sure how much you care about hiding the new value from the docs, but we have _some_ code for that in here:
> 
> https://github.com/zapier/zapier-platform/blob/e4cacfaa0b2e2bd8ede621905fc291fce71e0eae/packages/schema/lib/utils/buildDocs.js#L121-L136
> 
> We don't have anything for hiding a specific enum, but we could add it.

It's ready for feedback and suggestions :D 
I thought I'd leave it as draft until we decided how/when we wanted to release it - thanks I'll take a look at the snippet you provided

packages/schema/lib/schemas/TriggerSchema.js
```diff
@@ -34,6 +35,7 @@ module.exports = makeSchema(
         anyOf: [
           { $ref: BasicPollingOperationSchema.id },
           { $ref: BasicHookOperationSchema.id },
+          { $ref: BasicHookToPollOperationSchema.id },
```
casshill, line 38: I don't think we can hide just this 3rd item from the docs, we'd have to hide the full property (which isn't what we want)

packages/schema/lib/utils/buildDocs.js
```diff
@@ -27,7 +27,9 @@ const walkSchemas = (InitSchema, callback) => {
 const collectSchemas = (InitSchema) => {
   const schemas = {};
   walkSchemas(InitSchema, (Schema) => {
-    schemas[Schema.id] = Schema;
+    if (!_.get(Schema, 'schema.docAnnotation.hide')) {
```
casshill, line 30: looks like I could add support for hiding docAnnotations at the top level

packages/schema/exported-schema.json
```diff
@@ -1063,6 +1063,87 @@
...
+          "enum": ["hook_to_poll"],
+          "required": {
+            "type": "replace",
+            "value": "**yes** (with exceptions, see description)"
+          }
+        },
+        "resource": {
+          "description": "Optionally reference and extends a resource. Allows Zapier to automatically tie together samples, lists and hooks, greatly improving the UX. EG: if you had another trigger reusing a resource but filtering the results.",
+          "$ref": "/KeySchema"
+        },
```
casshill, line 1084: Should I take this bit out? Not super sure how it works, but could be a bit complicated to support resources?

xavdid: It's doable, but it will take updates in core too. I don't love resources, so I vote circle back to them as needed

xavdid (commented):
Nice! this is looking good.

I haven't taken a super close look, but if you've done some testing then this should be safe enough! We'll want to push a test app to staging and actually make sure it all runs, but it should be pretty straightforward.

xavdid (approved):

PR #444, opened 2021-11-04
fix(legacy-scripting-runner): make sure redirect_uri is available for oauth refreshes

xavdid:
Fixes [PDE-2871](https://zapierorg.atlassian.net/browse/PDE-2871), where a partner's API depends on `redirect_uri` always being there (even for refresh)

I'm _pretty_ sure this'll fix it, but it's worth double checking the GL links in the jira ticket.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -1051,6 +1051,7 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
     body.client_id = clientId;
     body.client_secret = clientSecret;
     body.refresh_token = bundle.authData.refresh_token;
+    body.redirect_uri = bundle.inputData.redirect_uri;
     body.grant_type = 'refresh_token';
 
     let result;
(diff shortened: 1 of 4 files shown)


eliangcs (approved):

packages/legacy-scripting-runner/index.js
```diff
@@ -1071,6 +1072,7 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
       params.client_id = clientId;
       params.client_secret = clientSecret;
       params.refresh_token = bundle.authData.refresh_token;
+      body.redirect_uri = bundle.inputData.redirect_uri;
```
eliangcs, line 1075: Should we use `params` here?

```suggestion
      params.redirect_uri = bundle.inputData.redirect_uri;
```

eliangcs (dismissed):

packages/legacy-scripting-runner/index.js
xavdid: ah yeah, probably! good call, i'll fix.

eliangcs (approved):

PR #445, opened 2021-11-05
PDE-2872 fix(legacy-scripting-runner): `z.dehydrateFile(url, {})` should not send auth

eliangcs:
## Legacy scripting code to reproduce the issue

```js
var Zap = {
  email_catch_hook: {
    bundle.cleaned_request.file = z.dehydrateFile('https://httpbin.zapier-tooling.com/get', {});
    return bundle.cleaned_request;
  }
};
```

## Expected behavior

When hydrated, we should NOT see the auth headers or params in the request.

But if you do `z.dehydrateFile('https://httpbin.zapier-tooling.com/get')` (omitting the second argument for request options), we _should_ see the auth headers or params.

## Current behavior

When hydrated, we see the auth headers or params in the request. Some server doesn't like it when the URL you're requesting for doesn't require auth but you give it auth info. For example, AWS S3 would give you HTTP 400 error saying `Unsupported Authorization Type`.

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -1389,15 +1389,17 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
 
   const runHydrateFile = (bundle) => {
     const meta = bundle.inputData.meta || {};
+
+    // Legacy z.dehydrateFile(url, request, meta) behavior: if request argument
+    // is provided, the dev is responsible for doing auth themselves, so we use
+    // an internal request client to avoid running the app's http middlewares.
+    const request =
+      bundle.inputData.request === null ||
+      bundle.inputData.request === undefined
+        ? zcli.request
+        : createInternalRequestClient(input);
+
     const requestOptions = bundle.inputData.request || {};
-
-    // Legacy z.dehydrateFile(url, request, meta) behavior: if request argument is
-    // provided, the dev is responsible for doing auth themselves, so we use an internal
-    // request client to avoid running the app's http middlewares.
-    const request = _.isEmpty(requestOptions)
-      ? zcli.request
-      : createInternalRequestClient(input);
-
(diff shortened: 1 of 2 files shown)


eliangcs:
@xavdid let's ship this together with your PR https://github.com/zapier/zapier-platform/pull/444 in version 3.8.4. I can do it on Monday if you don't do it on your Friday.

packages/legacy-scripting-runner/index.js
```diff
@@ -1389,15 +1389,17 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
...
-    // provided, the dev is responsible for doing auth themselves, so we use an internal
-    // request client to avoid running the app's http middlewares.
-    const request = _.isEmpty(requestOptions)
-      ? zcli.request
-      : createInternalRequestClient(input);
+    // Legacy z.dehydrateFile(url, request, meta) behavior: if request argument
+    // is provided, the dev is responsible for doing auth themselves, so we use
+    // an internal request client to avoid running the app's http middlewares.
+    const request =
+      bundle.inputData.request === null ||
```
xavdid, line 1397: nit, but you can do `bundle.inputData.request == null` (double eq), which covers both `null` and `undefined`

xavdid (approved):
Left a comment, but it's not worth fixing and having to wait for another review. So think of it as an FYI.

xavdid:
oh, and i've got a short friday so I'll let you handle the release. I added your PR to the changelog in my branch, so you should be able to approve+merge mine, merge yours, and do the release. I can also check in during my sunday evening to clear any blockers so this gets out on monday

packages/legacy-scripting-runner/index.js
eliangcs: I'll merge this PR as-is because I don't want to wait for another review. But that's another "good" thing to know about JavaScript. Thanks!

PR #447, opened 2021-11-09
PDE-2875 fix(legacy-scripting-runner): handle string array output from a scriptless create/action

benzapier:
Some converted apps possess actions which return an array, rather than an object, which causes the platform to raise an error `JSON results object could not be located.`

In these situations, the legacy scripting runner should return an object `{ "message": results[0] }`

### DIFF
--- a/packages/legacy-scripting-runner/index.js
+++ b/packages/legacy-scripting-runner/index.js
@@ -860,6 +860,12 @@ const legacyScriptingRunner = (Zap, zcli, input) => {
 
         try {
           result = zcli.JSON.parse(response.content);
+          if (postEventName) {
+            result = await parseFinalResult(result, {
+              key,
+              name: postEventName,
+            });
+          }
         } catch {
           result = {};
         }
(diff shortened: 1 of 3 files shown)


eliangcs:
@benzapier thanks for adding the test! It appears the test is already passing without changing legacy-scripting-runner, so I believe there should be a different way to replicate the issue.

After some investigation, the error only occurs when the action is "scriptless".

If you're using a `post_write` method returning an array of string, such as `['foo', 'bar']`, legacy-scripting-runner is able to resolve it to `{message: 'foo'}` correctly. However, if there's no `post_write` and the server returns a response with a JSON body `['foo', 'bar']`, the error `JSON results object could not be located` occurs. So to replicate the issue, we need to make a "scriptless" action and have the server return the array directly. I'll push commits directly to this PR to do that.

eliangcs:
OK, I pushed 3fbe412. That should fix the issue. To summarize, here's how you reproduce the issue.

## Scenario to reproduce the issue

- A scriptless action. No `pre_write`, no `post_write`.
- The server returns an array of strings, such as `['foo', 'bar']`, as the response body.

## Expected behavior

The result of the action should be an object `{ message: 'foo' }`.

## Current behavior

Error: `JSON results object could not be located.`

xavdid (approved):
Nice job, y'all!

PR #437, opened 2021-10-26
(chore/experimental): Setup CI using Github Actions

benzapier:
Exploring possible usage of Github Actions to run the CI tests

PR #453, opened 2021-11-26
PDE-2893 feat(core): improve `z.stashFile()` (9.x backport)

eliangcs:
Rewrites `z.stashFile()` with various improvements and bug fixes. This is a backport to 9.x. https://github.com/zapier/zapier-platform/pull/454 is the other PR targeting the latest 11.x.

Functional improvements:

- `knownLength` isn't required anymore for _any_ kinds of streams, including `z.request({ url, raw: true })` and `fs.createReadStream()`.  This is done by downloading the data to a temp file and getting the size from the temp file.

- Detect file types more smartly. `z.stashFile()` now detects the file type from the `Content-Type` header or the filename extension. That means users will see less `application/octet-stream`!

- Generate filenames more smartly. `z.stashFile()` now gets the filename from the `Content-Disposition` header or the URL. If necessary, it also infers the filename extension from `Content-Type`. That means users will see less `unnamedfile`!

- Requesting for S3 pre-signed POST data and downloading now happen concurrently, so the performance is slightly better.

Bug fixes:

- Fix `MalformedPOSTRequest` error. The error happened when the response was gzipped (`Content-Encoding: gzip`) and had a smaller `Content-Length`.  When the `Content-Length` was different from the actual file size, `z.stashFile()` generated badly-formatted form data, which caused S3 to return a 400 error saying `MalformedPOSTRequest`.

Code quality improvements:

- More test coverage. Previously, the test code for `z.stashFile()` missed out the crucial part to test. It didn't verify the form data sent to S3 and the download response sent from S3. The new test code does that, by emulating the S3 server with [Nock][nock] and [Dicer][dicer].

- Cleaner code. The old code was obscure and had unreachable code because many people had contributed to `z.stashFile()`. Now the new code is more modern and cleaner.

[nock]: https://github.com/nock/nock
[dicer]: https://github.com/mscdex/dicer


eliangcs:
@xavdid I was struggling getting Travis to pass, so I went ahead and replaced Travis with GitHub Actions. We can ignore the failures on Windows for now because they'd have failed in the first place.

Sorry if this PR is getting too big! Obviously, I'm going to do the same PR on latest 11.x.

eliangcs:
@xavdid I was struggling with Travis when working on this, so I went ahead and replace it with GitHub Actions in this PR (and the sibling 11.x [PR](https://github.com/zapier/zapier-platform/pull/454)). We can ignore those failed Windows tests for now because they would have failed in the first place.

Sorry if this PR is getting too big! cc @benzapier as I used your the config template in https://github.com/zapier/zapier-platform/pull/437 as a reference.

xavdid (approved):
nice! very awesome improvement.

xavdid (approved):

PR #455, opened 2021-11-29
Provide allow human label key attributes to include space and brackets in RefResourceSchema.

bryanhelmig:
This shouldn't be a breaking change, but we will want to backport to `9.x.x` series to complete the WB.

@xavdid @eliangcs would appreciate your 👀 here and guidance on how to ship this to the `11.x.x` series and also backport to `9.x.x` (and update the monolith to use it as the default for conversions?).

This should resolve many of the outstanding schema issues as defined in the [internal thread](https://zapier.slack.com/archives/C5Z9BP4U9/p1638214620358600?thread_ts=1638198315.336500&cid=C5Z9BP4U9) and [internal Coda doc](https://coda.io/d/di0MgBhlCWf/suzvT#_lufiU) around the usage of values like `lead_added.id.data[]id` or `lead_added.id.First Name` which should be okay according to https://zapier.github.io/zapier-platform-schema/build/schema.html#fieldschema's `key` type of broad `string`.


bryanhelmig:
@eliangcs I see you have a couple backports happening like https://github.com/zapier/zapier-platform/pull/453, I'm happy to replicate your work, or if you can backport it simply enough, I'll happily 👍 the PRs!

packages/schema/.nvmrc
```diff
@@ -1 +1 @@
-v8.10.0
+v10
```
xavdid, line 1: not a blocker, but this should probably be `14` at this point

packages/schema/lib/schemas/RefResourceSchema.js
```diff
@@ -7,18 +7,29 @@ module.exports = makeSchema({
   description:
     'Reference a resource by key and the data it returns. In the format of: `{resource_key}.{foreign_key}(.{human_label_key})`.',
```
xavdid, line 8: Could also expand these docs to include some of the fancier references (specifically the `[]`). But, only if we want people to use them. If it's just for compatibility, then we can probably ignore.

xavdid (approved):
Looks good!

While reading, I had one other thought. Why did we make the reference regex more restrictive in CLI? Do we want to try and keep the regex how CLI has always had it, or encourage these types of references for all apps?

If we want people to be able to use these expanded references, then we can merge as is. If we don't, we could add a new schema, `LegacyRefResource`, which has this regex and leave the existing one. Then, we could write a functional constraint to allow legacy apps to use legacy references but restrict native CLI apps to the more "modern" approach. As long as they both work in the editor/monolith, then it's just a stylistic choice for us.

bryanhelmig:
> Why did we make the reference regex more restrictive in CLI?

@xavdid I genuinely think it was just an omission brought about by just duplicating the `resource_key` portion of the regex with period `.` separators.

> Do we want to try and keep the regex how CLI has always had it, or encourage these types of references for all apps?

I think if we want to be more restrictive, we should make sure all the `FieldSchema` `key`s match the regex, right now its very open ended (any string). But that would be majorly backwards incompatible!

----

If we break down: `{resource_key}.{foreign_key}(.{human_label_key})`, we should have this:

* `resource_key` -- should really be `KeySchema` (`[a-zA-Z]+[a-zA-Z0-9_]*`) instead of `[a-zA-Z0-9_]+`. Is minorly backwards incompatible, and really only bans a leading `0-9` or `_`. This PR does not address this.
* `foreign_key` -- should really match `FieldSchema` `key`, which basically is a `.*` but without any literal periods `.` or commas `,` (which would bust most of our basic parsing). This PR attempts to partially address this issue, but only for the blocker ` ` and `[]` chars to minimize complexity of the change.
* `human_label_key` -- same as ☝️, but with awareness of `,`. Likewise, this PR attempts to partially address this issue.

eliangcs (approved):

PR #459, opened 2021-12-01
PDE-2085 chore: bump node-fetch to 2.6.6

eliangcs:
Same as #458 but for 11.x.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -59,7 +59,7 @@
     "lodash": "4.17.21",
     "marked": "2.0.3",
     "marked-terminal": "4.1.1",
-    "node-fetch": "2.6.1",
+    "node-fetch": "2.6.6",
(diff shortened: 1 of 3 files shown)


xavdid (approved):
cool. Is there a reason for the bump, or just staying current? Do we anticipate any issues in legacy-script runner on the 9.x branch?

eliangcs:
@xavdid thanks for asking! I wrote the reason in the description of https://github.com/zapier/zapier-platform/pull/458.

xavdid:
wow those checks were so fast 😅

eliangcs:
> wow those checks were so fast 

Exactly! We should've moved away from Travis earlier. I don't know why, but the latency between GitHub Actions and our test servers (https://httpbin.zapier-tooling.com and https://auth-json-server.zapier-staging.com) is super short!

PR #461, opened 2021-12-02
PDE-2826 fix(core): backpressure issue with node-fetch patch

eliangcs:
Using node-fetch 2.6.6 (Node.js 14), this program hangs on my laptop:

```js
const fs = require('fs');
const fetch = require('node-fetch');

async function main() {
  // 36864 bytes = 16 KB
  // If it's 36863, the bug won't happen
  const downloadUrl = 'https://httpbin.org/stream-bytes/36864';
  const downloadResponse = await fetch(downloadUrl);

  const uploadUrl = 'https://httpbin.org/put';
  const uploadOpts = {
    method: 'PUT',
    body: downloadResponse.body,
  };

  const uploadRequest = new fetch.Request(uploadUrl, uploadOpts);
  const uploadPromise = fetch(uploadRequest);
  const uploadResponse = await uploadPromise;
  console.log(uploadResponse.status);
}

main();
```

When you pass a `fetch.Request` object to `fetch()`, `fetch()` [copies](https://github.com/node-fetch/node-fetch/blob/v2.6.6/src/index.js#L46) the `Request` object. And if the `Request` has a streamable body, the `Request` constructor [clones](https://github.com/node-fetch/node-fetch/blob/v2.6.6/src/request.js#L106) the body by [teeing](https://github.com/node-fetch/node-fetch/blob/v2.6.6/src/body.js#L403-L412) the stream into two [`PassThrough`](https://nodejs.org/docs/latest-v14.x/api/stream.html#stream_class_stream_passthrough) streams. Then we read one of the `PassThrough` stream (A) and not the other (B). This causes B's internal buffer to fill up, so Node.js has to pause reading from the source until someone consumes B's buffer.

To fix, we add that someone to consume B. Right after the `fetch(uploadRequest)` call, we pipe `uploadRequest.body` to `/dev/null`:

```js
  const uploadPromise = fetch(uploadRequest);

  uploadRequest.body.pipe(fs.createWriteStream('/dev/null'));  // <---

  const uploadResponse = await uploadPromise;
```

And that solves the hanging problem.

This PR does something similar with our node-fetch patch.

### DIFF
--- a/packages/core/src/tools/create-file-stasher.js
+++ b/packages/core/src/tools/create-file-stasher.js
@@ -60,7 +60,11 @@ const resolveRemoteStream = async (stream) => {
   try {
     await streamPipeline(stream, fs.createWriteStream(tmpFilePath));
   } catch (error) {
-    fs.unlinkSync(tmpFilePath);
+    try {
+      fs.unlinkSync(tmpFilePath);
+    } catch (e) {
+      // File doesn't exist? Probably okay
+    }
     throw error;
   }
 
@@ -69,7 +73,11 @@ const resolveRemoteStream = async (stream) => {
 
   readStream.on('end', () => {
     // Burn after reading
-    fs.unlinkSync(tmpFilePath);
+    try {
+      fs.unlinkSync(tmpFilePath);
+    } catch (e) {
+      // TODO: We probably want to log warning here
+    }
   });
 
   return {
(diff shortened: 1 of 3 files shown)


xavdid (dismissed):
wowza! Nice one

xavdid (approved):

PR #469, opened 2022-01-12
PDE-2650 perf(core): improve logging performance

eliangcs:
Improves `z.console.log` and HTTP logging performance. In this PR, we stream many logs in a single request to the `/input` endpoint on the log server. This reduces the overhead where the original `/input` endpoint only took one log in a request.

I used the following trigger code for benchmark:

```js
  const response = await z.request('https://httpbin.zapier-tooling.com/get');
  const requestId = response.data.headers['X-Request-Id'][0];

  for (let i = 0; i < 200; i++) {
    z.console.log(`hello ${i}`);
  }

  return [{ id: requestId }];
};
```

The above example code produces one HTTP log and 200 console logs. The old way (one log, one request) takes ~6.6 seconds to invoke the trigger. The new way (stream many logs in a request) takes ~3.5 seconds. I was expecting a bigger improvement but I think this is as good as it gets.

### DIFF
--- a/packages/cli/src/smoke-tests/smoke-tests.js
+++ b/packages/cli/src/smoke-tests/smoke-tests.js
@@ -1,6 +1,7 @@
 const fs = require('fs-extra');
 const os = require('os');
 const path = require('path');
+const { promisify } = require('util');
 
 require('should');
 
@@ -10,6 +11,8 @@ const { runCommand, npmPackCore } = require('../tests/_helpers');
 const { PLATFORM_PACKAGE } = require('../constants');
 
 const REGEX_VERSION = /\d+\.\d+\.\d+/;
+
+const sleep = promisify(setTimeout);
 
 const setupZapierRC = () => {
   let hasRC = false;
@@ -65,7 +68,9 @@ const yarnInstall = (coreZipPath, workdir) => {
   }
 };
 
-describe('smoke tests - setup will take some time', () => {
+describe('smoke tests - setup will take some time', function () {
+  this.retries(3);
+
   const context = {
     // Global context that will be available for all test cases in this test suite
     package: {
@@ -100,6 +105,21 @@ describe('smoke tests - setup will take some time', () => {
   after(() => {
     fs.unlinkSync(context.package.path);
     fs.removeSync(context.workdir);
+  });
+
+  beforeEach(async function () {
+    // delay after first failure
+    const currentRetry = this.currentTest.currentRetry();
+    if (currentRetry > 0) {
+      await sleep(Math.pow(2, currentRetry) * 1000);
+    }
+  });
+
+  afterEach(function () {
+    const currentRetry = this.currentTest.currentRetry();
+    if (currentRetry >= 0) {
+      this.currentTest.title += ` (${currentRetry} retries)`;
+    }
   });
 
   it('package size should not change much', async function () {
--- a/packages/core/src/tools/create-lambda-handler.js
+++ b/packages/core/src/tools/create-lambda-handler.js
@@ -147,38 +147,36 @@ const createLambdaHandler = (appRawOrPath) => {
 
     environmentTools.cleanEnvironment();
 
+    // Copy bundle environment into process.env *before* creating the logger and
+    // loading app code, so that the logger gets the endpoint from process.env,
+    // and top level app code can get bundle environment vars via process.env.
+    environmentTools.applyEnvironment(event);
+
     // Create logger outside of domain, so we can use in both error and run callbacks.
     const logBuffer = [];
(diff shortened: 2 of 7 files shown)


packages/core/src/tools/create-lambda-handler.js
```diff
@@ -203,10 +204,10 @@ const createLambdaHandler = (appRawOrPath) => {
 
           const { skipHttpPatch } = appRaw.flags || {};
           // Adds logging for _all_ kinds of http(s) requests, no matter the library
-          if (!skipHttpPatch) {
+          if (!skipHttpPatch && !event.calledFromCli) {
```
eliangcs, line 207: I was running `zapier push` and found the CLI tool sent HTTP logs to the log server. So I added `!event.calledFromCli` to make sure we don't patch `http` when it was from CLI.

packages/core/src/tools/create-lambda-handler.js
```diff
@@ -203,10 +204,10 @@ const createLambdaHandler = (appRawOrPath) => {
...
             const httpPatch = createHttpPatch(event);
-            httpPatch(require('http'));
-            httpPatch(require('https')); // 'https' needs to be patched separately
+            httpPatch(require('http'), logger);
+            httpPatch(require('https'), logger); // 'https' needs to be patched separately
```
eliangcs, line 210: This ensures HTTP logs and console logs share the same `logger` instance created by `createLogger()`. In the old implementation, there's no difference because every log is a separate request. But in the new implementation, every time you call `createLogger()`, you create a new request, ready to stream logs. Here we want to reuse the same request as much as possible.

eliangcs:
@xavdid I found there's an issue with lost logs, which I'm still figuring out. So you don't have to review this PR just yet!

packages/core/src/tools/create-lambda-handler.js
```diff
@@ -147,37 +147,36 @@ const createLambdaHandler = (appRawOrPath) => {
 
     environmentTools.cleanEnvironment();
 
+    // Copy bundle environment into process.env *before* creating the logger and
+    // loading app code, so that the logger gets the endpoint from process.env,
+    // and top level app code can get bundle environment vars via process.env.
+    environmentTools.applyEnvironment(event);
```
eliangcs, line 153: I moved `applyEnvironment` up because I found on the logging endpoint could be different on the first invocation than the subsequent invocations.

[Here](https://github.com/zapier/zapier-platform/blob/5d6b06d39a149c600be6e7f838a0855c257f76ac/packages/core/src/tools/create-logger.js#L199) you can see the logging endpoint is either derived from the environment variable `LOGGING_ENDPOINT` or the constant `DEFAULT_LOGGING_HTTP_ENDPOINT`. But in the original (bad) code, we [`applyEnvironment`](https://github.com/zapier/zapier-platform/blob/5d6b06d39a149c600be6e7f838a0855c257f76ac/packages/core/src/tools/create-lambda-handler.js#L196) AFTER [`createLogger`](https://github.com/zapier/zapier-platform/blob/5d6b06d39a149c600be6e7f838a0855c257f76ac/packages/core/src/tools/create-lambda-handler.js#L152). So on the first invocation, it would get the logging endpoint from the constant. But on the second or subsequent calls, since the environment variables have been loaded, it gets the logging endpoint from the environment variable.

We don't see any issues on production because the constant and the environment variables are the same (`https://httplogger.zapier.com/input`) on production.

To fix, we need to make sure we `applyEnvironment` BEFORE `createLogger`.

eliangcs:
@xavdid this is ready for review!

packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -102,6 +107,21 @@ describe('smoke tests - setup will take some time', () => {
     fs.removeSync(context.workdir);
   });
 
+  beforeEach(async function () {
+    // delay after first failure
+    const currentRetry = this.currentTest.currentRetry();
+    if (currentRetry > 0) {
+      await sleep(Math.pow(2, currentRetry) * 1000);
```
eliangcs, line 114: Hopefully this fixes the test timeout issue.

packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -65,7 +68,9 @@ const yarnInstall = (coreZipPath, workdir) => {
   }
 };
 
-describe('smoke tests - setup will take some time', () => {
+describe('smoke tests - setup will take some time', function () {
+  this.retries(3);
```
eliangcs, line 72: Hopefully this fixes the test timeout issue.

xavdid (approved):
Very awesome! Looks great here.

PR #475, opened 2022-01-21
fix(core): handle case where scriptingSource is an empty string

xavdid:
Handle a rare case where `legacy.scriptingSource` is an empty string on a legacy app, so we don't initialize LSR. Then, `z.legacyScripting.run(...)` fails w/ `Cannot read property 'run' of undefined`.

### DIFF
--- a/packages/core/ARCHITECTURE.md
+++ b/packages/core/ARCHITECTURE.md
@@ -31,7 +31,7 @@
 
 - At runtime, a developer's `perform` function is invoked with the following signature: `perform(z, bundle)`.
 - The `z` object is a collection of functions used commonly by developers and is defined in `core/src/tools/create-lambda-handler.js`
-- Its functions are either Zapier-specific (such as `z.cursor` and `z.dehyrate`) or wrappers around common JS functionality (such as `z.JSON.parse` and `z.console.log`) with better error handling or extra logging
+- Its functions are either Zapier-specific (such as `z.cursor` and `z.dehydrate`) or wrappers around common JS functionality (such as `z.JSON.parse` and `z.console.log`) with better error handling or extra logging
 - Each of those functions has its own file in `core/src/tools`
 - The most important method is probably `z.request` (from `core/src/tools/create-app-request-client.js`), which devs use to make external requests. This is different from normal http requests in a number of ways:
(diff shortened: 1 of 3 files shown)


packages/core/ARCHITECTURE.md
```diff
@@ -31,7 +31,7 @@
 
 - At runtime, a developer's `perform` function is invoked with the following signature: `perform(z, bundle)`.
 - The `z` object is a collection of functions used commonly by developers and is defined in `core/src/tools/create-lambda-handler.js`
-- Its functions are either Zapier-specific (such as `z.cursor` and `z.dehyrate`) or wrappers around common JS functionality (such as `z.JSON.parse` and `z.console.log`) with better error handling or extra logging
+- Its functions are either Zapier-specific (such as `z.cursor` and `z.dehydrate`) or wrappers around common JS functionality (such as `z.JSON.parse` and `z.console.log`) with better error handling or extra logging
```
xavdid, line 34: unrelated typo

packages/core/src/tools/create-legacy-scripting-runner.js
```diff
@@ -8,8 +8,11 @@ const semver = require('semver');
 const createLegacyScriptingRunner = (z, input) => {
   const app = _.get(input, '_zapier.app');
 
-  let source =
-    _.get(app, 'legacy.scriptingSource') || app.legacyScriptingSource;
+  // once we have node 14 everywhere, this can be:
```
xavdid, line 11: Node 12 is EOL may 2022, so not too long!

xavdid:
re: failing tests, I was looking at them today and they all passed normally on node 12 & 14 on my local. Maybe GH actions is slower and needs a bigger timeout? Each failing run, it's been a different combination of test & node version failing, so maybe we're just getting unlucky. None of the other tests take near 2 min, but it's possible there's a lot of variance there?

eliangcs (approved):
Good catch! We probably need to backport this to 9.x. To do that, make a new branch based off `release-9.x`, make the changes, and create a PR merging to `release-9.x`.

eliangcs:
Regarding the test timeout, I guess the jobs are more likely to timeout when they are run in parallel on GitHub Actions. And I agree, we could try a bigger time limit first.

PR #480, opened 2022-02-01
PDE-2981: feature(cli) - Update "promote" and "migrate" to use new endpoint

standielpls:
Internally points the CLI `promote` and `migrate` command to use the new [migration endpoint](https://github.com/zapier/zapier/blob/8379df093af726b2318083282889cd73eda4c81c/web/backend/zapier/app_platform/developer_cli/api/urls.py). 

These commands currently point to an endpoint that calls out to the new migration endpoint so this is work to remove that indirection.

### DIFF
--- a/packages/cli/src/oclif/commands/migrate.js
+++ b/packages/cli/src/oclif/commands/migrate.js
@@ -15,6 +15,7 @@ class MigrateCommand extends BaseCommand {
     const user = this.flags.user;
     const fromVersion = this.args.fromVersion;
     const toVersion = this.args.toVersion;
+    const email = this.args.email;
 
     if (user && percent !== 100) {
       this.error(
@@ -46,7 +47,9 @@ class MigrateCommand extends BaseCommand {
     }
 
     const body = {
-      percent,
+      name: 'migrate',
+      from_version: fromVersion,
+      to_version: toVersion,
     };
     if (user) {
       body.user = user;
@@ -58,8 +61,16 @@ class MigrateCommand extends BaseCommand {
         `Starting migration from ${fromVersion} to ${toVersion} for ${percent}%`
       );
     }
+    if (percent) {
+      body.percent_human = percent;
+    }
+    if (email) {
+      body.email = email;
+    }
 
-    const url = `/apps/${app.id}/versions/${fromVersion}/migrate-to/${toVersion}`;
+    // const url = `/apps/${app.id}/versions/${fromVersion}/migrate-to/${toVersion}`;
+    const url = `/apps/${app.id}/migrations`;
+
     try {
       await callAPI(url, { method: 'POST', body });
     } finally {
--- a/packages/cli/src/oclif/commands/migrate.js
+++ b/packages/cli/src/oclif/commands/migrate.js
@@ -47,12 +47,14 @@ class MigrateCommand extends BaseCommand {
     }
 
     const body = {
-      name: 'migrate',
-      from_version: fromVersion,
-      to_version: toVersion,
+      job: {
+        name: 'migrate',
+        from_version: fromVersion,
+        to_version: toVersion,
+      },
     };
     if (user) {
-      body.user = user;
+      body.job.user = user;
       this.startSpinner(
         `Starting migration from ${fromVersion} to ${toVersion} for ${user}`
       );
@@ -62,13 +64,12 @@ class MigrateCommand extends BaseCommand {
       );
     }
     if (percent) {
-      body.percent_human = percent;
+      body.job.percent_human = percent;
     }
     if (email) {
-      body.email = email;
+      body.job.email = email;
     }
 
-    // const url = `/apps/${app.id}/versions/${fromVersion}/migrate-to/${toVersion}`;
     const url = `/apps/${app.id}/migrations`;
 
     try {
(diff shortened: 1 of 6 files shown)


packages/cli/src/oclif/commands/migrate.js
```diff
@@ -46,7 +47,9 @@ class MigrateCommand extends BaseCommand {
     }
 
     const body = {
-      percent,
+      name: 'migrate',
+      from_version: fromVersion,
+      to_version: toVersion,
```
standielpls, line 52: Is this the right place to review map from? 
https://github.com/zapier/zapier/blob/16c0ad7ce5dbb39f9802bd8b45c2305c35b77699/web/backend/zapier/app_platform/developer_cli/api/migration_resources.py

standielpls:
@eliangcs - not sure what testing this looks like but I can supply a screenshot of this working locally when pointing to zapier-staging.com

https://cdn.zappy.app/aa83a9fda53ee99209fd85dc9749cc49.png

packages/cli/src/oclif/commands/promote.js
```diff
@@ -28,6 +28,7 @@ class PromoteCommand extends BaseCommand {
     const app = await this.getWritableApp();
 
     const version = this.args.version;
+    console.log('hey buddy!');
```
xavdid, line 31: probably don't want to ship this 😁

packages/cli/src/oclif/commands/promote.js
```diff
@@ -60,19 +61,24 @@ class PromoteCommand extends BaseCommand {
...
-    const body = {};
+    const body = {
+      job: {
+        name: 'promote',
+        to_version: version,
+      },
+    };
     if (changelog) {
-      body.changelog = changelog;
+      body.job.changelog = changelog;
```
xavdid, line 71: I know you were just following existing code, but you can move this above for the same effect:

```js
const body = {
      job: {
        name: 'promote',
        to_version: version,
        changelog,
      },
    };
```

Since it'll be undefined if it's missing, it just works

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -15,6 +15,7 @@ class MigrateCommand extends BaseCommand {
     const user = this.flags.user;
     const fromVersion = this.args.fromVersion;
     const toVersion = this.args.toVersion;
+    const email = this.args.email;
```
xavdid, line 18: This won't quite work. If you look below at `MigrateCommand.flags`, we only define the `user` flag. I think if you provide `--email`, the CLI will error at you.

Because the name `email` vs `user` doesn't matter, I'd keep the existing flag name and just save it as `email` in the request body. That way the flags don't change for our users, but we send the right data to the API. Does that make sense?

packages/cli/src/oclif/commands/promote.js
```diff
@@ -120,7 +126,7 @@ PromoteCommand.args = [
   {
     name: 'version',
     required: true,
-    description: 'The version you want promote.',
+    description: 'The version you want to promote.',
```
xavdid, line 129: Nice catch! 🙌

packages/cli/src/oclif/commands/migrate.js
xavdid: yep!

xavdid (commented):
This is a great start! There are some things we'll want to fix before we ship, but we're most of the way there.

re: testing- we don't have great unit test coverage on CLI now, so as long as you've done thorough testing against staging/local, we're probably fine. Make sure to test all the args/flags! 😁

packages/cli/src/oclif/commands/migrate.js
standielpls: Yep, makes sense, good catch!

> Because the name email vs user doesn't matter

Can you help me locate this for my own learning? I'm tracing where the endpoint builds the payload and find it set [here](https://github.com/zapier/zapier/blob/986861aa44dbbed208189c6f5b630e78828b4d9a/web/backend/zapier/migration_fsm/statedb.py#L63) but not sure where it will end up being used.

standielpls: https://cdn.zappy.app/eb7c153daa90542ec50c964bdc903dcb.png

nvm! I see now that user _is_ email in this system!

standielpls:
@xavdid Good notes, I've tested with the available flags and the commands are working

I've found some [logs](https://logs.internal.zapier.com/search?q=url%3Ahttps%5C%3A%5C%2F%5C%2Fzapier-staging.com%5C%2Fapi%5C%2Fplatform%5C%2Fcli%5C%2Fapps%5C%2F119258%2A&rangetype=relative&relative=432000) that indicate the requests went through, but I don't see a "request body" in them. Perhaps this field isn't logged?

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -58,8 +62,12 @@ class MigrateCommand extends BaseCommand {
         `Starting migration from ${fromVersion} to ${toVersion} for ${percent}%`
       );
     }
+    if (percent) {
+      body.job.percent_human = percent;
```
xavdid, line 66: I believe this can also be included rigtht in the `job` above, since it's always defined (either from the default or from the user passing it explicitly)

xavdid (approved):
Looks great! had one minor suggestion, but you can also go ahead and merge.

Also- the title of this PR is `chore(cli)`, but that's usually reserved for bumping deps, fixing typos, etc. This is a feature! New and better migrations!

eliangcs (approved):
I tested all the code branches locally, and they all work wonderfully! Nice work!

PR #482, opened 2022-02-03
[SUPPORTESC-297] docs(cli): Clarification on OAuth2 refresh details

armchairlinguist:
Add some additional context and links around OAuth2 refresh options, and some other auth cleanup for consistency/clarity.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -307,11 +307,11 @@ Most applications require some sort of authentication - and Zapier provides a ha
 
 ### Basic
 
-Useful if your app requires two pieces of information to authentication: `username` and `password` which only the end user can provide. By default, Zapier will do the standard Basic authentication base64 header encoding for you (via an automatically registered middleware).
+Useful if your app requires two pieces of information to authenticate: `username` and `password`, which only the end user can provide. By default, Zapier will do the standard Basic authentication base64 header encoding for you (via an automatically registered middleware).
 
 > Example App: Check out https://github.com/zapier/zapier-platform/tree/master/example-apps/basic-auth for a working example app for basic auth.
 
-> Note: If you do the common API Key pattern like `Authorization: Basic APIKEYHERE:x` you should look at the "Custom" authentication method instead.
+If your app uses Basic auth with an encoded API key rather than a username and password, like `Authorization: Basic APIKEYHERE:x`, consider the [Custom](#custom) authentication method instead.
 
 ```js
 [insert-file:./snippets/basic-auth.js]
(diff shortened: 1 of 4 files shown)


packages/cli/README-source.md
```diff
@@ -363,11 +363,11 @@ Zapier's OAuth1 implementation matches [Twitter's](https://developer.twitter.com
...
 The flow works like this:
 
-  1. Zapier makes a call to your API requesting a "request token" (also known as "temporary credentials")
-  2. Zapier sends the user to the authorization URL, defined by your app, along with the request token
+  1. Zapier makes a call to your API requesting a "request token" (also known as "temporary credentials").
+  2. Zapier sends the user to the authorization URL, defined by your app, along with the request token.
   3. Once authorized, your website sends the user to the `redirect_uri` Zapier provided. Use `zapier describe` command to find out what it is: ![](https://zappy.zapier.com/117ECB35-5CCA-4C98-B74A-35F1AD9A3337.png)
-  4. Zapier makes a call on our backend to your API to exchange the request token for an "access token" (also known as "long-lived credentials")
-  5. Zapier remembers the access token and makes calls on behalf of the user
+  4. Zapier makes a backend call to your API to your API to exchange the request token for an "access token" (also known as "long-lived credentials").
```
eliangcs, line 369: Looks like "to your API" is duplicated.

```suggestion
  4. Zapier makes a backend call to your API to exchange the request token for an "access token" (also known as "long-lived credentials").
```

packages/cli/README-source.md
```diff
@@ -1415,15 +1424,18 @@ Example: `throw new z.errors.HaltedError('Your reason.');`
 
 For apps that require manual refresh of authorization on a regular basis, Zapier
 provides a mechanism to notify users of expired credentials. With the
-`ExpiredAuthError`, the current operation is interrupted, the Zap is turned off
-(to prevent more calls with expired credentials), and a predefined email is sent
-out informing the user to refresh the credentials.
+`ExpiredAuthError`, the current operation is interrupted and a predefined email
+is sent out asking the user to refresh the credentials. While the auth is
+disconnected, Zap runs will not be executed, to prevent more calls with expired
```
eliangcs, line 1429: I think it's still necessary to mention the Zap is turned off when the auth is disconnected:

```suggestion
disconnected, the Zap will be turned off and Zap runs will not be executed, to prevent more calls with expired
```

eliangcs (dismissed):
Thanks for improving the docs. I have a couple of suggestions. Happy to re-approve if there are any further changes.

packages/cli/README-source.md
armchairlinguist: @eliangcs The reason I changed that is that it's not correct. The Zap is not turned off. 😄 Instead, it remains on and potential runs are held, as described.

eliangcs: @armchairlinguist glad you tested it! No need to change then 👍

eliangcs (approved):

PR #488, opened 2022-02-15
[SUPPORTESC-307] docs(cli): Clarify auth section wording, especially authData vs inputData

armchairlinguist:
The auth section needed some love, including clarification, additional cross-linking, and updated external links.

The main focus of the update is the slightly confusing wording about authData vs inputData. I've tried to re-write this so that it's still clear which is used where. The original phrasing seemed confusing to me, though, as it wasn't always about data that was pre-existing, or data input by the user. I've gone toward describing the distinction as long-term vs temporary use, as that's the pattern I saw as I worked.

It's possible I've gotten some details slightly wonked up in the process, or even that that distinction isn't quite right either, so please do correct me if anything is now/remains mistaken.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -436,7 +436,7 @@ Your auth definition would look something like this:
 [insert-file:./snippets/oauth2.js]
 ```
 
-For OAuth2, `authentication.oauth2Config.authorizeUrl`, `authentication.oauth2Config.getAccessToken`, and `authentication.oauth2Config.refreshAccessToken`  have fields like `redirect_uri` and `state` in `bundle.inputData`, instead of `bundle.authData`. After the code is exchanged for an access token and/or refresh token, those tokens are stored in `bundle.authData` for long-term use.
+For OAuth2, `authentication.oauth2Config.authorizeUrl`, `authentication.oauth2Config.getAccessToken`, and `authentication.oauth2Config.refreshAccessToken` have fields like `redirect_uri` and `state` in `bundle.inputData`, instead of `bundle.authData`. After the code is exchanged for an access token and/or refresh token, those tokens are stored in `bundle.authData` for long-term use.
 
 Also note that `authentication.oauth2Config.getAccessToken` has access to the additional return values in `rawRequest` and `cleanedRequest` should you need to extract other values (for example, from the query string).
 
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -301,9 +301,9 @@ zapier convert 1234 --version 1.0.1 my-app
 
 ## Authentication
 
-Most applications require some sort of authentication - and Zapier provides a handful of methods for helping your users authenticate with your application. Zapier will provide some of the core behaviors, but you'll likely need to handle the rest.
+Most applications require some sort of authentication. The Zapier platform provides core behaviors for several authentication methods that might be used with your application, as well as the ability to configure or customize authentication further for certain methods.
 
-> Hint: You can access the data tied to your authentication via the `bundle.authData` property in any method called in your app. Exceptions exist in OAuth and Session auth. Please see them below.
+Data tied to authentication is included in the [bundle object](#bundle-object). Data used on an ongoing basis to authenticate requests is included in `bundle.authData`, while temporary values used in the establishment of [OAuth](#oauth1) and [Session auth](#session) are stored in `bundle.inputData`.
 
 ### Basic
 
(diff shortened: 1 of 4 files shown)


packages/cli/README-source.md
```diff
@@ -301,9 +301,9 @@ zapier convert 1234 --version 1.0.1 my-app
 
 ## Authentication
 
-Most applications require some sort of authentication - and Zapier provides a handful of methods for helping your users authenticate with your application. Zapier will provide some of the core behaviors, but you'll likely need to handle the rest.
+Most applications require some sort of authentication. The Zapier platform provides core behaviors for several authentication methods that might be used with your application, as well as the ability to configure or customize authentication further for certain methods.
```
xavdid, line 304: might change `several authentication methods` to `several common authentication methods`- we do the popular ones and if you do them right, then there's usually no further configuration needed.

xavdid: also re:

> customize authentication further for certain methods.

can't you customize all methods?

packages/cli/README-source.md
```diff
@@ -301,9 +301,9 @@ zapier convert 1234 --version 1.0.1 my-app
...
 
-> Hint: You can access the data tied to your authentication via the `bundle.authData` property in any method called in your app. Exceptions exist in OAuth and Session auth. Please see them below.
+Data tied to authentication is included in the [bundle object](#bundle-object). Data used on an ongoing basis to authenticate requests is included in `bundle.authData`, while temporary values used in the establishment of [OAuth](#oauth1) and [Session auth](#session) are stored in `bundle.inputData`.
```
xavdid, line 306: This still reads a little confusing. During `perform`, all data returned from auth (including when you use oauth1 and session) is in `authData`. 

The only time it's in `inputData` instead is during auth operations. So, for an oauth refresh, the previous auth data is in `inputData`. 

I think the regular `perform` vs `auth operation` (refresh, etc) distinction is important here, so maybe lean on that

packages/cli/README-source.md
```diff
@@ -321,19 +321,19 @@ If your app uses Basic auth with an encoded API key rather than a username and p
...
+> Limitation: Currently, MD5-sess and SHA are not implemented. Only the MD5 algorithm is supported. In addition, server nonces are not reused. That means for every `z.request` call, Zapier will send an additional request beforehand to get the server nonce.
 
 ```js
 [insert-file:./snippets/digest-auth.js]
 ```
 
 ### Custom
 
-This is what most "API Key" driven apps should default to using. You'll likely provide some custom `beforeRequest` middleware or a `requestTemplate` to complete the authentication by adding/computing needed headers.
+Custom auth is most commonly used for apps that authenticate with API keys, although it also provides flexibility for any unusual authentication setup. You'll likely provide some custom `beforeRequest` middleware or a `requestTemplate` (see [Making HTTP Requests](#making-http-requests)) to pass in data returned from the authentication process, most commonly by adding/computing needed headers.
```
xavdid, line 336: This is a great improvement!

packages/cli/README-source.md
```diff
@@ -343,21 +343,21 @@ This is what most "API Key" driven apps should default to using. You'll likely p
...
+Session auth gives you the ability to exchange some user provided data for some authentication data; for example, username and password for a session key. It can be used to implement almost any authentication method that uses that pattern - for example, alternative OAuth flows.
 
 > Example App: Check out https://github.com/zapier/zapier-platform/tree/master/example-apps/session-auth for a working example app for session auth.
 
 ```js
 [insert-file:./snippets/session-auth.js]
 ```
 
-> Note: For Session auth, `authentication.sessionConfig.perform` will have the provided fields in `bundle.inputData` instead of `bundle.authData` because `bundle.authData` will only have "previously existing" values, which will be empty the first time the Zap runs.
+For Session auth, the function that fetches the additional authentication data needed to make API calls ( `authentication.sessionConfig.perform`) will have the user-provided fields in `bundle.inputData` instead of `bundle.authData`. Afterwards, `bundle.authData` will contain the data provided by that function (usually the session key or token).
```
xavdid, line 354: > contain the data provided by that function

I'd change to 

> contain the data returned from that function

xavdid (commented):
Awesome! Great improvements here.

Had a couple of minor suggestions, but nothing too tough.

packages/cli/README-source.md
armchairlinguist: @xavdid Thanks for the great comments here. Now that I think of it, all methods are customizable to an extent - we've even added additional fields for Basic (although that's not mentioned in this text), plus the test request is always custom. No problem to remove that last bit.

packages/cli/README-source.md
armchairlinguist: Thanks for the analysis on this! The point about the existing authData going into inputData for OAuth refresh does mean that my framing isn't quite there. I guess I think of `perform` as also being an "auth operation", so I'm still not sure if that will improve clarity overall. 🤔 I'll give this some more thought.

xavdid:
@armchairlinguist is this work still relevant?

armchairlinguist:
That's a question I was asking myself as well! 
What I might do here is close this PR, and create a different branch with just the straightforward changes, and try to tackle authData vs inputData later.

armchairlinguist:
Superseded by https://github.com/zapier/zapier-platform/pull/527

PR #497, opened 2022-02-24
feat(core,schema): enable skipping encoding specific characters from query params

xavdid:
Proof of concept to enable apps (especially legacy apps) to skip encoding certain params. 

Once thing I noticed (but didn't have time to dig into) was that if the url already includes params, we pull them out before the request object reaches `addQueryParams`:

```js
z.request('https://httpbin.org/get?in_url=asdf$$asdf')

// in `addQueryParams`
console.log(req)

{
  method: 'GET',
  params: { in_url: 'asdf$$asdf' },
  url: 'https://httpbin.org/get',
  // ...
}

// in request-client.js, directly before the call to `fetch(url, options)`
console.log('calling', url)
// calling https://httpbin.org/get?in_url=asdf%24%24asdf
```

But, as written, this will at least partially unblock us

### DIFF
--- a/packages/core/src/http-middlewares/before/add-query-params.js
+++ b/packages/core/src/http-middlewares/before/add-query-params.js
@@ -16,21 +16,22 @@ const addQueryParams = (req) => {
 
     let stringifiedParams = querystring.stringify(req.params);
 
-    // it goes against spec, but for compatibility, some APIs want certain characters (mostly $) unencoded
-    (req.skipEncodingChars || []).forEach((char) => {
-      if (char.length > 1) {
-        throw new Error('only pass single characters to skipEncodingChars');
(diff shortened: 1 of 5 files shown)


eliangcs:
@xavdid I made some changes:

- changed the type of `skipEncodingChars` from array to string. Because with string, we don't have to validate if each element is a single character.
- added it to schema in `RequestSchema`
- I think this is more like a feature as opposed to a bug fix, so I changed the PR title from `fix(core)` to `feat(core,schema)`

I'm approving. Let me know if there's anything else you want to change. Thanks!

eliangcs (approved):

xavdid (commented):
I can't approve my own PR, but your changes look great! 👍

PR #498, opened 2022-02-25
chore: allow to build boilerplate from local or npm

eliangcs:
Allows to build boilerplate with one of these combinations of source:

| core  | legacy-scripting-runner | example command      |
|-------|-------------------------|----------------------|
| local | local                   | `build.sh`             |
| npm   | local                   | `build.sh 9.7.1`       |
| local | npm                     | `build.sh '' 3.8.5`    |
| npm   | npm                     | `build.sh 9.7.1 3.8.5` |

PR #505, opened 2022-02-28
fix(cli): bump marked-terminal to fix peer dep issue

xavdid:
Fixes #498 by updating `marked-terminal` to a later version, which includes `marked@4` as a peer dep.


xavdid:
Ah, looks like `marked-terminal@5` requires Node.js 14, but we're still supporting node 12 until the end of april (2 more months), so we can't merge this for now.

Given that the CLI still installs and works successfully, i'll close this for now and set a reminder to re-open it when we can drop node 12.

PR #509, opened 2022-03-10
fix(legacy-scripting-runner): replace deasync with synckit

eliangcs:
Attempts to fix https://github.com/zapier/zapier/issues/53430.

[deasync](https://github.com/abbr/deasync) is problematic. We've seen Lambda calls hang and never resolve because of the sync usage of `z.request` in legacy scripting as follows:

```js
var Zap = {
  key_post_write: function(bundle) {
    // this could hang!
    var response = z.request({ ... });
    return { ... };
  },
  
  another_post_write: function(bundle, callback) {
    // on the other hand, callback always works
    z.request({ ... }, function(err, response) {
      ...
      callback(null, { ... });
    });
  }
};
```

Oddly, the `z.request()` call itself is not the one who hangs. We can actually see `post_write` returns correctly. But somehow the Lambda handler just doesn't reach to an end. When Lambda doesn't respond, Zapier retries, which causes duplicate items to be created if the Zap step is a write action.

There are some issues reported on GitHub that seem related: https://github.com/abbr/deasync/issues/21, https://github.com/abbr/deasync/issues/55, https://github.com/abbr/deasync/issues/151. None of them seem conclusive.

It's hard to figure out why because it happens intermittently. In addition, the way how deasync [uses](https://github.com/abbr/deasync/blob/50cc7daba1edd6be77d1b7e0bc297933fefbd1f2/src/deasync.cc) v8 C++ API makes it even harder to debug (at least for me).

I've found by replacing deasync with [synckit](https://github.com/rx-ts/synckit) solves the problem. I tested this patch with some production apps and confirmed it does fix the issue.

synckit uses Node.js' [work_threads](https://nodejs.org/docs/latest-v14.x/api/worker_threads.html) module, which is less intrusive on the Node.js event loop than deasync. This also means we can say goodbye to those deasync binary bindings!


xavdid (dismissed):
Very cool!

long term, we might also want to look into replacing the deprecated `request` package, but I think doing these big replacements one at a time is the way to go. 👍

xavdid (approved):

PR #513, opened 2022-03-17
fix(core): fix another scenario that causes hanging (11.x)

eliangcs:
Related to #509, #490, and https://github.com/zapier/zapier/issues/53430, this addresses another scenario that causes the Lambda handler to hang/freeze.

Suppose a converted app has this (bad) legacy script:

```js
var Zap = {
  bad_post_write: function(bundle, callback) {
    z.request({ ... }, function(err, response) {
      if (err) {
        callback(err);
      } else {
        callback(null, z.JSON.parse(response.content));
      }
    });
    callback(null, { message: 'ok' }); // double callback?!
  }
};
```

Notice how the `post_write` function returns immediately by calling `callback(null, { message: 'ok' })` while makes an outbound `z.request()` and calls `callback` asynchronously.

When legacy-scripting-runner [promisifies](https://github.com/zapier/zapier-platform/blob/1632992ad4867b6cce1e8abb8fe60a866ba19fbc/packages/legacy-scripting-runner/index.js#L740) this `post_write` function, the `callback` in the `z.request` will come after the Lambda handler [returns](https://github.com/zapier/zapier-platform/blob/1632992ad4867b6cce1e8abb8fe60a866ba19fbc/packages/core/src/tools/create-lambda-handler.js#L164). Since we only [end](https://github.com/zapier/zapier-platform/blob/1632992ad4867b6cce1e8abb8fe60a866ba19fbc/packages/core/src/tools/create-lambda-handler.js#L161) the logger once when the Lambda handler finishes, the HTTP logger corresponding to the `z.request` call, which comes after the Lambda handler, will stay open and never end until a long timeout.

To fix, we need to make sure to end any loggers allocated after the Lambda handler returns.

### DIFF
--- a/packages/core/integration-test/integration-test.js
+++ b/packages/core/integration-test/integration-test.js
@@ -3,6 +3,7 @@
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const { promisify } = require('util');
 
 const AWS = require('aws-sdk');
 const crypto = require('crypto');
@@ -12,6 +13,8 @@ const should = require('should');
 const createLambdaHandler = require('../src/tools/create-lambda-handler');
 const mocky = require('../test/tools/mocky');
 const { HTTPBIN_URL } = require('../test/constants');
+
+const sleep = promisify(setTimeout);
 
 const lambda = new AWS.Lambda({
   apiVersion: '2015-03-31',
@@ -55,8 +58,9 @@ const runLocally = (event) => {
     handler(event, {}, (err, data) => {
       if (err) {
         reject(err);
+      } else {
+        resolve(data);
       }
-      resolve(data);
     });
   });
 };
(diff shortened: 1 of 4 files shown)


packages/core/integration-test/integration-test.js
```diff
@@ -55,8 +58,9 @@ const runLocally = (event) => {
     handler(event, {}, (err, data) => {
       if (err) {
         reject(err);
+      } else {
+        resolve(data);
```
eliangcs, line 62: Fixed a problem while I was here. We should call either `resolve` or `reject` but not both.

packages/core/test/userapp/index.js
```diff
@@ -514,6 +516,39 @@ const ExecuteCallbackRequest = {
...
+            });
+            res.on('end', () => {
+              // Set a global variable so we have something to assert in the test
+              // to prove we can reach here
+              process.teapot = body;
+              // body === "I'm a teapot!"
+            });
+          })
+          .end();
+        return { message: 'ok' };
```
eliangcs, line 546: This code has essentially the same effect as the "double callback" example I wrote in the PR description.

eliangcs:
@xavdid sorry I haven't reviewed your PRs. I wanted to merge this PR to 11.x before your PRs targeting 12.x. I'll make sure to review your PRs my tomorrow! 🙇

eliangcs:
@xavdid can you also review/approve this PR when you're online? Thanks!

xavdid (approved):
Ah, sorry! I approved the other one and I guess I missed this one in the shuffle.

I'll hold off merging mine so you have a clean path to get these out.

PR #514, opened 2022-03-18
fix(core): fix another scenario that causes hanging (9.x)

eliangcs:
Backport #513 to 9.x.

### DIFF
--- a/packages/core/integration-test/integration-test.js
+++ b/packages/core/integration-test/integration-test.js
@@ -3,6 +3,7 @@
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const { promisify } = require('util');
 
 const AWS = require('aws-sdk');
 const crypto = require('crypto');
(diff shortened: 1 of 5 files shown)


packages/core/integration-test/integration-test.js
```diff
@@ -520,47 +524,94 @@ const doTest = runner => {
...
+        // Hopefully long enough for the callback to complete
+        await sleep(2000);
+
+        process.teapot.should.eql("I'm a teapot!");
+
+        const logs = mocky.getLogs();
+        logs.length.should.eql(1);
+
+        const log = logs[0];
+        log.message.should.startWith(`418 GET https://httpbin.`);
```
eliangcs, line 599: Changed this from `should.eql` to `should.startWith` because the URL in the log message was censored. Apparently, GitHub Actions has a `zapier` in one of the environment variables and our log sanitizer uses to censor `https://httpbin.zapier-tooling.com` into `https://httpbin.:censored:6:e19648f405:-tooling.com`.

xavdid (approved):

PR #517, opened 2022-03-22
docs(core): document changes to autoRefresh

xavdid:
Documentation for #512. The app-wide `skipThrowForStatus` is intentionally not documented because I don't anticipate CLI devs using that - they can do it in middleware if they'd like. 

Did some other cleaning while I was in there.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -1486,62 +1486,14 @@ credentials. (The runs will be
 [Held](https://zapier.com/help/manage/history/view-and-manage-your-zap-history#holding),
 and the user will be able to replay them after reconnecting.)
 
-Example: `throw new z.errors.ExpiredAuthError('Your message.');`
+Example: `throw new z.errors.ExpiredAuthError('You must manually reconnect this auth.');`
 
-For apps that use OAuth2 with `autoRefresh: true` or Session Auth, the core injects
+For apps that use OAuth2 with `autoRefresh: true` or Session Auth, `core` injects
 a built-in `afterResponse` middleware that throws an error when the response status
 is 401. The error will signal Zapier to refresh the credentials and then retry the
-failed operation. For some cases, e.g, your server doesn't use the 401 status
-for auth refresh, you may have to throw the `RefreshAuthError` on your own,
-which will also signal Zapier to refresh the credentials.
+failed operation. You can also throw this error manually if your server doesn't use the 401 status or you want to trigger an auth refresh even if the credentials aren't stale.
 
 Example: `throw new z.errors.RefreshAuthError();`
-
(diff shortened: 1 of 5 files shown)


packages/cli/README-source.md
```diff
@@ -1108,12 +1108,33 @@ Middleware functions can be asynchronous - just return a promise from the middle
...
+7. log the request to Zapier's logging server
+8. if the status code is `401`, you're using a refresh-able auth (such as `oauth2` or `session`) _and_ `autoRefresh` is `true` in your auth configuration, throw a `RefreshAuthError`. The server will attempt to refresh the authentication again and retry the whole step
+9. run your `afterResponse` middleware functions in order
+10. call `response.throwForStatus()` unless `response.skipThrowForStatus` is `true`
+
+The resulting response object is returned from `z.request()`.
+
 #### Error Response Handling
 
 Since v10, we call `response.throwForStatus()` before we return a response. You can prevent this by setting `skipThrowForStatus` on the request or response object. You can do this in `afterResponse` middleware if the API uses a status code >= 400 that should not be treated as an error.
```
eliangcs, line 1128: I found this paragraph is bit confusing. I suggest we clarify:

```suggestion
Since v10, `z.request()` calls `response.throwForStatus()` before it returns a response. You can disable automatic error throwing by setting `skipThrowForStatus` on the request object:

// Disable automatic error throwing on the request object
const perform = async (z, bundle) => {
  const response = await z.request({
    url: '...',
    skipThrowForStatus: true
  });
  // Now you handle error response on your own.
  // The following is equivalent to response.throwForStatus().
  if (response.status >= 400) {
    throw new z.errors.ResponseError(response);
  }
};

You can also do it in `afterResponse` if the API uses a status code >= 400 that should not be treated as an error.

// Don't throw an error when response status is 456
const disableAutoErrorThrowing = (response, z) => {
  if (response.status === 456) {
    response.skipThrowForStatus = true;
  }
  return response;
};

const App = {
  // ...
  afterResponse: [disableAutoErrorThrowing],
  // ...
};

eliangcs (approved):
Looks good! I suggest we clarify on a paragraph.

In addition, I think we should also edit the [v10 Breaking Change: Auth Refresh](https://github.com/zapier/zapier-platform/blob/3c95a924b9bf896830ff4be608659eeb392b113e/packages/cli/README-source.md#v10-breaking-change-auth-refresh) section, because it would look outdated when v12 is out. But that doesn't have to be in this PR. Your call.

xavdid:
Ah, good call! So many sections.

I've gone ahead and pulled that whole section- it's not relevant in the 12.0 docs, but will live on in the older ones if we need it.

PR #520, opened 2022-03-24
fix(core): tweak global `skipThrowForStatus` flag to ignore shorthand requests

xavdid:
In `9.x`, shorthand requests (using objects instead of JS functions) _always_ threw for bad responses. We want to maintain that behavior in the 12.x release line. But, the global `skipThrowForStatus` changed the defaults for all requests, not just shorthand. So, this adds an explicit `skipThrowForStatus: false` default to shorthand requests. Native devs can still have specified `skip: true` for shorthand requests. This keeps the default behavior for 9.x shorthand requests even when we set the global flag on all godzilla apps 😵‍💫😵‍💫😵‍💫

Here's a table I made to help myself reason about the change:

> Behavior of a failing shorthand request when migrated between core versions (9 and 12 should match)

|global flag (on 12) | req flag | 9.x   | 10-11.x | 12.x  |
 | ----------- | -------- | ----- | ------- | ----- |
 | false       | false    | throw | throw   | throw |
 | true        | false    | throw | pass    | pass❗️  |
 | false       | true     | pass | pass    | pass  |
 | true        | true     | pass | pass    | pass  |

❗ == bug introduced, should throw

if shorthand request opts out, honor that. Otherwise, always throw (ignoring new flag)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ _released `2022-03-24`_
 
 ### core
 
-- Fix regression where the global `skipThrowForStatus` incorrectly applied to shorthand requests. It's only intended to modify the behavior of requests made with `z.request()`. The docs and changelog have been updated accordingly ([#520](https://github.com/zapier/zapier-platform/pull/520))
+- :bug: Fix regression where the global `skipThrowForStatus` incorrectly applied to shorthand requests. It's only intended to modify the behavior of requests made with `z.request()`. The docs and changelog have been updated accordingly ([#520](https://github.com/zapier/zapier-platform/pull/520))
 
 ### schema
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ _released `2022-03-24`_
 
 ### core
 
-- Fix regression where the global `skipThrowForStatus` incorrectly applied to shorthand requests. It's only intended to modify the behavior of requests made with `z.request()`. The docs and changelog have been updated accordingly.
+- Fix regression where the global `skipThrowForStatus` incorrectly applied to shorthand requests. It's only intended to modify the behavior of requests made with `z.request()`. The docs and changelog have been updated accordingly ([#520](https://github.com/zapier/zapier-platform/pull/520))
 
 ### schema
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
...
+
+_released `2022-03-24`_
+
+### cli
+
+- None
+
+### core
+
+- Fix regression where the global `skipThrowForStatus` incorrectly applied to shorthand requests. It's only intended to modify the behavior of requests made with `z.request()`. The docs and changelog have been updated accordingly.
```
eliangcs, line 11:
```suggestion
- :bug: Fix regression where the global `skipThrowForStatus` incorrectly applied to shorthand requests. It's only intended to modify the behavior of requests made with `z.request()`. The docs and changelog have been updated accordingly. ([#520](https://github.com/zapier/zapier-platform/pull/520))
```

packages/core/test/create-app.js
```diff
@@ -665,4 +665,155 @@ describe('create-app', () => {
...
+          method,
+        };
+        const { results } = await app(
+          createInput(appDefinition, event, testLogger)
+        ).should.be.fulfilled();
+
+        results[0].status.should.eql(400);
+        results[0].message.should.eql('mocked response');
+      });
+    });
```
eliangcs, line 817: I suggest we add another test case `should skip throw if global flag is set and local flag is explicitly false` under `describe('z.request requests')`. In that case, `z.request` should _not_ throw.

eliangcs (dismissed):
Thanks for fixing the bug so quickly! The logic looks correct. See comments below for a few non-blocking suggestions.

xavdid:
Great suggestions! Just pushed those up.

packages/core/test/create-app.js
```diff
@@ -665,4 +665,182 @@ describe('create-app', () => {
...
+        };
+        const { results } = await app(
+          createInput(appDefinition, event, testLogger)
+        ).should.be.fulfilled();
+
+        results[0].status.should.eql(400);
+        results[0].message.should.eql('mocked response');
+      });
+
+      it('should skip throw if local flag is false and global flag is set', async () => {
```
eliangcs, line 818: The test case description seems inaccurate. I'll go ahead and apply the commit.

```suggestion
      it('should throw if local flag is false and global flag is set', async () => {
```

eliangcs (approved):
👍

eliangcs (approved):

PR #522, opened 2022-03-30
fix(core): properly replace curlies that match multiple times in a row

xavdid:
Fixes bug introduced in core version `11.1.1` where a quirk with regex objects wouldn't replace multiple curly strings in a row


standielpls (approved):
lgtm

eliangcs:
Thanks for fixing! For reference, the bug was introduced by https://github.com/zapier/zapier-platform/pull/428. And this is not the first time we get bitten by `RegExp.test()`: https://github.com/zapier/zapier-platform/pull/467.

PR #524, opened 2022-04-01
fix(core): log censoring is slow if response has many URLs

eliangcs:
Fixes an issue where log censoring becomes too slow when an app gives a response that has many (like thousands) URLs. The new test case should be self-explanatory.

### DIFF
--- a/packages/core/src/tools/create-logger.js
+++ b/packages/core/src/tools/create-logger.js
@@ -17,8 +17,11 @@ const {
   findSensitiveValues,
   recurseExtract,
 } = require('@zapier/secret-scrubber');
-// not really a public function, but it came from here originally
-const { isUrlWithSecrets } = require('@zapier/secret-scrubber/lib/convenience');
+// not really public functions, but they came from here originally
+const {
+  isSensitiveKey,
+  isUrlWithSecrets,
+} = require('@zapier/secret-scrubber/lib/convenience');
 
 // The payload size per request to stream logs. This should be slighly lower
 // than the limit (16 MB) on the server side.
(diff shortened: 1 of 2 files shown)


xavdid (approved):
Ok, so the salient change here is that instead of using `findSensitiveValues(data)` (which finds sensitive keys and urls w/ querystrings), we instead only find sensitive keys (under the assumption that the logged body won't contain any _new_ secrets not already present in `authData`.

I'm a little wary of censoring _fewer_ things, but that does seem like a corner case. All else equal, I'd have focused on speed improvements in secret scrubber (if possible), but I think this is a reasonable first step.

Thanks for digging in here!

eliangcs:
@xavdid I was wondering why we were censoring all the URLs (with query params) from the log extra in the first place. Did we do that to fix another issue where logs are not censored enough? If yes, merging this PR would create another regression (where logs are not censored enough).

> I'd have focused on speed improvements in secret scrubber (if possible)

I also think fixing in secret-scrubber is better. Can you do it? We don't have to rush to merge this PR, since we can downgrade an app to 9.7.3 to fix the timeout issue.

xavdid:
> Did we do that to fix another issue where logs weren't censored enough?

Yes, I think in a couple of different places. Especially during auth refresh, urls will have query params that need to be censored (but have new values that aren't in `authData`, since they're novel tokens). I think https://github.com/zapier/zapier-platform/pull/85 is one such example, but I bet there are more in the monorepo.

I think it's fair to exclude the request body from checking sensitive urls - that _does_ seem like a corner case. But, we should probably keep checking them for auth requests and in messages itself. 

I think in interest of not causing regressions, downgrading affected apps for now is the way to go and I can take a pass at profiling and improving the url censoring in scrubber. If I can't get to the bottom of it in a day or two, we can move forward here and adjust if we notice any regressions.

packages/core/src/tools/create-logger.js
```diff
@@ -119,10 +123,18 @@ const buildSensitiveValues = (event, data) => {
     }
     return true;
   });
+
+  // For log extra data, we only want to censor with sensitive keys. URLs in log
+  // extra should be safe.
+  const sensitiveLogData = recurseExtract(
+    data,
+    (key, value) => _.isString(value) && isSensitiveKey(key)
```
xavdid, line 131: sensitive key -> numeric value should also be caught. eg `{secretNumber: 123456789}`

PR #525, opened 2022-04-06
fix(core): ensure string content is parsed pre-logging in case it contains secrets

xavdid:
We've noticed a couple of edge cases where `access_tokens` sometimes aren't censored correctly in logs. Most of our censoring works by looking for explicit data that should be removed. But, during an auth refresh, we get a token back we've never seen before. So, we have to make sure to (at least attempt to) parse the response, search it for secrets, and censor those for logging.

### DIFF
--- a/packages/core/src/tools/create-logger.js
+++ b/packages/core/src/tools/create-logger.js
@@ -108,6 +108,17 @@ const toStdout = (event, msg, data) => {
   }
 };
 
+// try to parse json; if successful, find secrets in it
+const attemptFindSecretsInStr = (s) => {
+  let parsedRespContent;
+  try {
+    parsedRespContent = JSON.parse(s);
+  } catch {
+    return [];
+  }
+  return findSensitiveValues(parsedRespContent);
+};
+
 const buildSensitiveValues = (event, data) => {
   const bundle = event.bundle || {};
   const authData = bundle.authData || {};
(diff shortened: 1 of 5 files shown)


packages/core/test/logger.js
```diff
@@ -210,52 +235,88 @@ describe('logger', () => {
...
     const bundle = {
       authData: {
-        refresh_token: 'whatever',
+        refresh_token: 'very_secret',
       },
     };
     const logger = createlogger({ bundle }, options);
 
-    const data = {
-      response_json: {
```
xavdid, line 228: the crux of why this test wasn't catching what it was supposed to was that when we log an http request, the body for the `req` and `res` is a string. So `response_json` being here let us find the secrets when we actually don't. If you removed this property, the test failed (hence the logged secrets)

packages/core/src/tools/create-logger.js
```diff
@@ -119,11 +130,24 @@ const buildSensitiveValues = (event, data) => {
...
+  // (such as what comes back in the response during an auth refresh)
+
+  for (const prop of ['response_content', 'request_data']) {
+    if (data[prop]) {
+      result.push(...attemptFindSecretsInStr(data[prop]));
+    }
+  }
+
+  // unique- no point in duplicates
+  return [...new Set(result)];
```
xavdid, line 150: this should also help keep our "secrets to scan" length low

packages/core/test/logger.js
```diff
@@ -270,7 +331,7 @@ describe('logger', () => {
     const logger = createlogger({ bundle }, options);
 
     const data = {
-      response_json: {
+      whatever: {
```
xavdid, line 334: wanted to make sure these didn't make assumptions about the fact that the response was present in parsed form

eliangcs (approved):
Looks good!

PR #526, opened 2022-04-06
fix(core): censor novel secrets in querystring

xavdid:
Follow up to https://github.com/zapier/zapier-platform/pull/525 that covers similar behavior in the querystring.

---

Didn't catch it before because I passed the url `example.com?api_key=secret`, but prep request separates them, so we have to search `request_params` explicitly too.

Underscores the importance of our unit tests matching what actually gets called at runtime

### DIFF
--- a/packages/core/src/tools/create-logger.js
+++ b/packages/core/src/tools/create-logger.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { Transform } = require('stream');
+const { parse: querystringParse } = require('querystring');
 
 const _ = require('lodash');
 
@@ -18,7 +19,10 @@ const {
   recurseExtract,
 } = require('@zapier/secret-scrubber');
 // not really a public function, but it came from here originally
-const { isUrlWithSecrets } = require('@zapier/secret-scrubber/lib/convenience');
+const {
+  isUrlWithSecrets,
+  isSensitiveKey,
+} = require('@zapier/secret-scrubber/lib/convenience');
 
 // The payload size per request to stream logs. This should be slighly lower
 // than the limit (16 MB) on the server side.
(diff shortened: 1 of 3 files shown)


packages/core/test/logger.js
```diff
@@ -14,22 +14,29 @@ const {
 } = require('../src/http-middlewares/after/log-response');
 
 // little helper to prepare a req/res pair like the http logger does
-const prepareTestRequest = (reqBody = {}, resBody = {}) =>
+const prepareTestRequest = ({
+  reqBody = {},
```
xavdid, line 18: Declaring params like this make them _mostly_ function like python kwargs

packages/core/src/tools/create-logger.js
```diff
@@ -124,7 +128,13 @@ const buildSensitiveValues = (event, data) => {
   const authData = bundle.authData || {};
   // for the most part, we should censor all the values from authData
   // the exception is safe urls, which should be filtered out - we want those to be logged
+  // but, we _should_ censor-no-matter-what sensitive keys, even if their value is a safe url
```
xavdid, line 131: weirdly, the `logger.safe urls` test was failing for me locally, but not in CI in the last PR. Bumping `secret-scrubber` locally didn't seem to change that. Not sure why CI didn't flag anything ¯\\\_(ツ)_/¯

eliangcs (approved):
👍

packages/core/src/tools/create-logger.js
eliangcs: That's odd. It passes for me locally. Just curious, what's your failing message?

xavdid: Right? Weirdest thing. 

Running `yarn main-tests -g "safe urls"` in `core` directory gives:

```
      -      "password": "https://a-url-like.password.com"
      +      "password": ":censored:31:0dbc81268a:"
```

Same thing happens when running the full suite.

It sort of makes sense- the value is a url and doesn't have basic auth or a querystring. So I'm not sure why the test passed before. But, I think the current behavior is what we want. If something is explicitly a sensitive key, we should censor it. If it's not (but is in auth data), it's probably safe.

PR #527, opened 2022-04-06
SUPPORTESC-307] docs(cli): Auth link update and wording cleanup

armchairlinguist:
The auth section needed some love, including clarification, additional cross-linking, and updated external links. My [previous effort](https://github.com/zapier/zapier-platform/pull/488) on this focused on authData vs inputData a bit more, but in this version I've stuck more closely to making clarifications of exact behavior rather than attempting any conceptual regrouping.


xavdid (approved):
looks awesome!

PR #530, opened 2022-04-12
[PDE-3149] Use async/await in Trigger example app

reganstarr:
Part of hack week q1 2022 https://coda.io/d/HackWeek-2022-Q1_d_CnAb9OV3_/Improve-Dev-Platform-Docs_susOs#_luWyc

This PR assumes that async/await is the preferred approach. Perhaps it is not.

### DIFF
--- a/example-apps/trigger/README.md
+++ b/example-apps/trigger/README.md
@@ -2,6 +2,4 @@
 
 [![Build Status](https://travis-ci.org/zapier/zapier-platform-example-app-trigger.svg?branch=master)](https://travis-ci.org/zapier/zapier-platform-example-app-trigger)
 
-A barebones app that has a resource defined.
-
-> We recommend using the zapier-platform-cli and `zapier init . --template=trigger` to create an app.
+A barebones app that has a trigger defined.
--- a/example-apps/trigger/test/triggers.js
+++ b/example-apps/trigger/test/triggers.js
@@ -3,46 +3,44 @@
 require('should');
 
 const zapier = require('zapier-platform-core');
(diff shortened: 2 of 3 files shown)


example-apps/trigger/README.md
```diff
@@ -2,6 +2,4 @@
 
 [![Build Status](https://travis-ci.org/zapier/zapier-platform-example-app-trigger.svg?branch=master)](https://travis-ci.org/zapier/zapier-platform-example-app-trigger)
 
-A barebones app that has a resource defined.
-
-> We recommend using the zapier-platform-cli and `zapier init . --template=trigger` to create an app.
```
reganstarr, line 7: This is no longer supported.

xavdid: This is replaced by `zapier init .` and `zapier scaffold trigger my_trigger`

xavdid (approved):
Cool! 

For the contents of `example-apps`, I don't _believe_ those are still in use. I think older CLI versions depended on them existing in github, but IIRC (which I may not, it's been a wihle) they're on the way out. 

Instead, we were going to put individual files in the `examples` directory and use those in conjunction with `scaffold` to make working apps (rather than trying to maintain a bunch of micro apps). 

Here's the trigger that gets scaffolded out today: https://github.com/zapier/zapier-platform/blob/master/packages/cli/scaffold/trigger.template.js

So, all that is to say, this is totally fine to merge, but I wouldn't spend too much more time in the `example-apps` directory this week. Thank you for your effort though!

PR #531, opened 2022-04-13
[PDE-3149] Update example-apps links to recommend `zapier init`

tylerrobertsonzapier:
Part of _Improve Dev Platform Docs_ for HackWeek 2022 Q1

As the example-apps folder is primarily used for older CLI versions now, we should update the links to point users towards using `zapier init` to build new projects as the first suggestion, with links to example-apps only as reference.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1156,7 +1156,7 @@ zapier convert 1234 --version 1.0.1 my-app
   <div class="row-height">
     <div class="col-md-5 col-sm-12 col-height  docs-primary">
       <p>Useful if your app requires two pieces of information to authenticate: <code>username</code> and <code>password</code>, which only the end user can provide. By default, Zapier will do the standard Basic authentication base64 header encoding for you (via an automatically registered middleware).</p><blockquote>
-<p>To create a new integration with basic authentication, run <code>zapier init [your app name]</code> and select &quot;basic-auth&quot; from the list that appears. You can also review an example of that code <a href="https://github.com/zapier/zapier-platform/tree/master/example-apps/basic-auth">here</a>.</p>
+<p>To create a new integration with basic authentication, run <code>zapier init [your app name] --template basic-auth</code>. You can also review an example of that code <a href="https://github.com/zapier/zapier-platform/tree/master/example-apps/basic-auth">here</a>.</p>
 </blockquote><p>If your app uses Basic auth with an encoded API key rather than a username and password, like <code>Authorization: Basic APIKEYHERE:x</code>, consider the <a href="#custom">Custom</a> authentication method instead.</p>
     </div>
     <div class="col-md-7 col-sm-12 col-height  docs-code">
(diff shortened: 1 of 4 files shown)


docs/index.html
```diff
@@ -1814,13 +1814,7 @@ <h2 id="triggerssearchescreates">Triggers/Searches/Creates</h2>
...
-</blockquote><blockquote>
-<p>Example App: Check out <a href="https://github.com/zapier/zapier-platform/tree/master/example-apps/rest-hooks">https://github.com/zapier/zapier-platform/tree/master/example-apps/rest-hooks</a> for a working example app using REST hook triggers.</p>
-</blockquote><blockquote>
-<p>Example App: Check out <a href="https://github.com/zapier/zapier-platform/tree/master/example-apps/search">https://github.com/zapier/zapier-platform/tree/master/example-apps/search</a> for a working example app using searches.</p>
-</blockquote><blockquote>
-<p>Example App: Check out <a href="https://github.com/zapier/zapier-platform/tree/master/example-apps/create">https://github.com/zapier/zapier-platform/tree/master/example-apps/create</a> for a working example app using creates.</p>
+<p>To create a new integration with a premade trigger, search, or create, run <code>zapier init [your app name]</code> and select from the list that appears. You can also check out our working example apps <a href="https://github.com/zapier/zapier-platform/tree/master/example-apps">here</a>.</p>
```
reganstarr, line 1817: @tylerrobertsonzapier Should we also direct devs to use `zapier scaffold` after their `init` as mentioned in [this comment](https://github.com/zapier/zapier-platform/pull/530#pullrequestreview-940056946)?

tylerrobertsonzapier: @reganstarr Great call out! I don't think `zapier scaffold` is strictly necessary here, as the templates with `init` will already have scaffolded triggers/actions/auths/etc. added in. Unless I'm wrong about that?

Regardless: we should definitely make sure we're calling out `zapier scaffold` _somewhere_ as a way to quickly add triggers and actions.

xavdid: Not quite, the minimal example is _truly_ minimal. This is the `index.js`:

```js
module.exports = {
  // This is just shorthand to reference the installed dependencies you have.
  // Zapier will need to know these before we can upload.
  version: require('./package.json').version,
  platformVersion: require('zapier-platform-core').version,

  // If you want your trigger to show up, you better include it here!
  triggers: {},

  // If you want your searches to show up, you better include it here!
  searches: {},

  // If you want your creates to show up, you better include it here!
  creates: {},

  resources: {},
};
```

So scaffolding should definitely be part of the story!

reganstarr: I tried out `zapier init` for custom-auth and it did not create any triggers/actions/searches.

packages/cli/README-source.md
```diff
@@ -309,7 +309,7 @@ When a user authenticates to your application through Zapier, a "connection" is
 
 Useful if your app requires two pieces of information to authenticate: `username` and `password`, which only the end user can provide. By default, Zapier will do the standard Basic authentication base64 header encoding for you (via an automatically registered middleware).
 
-> Example App: Check out https://github.com/zapier/zapier-platform/tree/master/example-apps/basic-auth for a working example app for basic auth.
+> To create a new integration with basic authentication, run `zapier init [your app name]` and select "basic-auth" from the list that appears. You can also review an example of that code [here](https://github.com/zapier/zapier-platform/tree/master/example-apps/basic-auth).
```
xavdid, line 312: For each of these, you can provide the flag in the command:

> `zapier init [your app name] --template basic-auth`

So I'd include that command in full so people can just paste it.

packages/cli/README-source.md
```diff
@@ -509,13 +509,7 @@ The definition for each of these follows the same structure. Here is an example
...
 [Search Schema](https://github.com/zapier/zapier-platform/blob/master/packages/schema/docs/build/schema.md#searchschema), and [Create Schema](https://github.com/zapier/zapier-platform/blob/master/packages/schema/docs/build/schema.md#createschema).
 
-> Example App: Check out https://github.com/zapier/zapier-platform/tree/master/example-apps/trigger for a working example app using triggers.
-
-> Example App: Check out https://github.com/zapier/zapier-platform/tree/master/example-apps/rest-hooks for a working example app using REST hook triggers.
-
-> Example App: Check out https://github.com/zapier/zapier-platform/tree/master/example-apps/search for a working example app using searches.
-
-> Example App: Check out https://github.com/zapier/zapier-platform/tree/master/example-apps/create for a working example app using creates.
+> To create a new integration with a premade trigger, search, or create, run `zapier init [your app name]` and select from the list that appears. You can also check out our working example apps [here](https://github.com/zapier/zapier-platform/tree/master/example-apps).
```
xavdid, line 512: The examples are mostly auth-based. The only ones that have triggers are the ones that need it for the example (eg `dynamic-dropdown`). This is where I'd plug `zapier scaffold`

docs/index.html
tylerrobertsonzapier: Thank you both! Updating that line that @reganstarr pointed out to include `zapier scaffold`.

packages/cli/README-source.md
tylerrobertsonzapier: Good shout! Adding those now.

xavdid (approved):

xavdid (approved):

PR #537, opened 2022-04-15
PDE-3149 Added SSO sign on instructions

justin-lee-zapier:
This PR updates the Getting Started section of the CLI docs here: https://github.com/zapier/zapier-platform/blob/master/packages/cli/README.md#quick-setup-guide

At the moment, it's not clear on how you would login to Zapier through the CLI platform if your Zapier account was created using SSO. 

The terminal will give you steps on how to login via SSO:

However, reviewing Zendesk, I did find a handful of tickets where developers wrote in having issues searching "CLI SSO". This is one example: https://zapier.zendesk.com/agent/tickets/62778

And this more recent one: https://zapier.zendesk.com/agent/tickets/587189

I worked with Sydney on the paid support team to get this Look: https://zapier.looker.com/looks/5957

And found that ~58k developers use SSO.

I think that by bringing the SSO information upfront in our docs, we can help reduce a few obstacles for new developers.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -92,6 +92,7 @@ npm install -g zapier-platform-cli
 # setup auth to Zapier's platform with a deploy key
 zapier login
 ```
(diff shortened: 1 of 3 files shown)


xavdid (approved):
awesome!

PR #538, opened 2022-04-21
PDE-3152 feat(cli) - Update `team:get` CLI command to fetch Collaborator.

standielpls:
JIRA: https://zapierorg.atlassian.net/browse/PDE-3152

Removes `subscribers` from `team:get` command
Fetches `limited collaborators` for `team:get` command.

### DIFF
--- a/packages/cli/src/oclif/commands/team/get.js
+++ b/packages/cli/src/oclif/commands/team/get.js
@@ -2,27 +2,22 @@ const ZapierBaseCommand = require('../../ZapierBaseCommand');
 const { cyan } = require('colors/safe');
 const { listEndpointMulti } = require('../../../utils/api');
 const { buildFlags } = require('../../buildFlags');
-const { BASE_ENDPOINT } = require('../../../constants');
 
 class TeamListCommand extends ZapierBaseCommand {
   async perform() {
     this.startSpinner('Loading team members');
-    const { admins, subscribers } = await listEndpointMulti(
+    const { admins, limitedCollaborators } = await listEndpointMulti(
       { endpoint: 'collaborators', keyOverride: 'admins' },
-      {
-        endpoint: (app) =>
-          `${BASE_ENDPOINT}/api/platform/v3/integrations/${app.id}/subscribers`,
-        keyOverride: 'subscribers',
-      }
+      { endpoint: 'limited_collaborators', keyOverride: 'limitedCollaborators' }
     );
 
     this.stopSpinner();
 
-    const cleanedUsers = [...admins, ...subscribers].map(
+    const cleanedUsers = [...admins, ...limitedCollaborators].map(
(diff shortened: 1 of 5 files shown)


xavdid:
cool, so the code here looks good. The only note is that this as written should be marked as a breaking change. If I have CLI code that reads subscribers, then it would break after this update. eg:

```
zapier team:get --format json | jq '.subscribers'
```

So, either we:

- mark this as a major version and release it when we do that next (probably ~ may, when the new node version rotates)
- use the new endpoint, but output the data under the old key (assuming the structure of what comes out of the API call is otherwise identical)

I'm good either way, just wanted to flag!

eliangcs:
@xavdid that's a good callout! Sorry, I didn't think of people might have workflows relying on the `subscribers` field when writing the ticket [PDE-3152](https://zapierorg.atlassian.net/browse/PDE-3152). I'd suggest we keep it backward compatible. So how about we do these:

- the `team:get` still gets `subscribers` from the `/subscribers` endpoint
- when the time comes, we deprecate the `/subscribers` endpoint by making it return an empty result ([PDE-3161](https://zapierorg.atlassian.net/browse/PDE-3161))

> use the new endpoint, but output the data under the old key (assuming the structure of what comes out of the API call is otherwise identical)

FWIW, `subscriber` and `limited_collaborator` are not the same, so we probably shouldn't use the old key `subscribers` for `limited_collaborators`.

@standielpls does that make sense?

standielpls:
Yep that makes sense! 
As `subscriber` and `limited_collaborator` are not the same, when we display the table, the user can be 1 of 3 roles `admin`, `subscribers` and `collaborator`.

packages/cli/src/oclif/commands/team/get.js
```diff
@@ -48,10 +60,11 @@ class TeamListCommand extends ZapierBaseCommand {
 TeamListCommand.flags = buildFlags({ opts: { format: true } });
 TeamListCommand.description = `Get team members involved with your integration.
 
-These users come in two levels:
+These users come in three levels:
 
   * \`admin\`, who can edit everything about the integration
-  * \`subscriber\`, who can't directly access the app, but will receive periodic email updates. These updates include quarterly health socores and more.
+  * \`subscriber\`, who can't directly access the app, but will receive periodic email updates. These updates include quarterly health scores and more.
+  * \`collaborator\`, who has read-only access for the app, but will receive periodic email updates. These updates include quarterly health scores and more.
```
eliangcs, line 67: Same comment as [this](https://github.com/zapier/zapier-platform/pull/539#discussion_r857359686). Maybe  `collaborator` should go before `subscriber`?

eliangcs (approved):

eliangcs (approved):
💯

PR #539, opened 2022-04-22
PDE-3153 feat(cli): Add limited collaborator role to `team:add` command.

standielpls:
JIRA: https://zapierorg.atlassian.net/browse/PDE-3153

Add limited_collaborator as another option to `team:add` command.

### DIFF
--- a/packages/cli/src/oclif/commands/team/add.js
+++ b/packages/cli/src/oclif/commands/team/add.js
@@ -33,7 +33,10 @@ class TeamAddCommand extends ZapierBaseCommand {
 
     const url = roleIsAdmin
       ? `/apps/${id}/collaborators`
-      : `https://zapier.com/api/platform/v3/integrations/${id}/subscribers`;
+      : this.args.role === 'subscriber'
+      ? `https://zapier.com/api/platform/v3/integrations/${id}/subscribers`
+      : `/apps/${id}/limited_collaborators`;
+
     await callAPI(url, {
       url: url.startsWith('http') ? url : undefined,
       method: 'POST',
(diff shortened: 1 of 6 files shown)


xavdid (approved):

packages/cli/src/oclif/commands/team/add.js
```diff
@@ -66,16 +69,18 @@ TeamAddCommand.args = [
 TeamAddCommand.flags = buildFlags();
 TeamAddCommand.description = `Add a team member to your integration.
 
-These users come in two levels:
+These users come in three levels:
 
   * \`admin\`, who can edit everything about the integration
-  * \`subscriber\`, who can't directly access the app, but will receive periodic email updates. These updates include quarterly health socores and more.
+  * \`subscriber\`, who can't directly access the app, but will receive periodic email updates. These updates include quarterly health scores and more.
+  * \`collaborator\`, who has read-only access for the app, but will receive periodic email updates. These updates include quarterly health scores and more.
```
eliangcs, line 76: As we're deprecating `subscriber` and `collaborator` is technically a superset of `subscriber`, maybe it makes more sense to move `collaborator` before `subscriber`? The same applies to `TeamAddCommand.args` and `TeamAddCommand.examples` as well.

On the help text, the "but" should probably be an "and":

```suggestion
  * \`collaborator\`, who has read-only access for the app, and will receive periodic email updates. These updates include quarterly health scores and more.
```

eliangcs (approved):

packages/cli/src/oclif/commands/team/add.js
standielpls: Makes sense!

eliangcs (approved):
💯

PR #540, opened 2022-04-25
PDE-3159 feat(cli): Add collaborator to `team:remove` command

standielpls:
JIRA: https://zapierorg.atlassian.net/browse/PDE-3159

Adds a new role to be deletable with the `team:remove` command

### DIFF
--- a/packages/cli/src/oclif/commands/team/remove.js
+++ b/packages/cli/src/oclif/commands/team/remove.js
@@ -1,42 +1,21 @@
 const ZapierBaseCommand = require('../../ZapierBaseCommand');
 const { cyan } = require('colors/safe');
 const { buildFlags } = require('../../buildFlags');
-const {
-  callAPI,
-  getWritableApp,
-  listEndpointMulti,
-} = require('../../../utils/api');
+const { callAPI, getWritableApp } = require('../../../utils/api');
 const { BASE_ENDPOINT } = require('../../../constants');
-
-const roleName = (role) =>
-  role === 'collaborator'
-    ? 'admin'
-    : role === 'subscriber'
-    ? 'subscriber'
-    : 'collaborator';
+const { listTeamMembers, transformUserRole } = require('../../../utils/team');
 
 class TeamRemoveCommand extends ZapierBaseCommand {
   async perform() {
     this.startSpinner('Loading team members');
     const { admins, subscribers, limitedCollaborators } =
-      await listEndpointMulti(
-        { endpoint: 'collaborators', keyOverride: 'admins' },
-        {
-          endpoint: (app) =>
-            `${BASE_ENDPOINT}/api/platform/v3/integrations/${app.id}/subscribers`,
-          keyOverride: 'subscribers',
-        },
-        {
-          endpoint: 'limited_collaborators',
-          keyOverride: 'limitedCollaborators',
-        }
-      );
+      await listTeamMembers();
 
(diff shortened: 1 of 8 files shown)


packages/cli/src/oclif/commands/team/remove.js
```diff
@@ -8,21 +8,31 @@ const {
 } = require('../../../utils/api');
 const { BASE_ENDPOINT } = require('../../../constants');
 
-const roleName = (role) => (role === 'collaborator' ? 'admin' : 'subscriber');
+const roleName = (role) =>
+  role === 'collaborator'
```
xavdid, line 12: if we're keeping this code around long term, it may be worth combining this into a function/mechanism that all of add/remove/get can use

packages/cli/src/oclif/commands/team/remove.js
```diff
@@ -8,21 +8,31 @@ const {
...
-    const { admins, subscribers } = await listEndpointMulti(
-      { endpoint: 'collaborators', keyOverride: 'admins' },
-      {
-        endpoint: (app) =>
-          `${BASE_ENDPOINT}/api/platform/v3/integrations/${app.id}/subscribers`,
-        keyOverride: 'subscribers',
-      }
-    );
+    const { admins, subscribers, limitedCollaborators } =
+      await listEndpointMulti(
```
xavdid, line 22: seems like this is the same as the new code in `get`. can we combine them? If there's a shared function,`get` can call it and print the results. `remove` can use it to pass to `choices` below

packages/cli/src/oclif/commands/team/remove.js
```diff
@@ -50,18 +64,19 @@ class TeamRemoveCommand extends ZapierBaseCommand {
...
+        ? `/apps/${appId}/collaborators/${invitationId}`
+        : role === 'subscriber'
+        ? `${BASE_ENDPOINT}/api/platform/v3/integrations/${appId}/subscribers/${invitationId}`
+        : `/apps/${appId}/limited_collaborators`;
 
     await callAPI(url, {
       url: url.startsWith('http') ? url : undefined,
       method: 'DELETE',
-      body: { email: this.args.email },
+      body: { email: this.args.email, email_id: invitationId },
```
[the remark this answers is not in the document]
xavdid: if you can remove `email` and everything still works, go for it! I know @ETCaton was looking at how this was all called on the BE. Don't remember what the conclusion was, but this is a great opportunity to trim some cruft if we can!

xavdid (commented):
looks good! I think we should DRY up our code a bit because all of these methods are so similar, but the gist looks good.

packages/cli/src/oclif/commands/team/remove.js
standielpls: on board!

xavdid (approved):
awesome! great cleanup

eliangcs (approved):
Thanks for the cleanup! I have a suggestion on the role ordering. Besides that, I think this is good to merge. 👍

xavdid (approved):

standielpls:
note this test is very flaky: https://github.com/zapier/zapier-platform/runs/6199317207?check_suite_focus=true
Would any of the work on the PR cause the setup for this smoke test take run longer than expected?

xavdid:
no, those tests are just a little weird. as long as one passes, you can ignore.

PR #542, opened 2022-04-26
CRIT-371 perf(core): Optimize secret-scrubbing input in logger

codebycaleb:
Optimizes the logging to only scrub secrets on data that gets logged (we cap log fields at 3500 characters), maintaining JSON structure. Cuts logging time for large objects in half or more!

### DIFF
--- a/packages/core/test/misc-tools.js
+++ b/packages/core/test/misc-tools.js
@@ -98,86 +98,31 @@ describe('Tools', () => {
     });
   });
 
-  it('should check truncateData length limitations', () => {
-    const tooShort = (o) => () => dataTools.truncateData(o, 0);
-    tooShort({}).should.throw(/maxLength must be at least 39/);
-    tooShort([]).should.throw(/maxLength must be at least 38/);
-  });
+  describe('truncateData', () => {
+    const testData = require('./fixtures/truncate-test-data.json');
 
-  it('should truncate data to improve efficiency', () => {
-    const tests = [
-      {
-        value:
-          '{"authData":{"api_key":"b81ed2f2-52c7-4c0b-80c7-f5142133172c"},"inputData":{},"inputDataRaw":{},"met' +
-          'a":{"isLoadingSample":false,"isFillingDynamicDropdown":false,"isTestingAuth":false,"isPopulatingDedu' +
-          'pe":false,"limit":-1,"page":0,"isBulkRead":false,"zap":{"help":["This data structure is provided for' +
-          ' backwards compatibility,","and should not be relied upon in a Zapier integration."],"id":"subscript' +
-          'ion:17256840","link":"https://zapier.com/app/editor/subscription:17256840","live":true,"name":"A Zap' +
-          '","user":{"timezone":"America/Indiana/Indianapolis"},"trigger":{"service":{"logos":{"16x16":"https:/' +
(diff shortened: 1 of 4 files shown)


packages/core/test/misc-tools.js
```diff
@@ -98,6 +98,89 @@ describe('Tools', () => {
...
+          '.80.187.137"},{"id":11,"first_name":"Marwin","last_name":"Espie","ip_address":"222.102.146.189"},{"i' +
+          'd":12,"first_name":"Lucian","last_name":"Catley","ip_address":"251.72.199.116"},{"id":13,"first_name' +
+          '":"Diarmid","last_name":"Scothorne","ip_address":"14.244.169.107"},{"id":14,"first_name":"Ricca","la' +
+          'st_name":"Ollerearnshaw","ip_address":"132.88.63.202"},{"id":15,"first_name":"Jacquenette","last_nam' +
+          'e":"Cowcha","ip_address":"163.126.95.174"}]',
+        length: 250,
+        expected:
+          '[{"id":15,"first_name":"Jacquenette","last_name":"Cowcha","ip_address":"163.126.95.174"},{"id":14,"f' +
+          'irst_name":"Ricca","last_name":"Ollerearnshaw","ip_address":"132.88.63.202"},{"id":13,"first_name":"' +
+          'Diarmid"},"NOTE: This data has been truncated."]',
```
codebycaleb, line 150: These tests are what I felt the least confident in - I grabbed some random data from a real log, spruced it up a bit and used that for my first test; the second test is from a random fake data generator. Both don't feel very... followable? But it's tricky because there's a pretty high "minimum" size to work with for `truncateData`; even if you use the minimum length of ~40, any truncated result will be `["NOTE: This data has been truncated."]` or `{"NOTE":"This data has been truncated."}`. You've got to use a minimum of like 100 to _see_ results. And large JSON objects take a ton of space due to prettier, so I tried stringifying them to save space. 😅

xavdid: Yeah, that's tricky. I think the json strings are a little harder to follow because of the lack of syntax highlighting, but I understand the rationale. Another option could be storing a big json in a file and reading it into the tests. If you had specific design considerations in mind, you could have something like

```json
{
    "a": {"smaller": true},
    "b": {"nested": {"ok": true}},
    "c": [{"list": true},1,2, null]
}
```

You could run general tests against the large root object and specific tests against the sub objects (so array-specific things could only worry about the `c` substructure. You could also use `[bigJson]` for array based things. That would give us one big json file that doesn't blow up the test size, but also can be more easily reasoned about.

xavdid (commented):
I'll need some time to play with this, but a good starting point would be a top level comment on the function describing it's purpose and general approach. Something like:

```js
/**
 * Takes a large, JSON-serializable object `item` and removes bits until its stringified length fits in `maxLength` characters
 *
 * In general, aims to capture all top-level keys and at least 1 full object (if there's an obviously-long repetitious array). Makes X tradeoffs, doesn't handle Y, best when used w/ Z
 */
const truncateData = (item, maxLength) => {
```

Past that, it may be worth taking a pass at breaking up the function- it's pretty long as is. Happy to pair with you on that or do some of it myself. 

Either way, great job here!

xavdid (approved):
looks awesome! This is a great improvement.

PR #546, opened 2022-04-29
BREAKING CHANGE (cli) remove WB conversion functionality

xavdid:
- removed WB conversion functionality
- make `--version` required on `zapier convert`. It functionally has been (since the other API was removed), but this makes it official
- improved error handling in convert command
- removed references to converting WB apps in docs

### DIFF
--- a/packages/cli/src/oclif/commands/convert.js
+++ b/packages/cli/src/oclif/commands/convert.js
@@ -5,72 +5,37 @@ const { callAPI } = require('../../utils/api');
 const { convertApp } = require('../../utils/convert');
 const { isExistingEmptyDir } = require('../../utils/files');
 const { initApp } = require('../../utils/init');
-const { BASE_ENDPOINT } = require('../../constants');
 
 const { flags } = require('@oclif/command');
 
 class ConvertCommand extends BaseCommand {
-  generateCreateFunc(isVisual, appId, version) {
+  generateCreateFunc(appId, version) {
     return async (tempAppDir) => {
-      if (isVisual) {
-        // has info about the app, such as title
-        // has a CLI version of the actual app implementation
-        this.throwForInvalidVersion(version);
-        this.startSpinner('Downloading integration from Zapier');
-        try {
-          const [appInfo, versionInfo] = await Promise.all([
-            callAPI(`/apps/${appId}`, undefined, true),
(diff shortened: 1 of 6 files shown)


packages/cli/src/oclif/commands/convert.js
```diff
@@ -5,72 +5,37 @@ const { callAPI } = require('../../utils/api');
...
-          }
-          this.error(e);
+        return convertApp(appInfo, versionInfo.definition_override, tempAppDir);
+      } catch (e) {
+        if (e.status === 404) {
+          this.error(
+            `Visual Builder integration ${appId} @ ${version} not found. Double check the integration id and version.`
+          );
         }
+        this.error(e.json.errors[0]);
```
xavdid, line 38: this was the change I actually came in here to do, but figured I'd clean up while I was here anyway.

packages/cli/src/oclif/commands/convert.js
```diff
@@ -92,15 +56,15 @@ class ConvertCommand extends BaseCommand {
...
+    await initApp(path, this.generateCreateFunc(appId, version));
   }
 }
 
 ConvertCommand.args = [
   {
     name: 'integrationId',
     required: true,
-    description: `To get the integration/app ID, go to "${BASE_ENDPOINT}/app/developer", click on an integration, and copy the number directly after "/app/" in the URL.`,
+    description: `To get the integration/app ID, go to "https://developer.zapier.com//app/developer", click on an integration, and copy the number directly after "/app/" in the URL.`,
```
eliangcs, line 67:
```suggestion
    description: `To get the integration/app ID, go to "https://developer.zapier.com", click on an integration, and copy the number directly after "/app/" in the URL.`,
```

packages/cli/src/oclif/commands/convert.js
eliangcs: Did you make this change because some developer was seeing an obscure error? If so, can you provide a before/after?

eliangcs (commented):
I have a suggested change on the typo.

Another question: do we want to keep this PR open until we're ready for a major release?

xavdid:
Thanks for catching the typo!

And yeah, I'm good to sit on it. The next major version will be node 16 support whenever AWS releases support. should be any day now!

eliangcs (approved):

PR #548, opened 2022-05-24
PDE-3210 fix(legacy-scripting-runner): fix `logResponse` import

eliangcs:
Makes a `logResponse` import backward compatible with core 12.0.2 and 12.0.3.


xavdid (approved):
great docs!

PR #556, opened 2022-06-27
PDE-2085 chore(cli): bump @oclif/plugin-help to ^3.2.14 to patch vulnerability in sub-dependency

### LINKED ISSUE #555: Command Injection on sub-dependency lodash.template via @oclif/plugin-help
tldr; please update `@oclif/plugin-help` to `3.2.14` to fix a vulnerability in [`zapier-platform-cli`](https://github.com/zapier/zapier-platform/blob/master/packages/cli/package.json)

---

`lodash.template` package has a security vulnerability. See here https://security.snyk.io/vuln/SNYK-JS-LODASHTEMPLATE-1088054. The CVE rating is HIGH 7.2.

`@oclif/plugin-help` [version should be updated](https://www.npmjs.com/package/@oclif/plugin-help). I believe [3.2.14](https://www.npmjs.com/package/@oclif/plugin-help/v/3.2.14) would suffice.

Please consider `^` syntax so that your dependents can update patch versions without updates required to the repo

mrjackdavis:
fixes #555

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -41,7 +41,7 @@
     "@oclif/command": "1.8.0",
     "@oclif/config": "1.17.0",
     "@oclif/plugin-autocomplete": "0.3.0",
-    "@oclif/plugin-help": "3.2.2",
+    "@oclif/plugin-help": "^3.2.14",
     "@oclif/plugin-not-found": "1.2.4",
     "adm-zip": "0.5.5",
     "archiver": "5.3.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -41,7 +41,7 @@
     "@oclif/command": "1.8.0",
     "@oclif/config": "1.17.0",
     "@oclif/plugin-autocomplete": "0.3.0",
-    "@oclif/plugin-help": "^3.2.14",
+    "@oclif/plugin-help": "~3.2.14",
     "@oclif/plugin-not-found": "1.2.4",
     "adm-zip": "0.5.5",
     "archiver": "5.3.0",


packages/cli/package.json
```diff
@@ -41,7 +41,7 @@
     "@oclif/command": "1.8.0",
     "@oclif/config": "1.17.0",
     "@oclif/plugin-autocomplete": "0.3.0",
-    "@oclif/plugin-help": "3.2.2",
+    "@oclif/plugin-help": "~3.2.14",
```
eliangcs, line 44: For this repo, we always pin the exact version of a dependency, to be extra safe that nothing breaks due to accidental version bump.

```suggestion
    "@oclif/plugin-help": "3.2.14",
```

eliangcs (changes requested):
Thanks for fixing! I understand conventionally you'd use a `^` or `~` to auto upgrade the dependencies. But for the packages in this repo, it's more important not to break, so we always use the exact versions to make sure the installation is as deterministic as possible.

mrjackdavis:
I patched a repo at work with this change, but unfortunately a bunch of other stuff breaks. For some reason that dependency patch version bump has a bunch of breaking changes. I'll swing back here and update the PR when I get time

AlonNavon:
@mrjackdavis 
Hey Jack, I know this is an old issue, but it's the kind of pain [we](https://seal.security) exist to solve. You want to fix CVE-2021-23337 in `lodash.template`, but it's a transitive dependency so you have to upgrade `@oclif/plugin-help`, but the patch version contains breaking changes so a simple version bump won't work. Oldest story in the (open-source) world.

At Seal Security we backport security patches and release them freely to the community. So instead of upgrading `@oclif/plugin-help` you can just use npm's override feature and override the existing `lodash.template` version with our [own](https://github.com/seal-community/patches/tree/main/npm/lodash.template/4.5.0/sp1), which is available for free on our artifact server. Hope this will be able to solve your issue (and future similar ones).

If you have any questions, feel free to contact us.

rnegron:
Fixed in #739. Thanks for flagging this issue!

PR #557, opened 2022-06-27
[SUPPORTESC-368] docs(cli): Fix unqualified use of `username` in connectionLabel for Basic auth

armchairlinguist:
It's possible we previously supported the use of unqualified references like `username` for the connection label, but in the present platform it seems to require either `bundle.authData` or `bundle.inputData` to figure out what to use.

### DIFF
--- a/packages/cli/snippets/basic-auth.js
+++ b/packages/cli/snippets/basic-auth.js
@@ -4,7 +4,7 @@ const authentication = {
   test: {
     url: 'https://example.com/api/accounts/me.json',
   },
-  connectionLabel: '{{username}}', // Can also be a function, check digest auth below for an example
+  connectionLabel: '{{bundle.authData.username}}', // Can also be a function, check Digest auth below for an example
   // you can provide additional fields, but we'll provide `username`/`password` automatically
 };
 
--- a/docs/index.html
+++ b/docs/index.html
@@ -1168,7 +1168,7 @@ zapier convert 1234 --version 1.0.1 my-app
   <span class="hljs-attr">test</span>: {
     <span class="hljs-attr">url</span>: <span class="hljs-string">&apos;https://example.com/api/accounts/me.json&apos;</span>,
   },
-  <span class="hljs-attr">connectionLabel</span>: <span class="hljs-string">&apos;{{username}}&apos;</span>, <span class="hljs-comment">// Can also be a function, check digest auth below for an example</span>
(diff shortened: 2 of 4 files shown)


packages/cli/snippets/basic-auth.js
```diff
@@ -4,7 +4,7 @@ const authentication = {
   test: {
     url: 'https://example.com/api/accounts/me.json',
   },
-  connectionLabel: '{{username}}', // Can also be a function, check digest auth below for an example
+  connectionLabel: '{{bundle.authData.username}}', // Can also be a function, check Digest auth below for an example
```
armchairlinguist, line 7: This makes the line a little long - I can adjust if necessary, but shortening the comment seemed challenging.

eliangcs (commented):
@armchairlinguist `{{username}}` should still work on the present platform. If the developer wants to be explicit, they can also use `{{bundle.inputData.username}}` or `{{bundle.authData.username}}`.

armchairlinguist:
Whoops - it looks like it did actually work, but as happens sometimes when testing I didn't notice that it was working. 🤦🏻‍♀️ 

However, I did get odd results when comparing CLI to Visual Builder output - CLI seems to resolve to `bundle.authData.username`, while VB resolves to `bundle.inputData.username`. So it seems advisable to use the fully qualified version?

eliangcs:
@armchairlinguist Visual Builder and CLI should behave the same. When creating a connected account, Zapier calls the app's test function to get the test result. Say the test function returns a response of `{"username":"abc", "id":1234}`. Behind the scenes, Zapier constitutes an object like this:

```json
{
  "username": "abc",
  "id": 1234,
  "bundle": {
    "authData": {
      "username": "abc",
      "id": 1234,
    },
    "inputData": {
      "username": "abc",
      "id": 1234,
    },
  }
}
```

Notice how the test result gets copied to `bundle.authData` and `bundle.inputData`. This object then is used to render the connection label template. So templates like `{{username}}`, `{{bundle.authData.username}}, and `{{bundle.inputData.username}}` all work.

When you say "it resolves to bundle.authData.username", do you mean that the connection label automatically changes from `{{username}}` to `{{bundle.authData.username}}`? I'm not aware of this behavior. Can you clarify? Thanks!

armchairlinguist:
What I found is that in the Visual Builder, the result I got for `username` was from the test request, and in the CLI, it was from the username input for the auth credentials.

It seems like it's not super-important to change this right now, since it does give some sensible result in either case for Basic auth. I'll close this PR and we can discuss separately. 👍🏻

PR #562, opened 2022-07-08
PDE-3251 fix(core): abort logger connection at the end of invocation

eliangcs:
Fixes the issue where a Lambda invocation can take ~1 minute before a `socket hang up` error with the log server shows up.

You can find the steps to reproduce the error on [Slack](https://zapier.slack.com/archives/C5Z9BP4U9/p1657183594848969?thread_ts=1656322076.695299&cid=C5Z9BP4U9).

I’m not sure under what situations the log server would stop responding to an open connection on production. (Maybe the log server sometimes crashes?) But we should at least make the core more resilient and not affected by a hanging connection.

The idea is to limit the wait time for the log server to respond. We use node-fetch to make requests to the log server. And to abort a node-fetch request, we can use AbortController. See node-fetch's [README](https://github.com/node-fetch/node-fetch/tree/v2.6.7#request-cancellation-with-abortsignal) for example. Since AbortController only became natively supported on Node.js 15+, I'm using the [node-abort-controller](https://github.com/southpolesteve/node-abort-controller) package.

### DIFF
--- a/packages/core/src/tools/create-logger.js
+++ b/packages/core/src/tools/create-logger.js
@@ -30,6 +30,8 @@ const {
 // than the limit (16 MB) on the server side.
 const LOG_STREAM_BYTES_LIMIT = 15 * 1024 * 1024;
 
+const DEFAULT_LOGGER_TIMEOUT = 200;
+
 const sleep = promisify(setTimeout);
 
 const isUrl = (url) => {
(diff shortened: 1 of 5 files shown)


zapzap:
I'm no Node expert so forgive me if I'm totally wrong here, but would we want to pass a value for the `timeoutToAbort` argument so that we don't kill off the connection _immediately_ but give the server at least some time to close it for us?

eliangcs:
@zapzap (not sure who you are) that makes sense. In the latest commits, apart from fixing a test, I've made the time limit default to 200 ms, so the server has a chance to close the connection for us.

packages/core/src/tools/create-logger.js
```diff
@@ -179,15 +186,20 @@ class LogStream extends Transform {
...
+        };
+      }
+
       // Swallow logging errors. This will show up in AWS logs at least.
-      console.error(
-        'Error making log request:',
-        err,
-        'http options:',
-        httpOptions
-      );
```
eliangcs, line 190: I'm removing `httpOptions` from the log because there were credentials in the AWS logs.

PR #564, opened 2022-08-08
[SUPPORTENG-389] docs(cli): Provide additional information about the connection label in the doc text

armchairlinguist:
We had a few examples of setting the connection label in the code snippets, but there wasn't previously any text discussing it. I've added some text to provide information on how it functions inside Zapier, as well as the difference between using a basic variable reference inside a string and a function to set it.


eliangcs (approved):
Love the addition! Thanks!

PR #566, opened 2022-08-23
PDE-3318 fix(legacy-scripting-runner): `z.request` in legacy script isn't logged

eliangcs:
Fixes a regression caused by #509, where calling `z.request` in the legacy script does not (but should) produce an HTTP log.

For an HTTP request to be logged, it must call the patched [`http(s).request`](https://github.com/zapier/zapier-platform/blob/b580d5f781d4b6e1aeff49e906bb9f36f6872589/packages/core/src/tools/create-http-patch.js#L21) function. The patch is done [at the Lambda handler initialization time](https://github.com/zapier/zapier-platform/blob/b580d5f781d4b6e1aeff49e906bb9f36f6872589/packages/core/src/tools/create-lambda-handler.js#L202-L204).

But since #509 adopted [synckit](https://github.com/rx-ts/synckit) to implement the [sync version of `z.request`](https://platform.zapier.com/legacy/scripting#making-outbound-requests-zrequest) in the legacy script, the HTTP request for the sync `z.request` has been happening in a separate thread, where `http(s).request` isn't patched like the main thread. This is why HTTP logs haven't been showing up for a lot of apps converted from Web Builder.

To fix, we let legacy-scripting-runner write the HTTP log explicitly when the sync `z.request` is used.

### DIFF
--- a/packages/legacy-scripting-runner/CHANGELOG.md
+++ b/packages/legacy-scripting-runner/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.8.11
+
+- :bug: sync `z.request` doesn't produce an HTTP log ([#566](https://github.com/zapier/zapier-platform/pull/566))
+
 ## 3.8.10
 
 - :bug: Add support for `StopRequestException` in more methods ([#561](https://github.com/zapier/zapier-platform/pull/561))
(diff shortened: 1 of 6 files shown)


standielpls (approved):
makes sense, nice solution!

szchenghuang (approved):

PR #569, opened 2022-08-26
[SUPPORTENG-406] Correcting ThrottledError example to include z.errors

robertscl:
I updated the ThrottledError example to call ThrottledError on z.errors. The current example does not use z.errors (`throw new ThrottledError()`).

current example

ThrottledError is listed under z.errors in [the docs](https://github.com/zapier/zapier-platform/blob/master/packages/cli/README.md#zerrors), so this change makes this example consistent with the usage of other errors.

In my own testing, ThrottledError only gets thrown when z.errors is used.


eliangcs (approved):
Nice catch, thanks!

PR #573, opened 2022-09-13
[PDE-3393] Log warnings at build time during validation

kola-er:
build(cli):  Log warnings at build time during validation


eliangcs (approved):
Looks good! When testing, I found we need one change on the backend for the L004 check to show up here. Don't worry about it - I'll make that change.

For this PR, I have one suggestion on printing an additional line.

eliangcs (approved):

kola-er:
@eliangcs Thanks for the review. I can't merge, please help with it.

PR #574, opened 2022-09-15
[PDE-3395] feat(cli): Add `--account` flag to migrate command

robwa10:
## Description  
This PR adds an `--account` flag to `zapier migrate` command and alters the `--user` flag behaviour.  

The `--user` flag will now only migrate a user's zap steps, in any account they are a part of, using the specified integration if the both the Zap and the integrations authentication used are Private for that user.  

The `--account` flag allows developers to retain the current behaviour of migrating a user's zap steps, in any account they are a part of, using the specified integration regardless of whether the Zap is Private.  

I've also updated the docs to clarify the behaviour expected from each flag.


packages/cli/src/oclif/commands/migrate.js
```diff
@@ -50,12 +59,15 @@ class MigrateCommand extends BaseCommand {
         name: 'migrate',
         from_version: fromVersion,
         to_version: toVersion,
-        email: user,
+        email: user || account,
+        accountType: flagType,
```
eliangcs, line 63: I think you need to use `email_type` to match the name on the backend.

```suggestion
        email_type: flagType,
```

eliangcs (commented):
Looks good at first glance. I like the doc improvement!

Note that we should only merge it after the backend is ready. I'll do a more thorough review and test then.

Question: Do you want to release it publicly or internally? Both are good. But if you're releasing it internally, we should merge it to a separate branch instead of `master`.

packages/cli/src/oclif/commands/migrate.js
benzapier: I'd lean towards releasing it publicly, so that the `user` behavior only does the user (rather than the current behavior, where `user` does the accounts)

docs/cli.html
```diff
@@ -790,18 +790,27 @@ <h2 id="migrate">migrate</h2>
...
+</li>
+</ul><p>Alternatively, you can pass the <code>--account</code> flag, (i.e. <code>zapier migrate 1.0.0 1.0.1 --account=account@example.com</code>). This will migrate all the user&apos;s Zap steps in any account the user is part of, regardless of Zap owner.</p><p><strong>The <code>--account</code> flag should be used cautiously as it can break shared Zaps for other users in Team or Company accounts.</strong>  </p><p>You cannot pass both <code>PERCENT</code> and <code>--user</code> or <code>--account</code>.</p><p>You cannot pass both <code>--user</code> and <code>--account</code>.</p><p><strong>Arguments</strong></p><ul>
 <li>(required) <code>fromVersion</code> | The version FROM which to migrate users.</li>
 <li>(required) <code>toVersion</code> | The version TO which to migrate users.</li>
 <li><code>percent</code> | Percentage (between 1 and 100) of users to migrate.</li>
 </ul><p><strong>Flags</strong></p><ul>
-<li><code>--user</code> | Migrate only this user</li>
+<li><code>--user</code> | Migrate a users private Zaps using the integration</li>
+<li><code>--account</code> | Migrate all a users Zaps using the integration, incl. shared Zaps</li>
```
benzapier, line 807: Seems like this needs a bit more info about the effect of the `account` flag - maybe something like `Migrates all of the Zaps (for all users) within each of user's accounts`

docs/cli.md
```diff
@@ -340,35 +340,51 @@ This won't show logs from running locally with `zapier test`, since those never
...
-You can migrate a single user by using `--user` (i.e. `zapier migrate 1.0.0 1.0.1 --user=user@example.com`).
+You can migrate a specific user's Zaps by using `--user` (i.e. `zapier migrate 1.0.0 1.0.1 --user=user@example.com`). This will migrate Zaps in any account the user is part of where the following criteria is met.  
+
+  - The Zap is owned by the user.  
+
+  - The Zap is not shared.  
+
+  - The integration auth used is not shared.  
+
+Alternatively, you can pass the `--account` flag, (i.e. `zapier migrate 1.0.0 1.0.1 --account=account@example.com`). This will migrate all the user's Zap steps in any account the user is part of, regardless of Zap owner.
```
benzapier, line 365: It's probably clearer and more accurate saying something like `This will migrate all users' Zaps within all accounts for which the specified user is a member`

docs/cli.md
```diff
@@ -340,35 +340,51 @@ This won't show logs from running locally with `zapier test`, since those never
...
 
 **Arguments**
 * (required) `fromVersion` | The version FROM which to migrate users.
 * (required) `toVersion` | The version TO which to migrate users.
 * `percent` | Percentage (between 1 and 100) of users to migrate.
 
 **Flags**
-* `--user` | Migrate only this user
+* `--user` | Migrate a users private Zaps using the integration
+* `--account` | Migrate all a users Zaps using the integration, incl. shared Zaps
```
benzapier, line 380: Similar change above - something like `This will migrate all users' Zaps within all accounts for which the specified user is a member`

packages/cli/docs/cli.md
```diff
@@ -340,35 +340,51 @@ This won't show logs from running locally with `zapier test`, since those never
...
-You can migrate a single user by using `--user` (i.e. `zapier migrate 1.0.0 1.0.1 --user=user@example.com`).
+You can migrate a specific user's Zaps by using `--user` (i.e. `zapier migrate 1.0.0 1.0.1 --user=user@example.com`). This will migrate Zaps in any account the user is part of where the following criteria is met.  
+
+  - The Zap is owned by the user.  
+
+  - The Zap is not shared.  
+
+  - The integration auth used is not shared.  
+
+Alternatively, you can pass the `--account` flag, (i.e. `zapier migrate 1.0.0 1.0.1 --account=account@example.com`). This will migrate all the user's Zap steps in any account the user is part of, regardless of Zap owner.
```
benzapier, line 365: Similar change above ...

packages/cli/docs/cli.md
```diff
@@ -340,35 +340,51 @@ This won't show logs from running locally with `zapier test`, since those never
...
 
 **Arguments**
 * (required) `fromVersion` | The version FROM which to migrate users.
 * (required) `toVersion` | The version TO which to migrate users.
 * `percent` | Percentage (between 1 and 100) of users to migrate.
 
 **Flags**
-* `--user` | Migrate only this user
+* `--user` | Migrate a users private Zaps using the integration
+* `--account` | Migrate all a users Zaps using the integration, incl. shared Zaps
```
benzapier, line 380: Similar change above ...

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -83,7 +95,11 @@ class MigrateCommand extends BaseCommand {
 MigrateCommand.flags = buildFlags({
   commandFlags: {
     user: flags.string({
-      description: 'Migrate only this user',
+      description: 'Migrate a users private Zaps using the integration',
+    }),
+    account: flags.string({
+      description:
+        'Migrate all a users Zaps using the integration, incl. shared Zaps',
```
benzapier, line 102: Similar change above ...

packages/cli/src/oclif/commands/migrate.js
robwa10: Done, didn't realise that was the key we were using for the backend. 👍

eliangcs (approved):
👍

standielpls:
This looks good, I've tested the functionality and the messaging seems clear. will merge to `master` and cut a release.

PR #576, opened 2022-09-20
[PDE-3262] Add skip-changelog flag to zapier promote CLI command

gregilo:
This flag suppresses prompts asking to confirm a corresponding changelog entry when running zapier promote. This is primarily for CI purposes and will still show warnings/info related to the changelog during zapier promote runs. I utilized oclif's [built-in boolean flag type](https://oclif.io/docs/flags) for this flag.

Usage: `zapier promote 1.0.0 --skip-changelog`

### DIFF
--- a/packages/cli/src/oclif/commands/promote.js
+++ b/packages/cli/src/oclif/commands/promote.js
@@ -29,7 +29,7 @@ class PromoteCommand extends BaseCommand {
     const app = await this.getWritableApp();
 
     const version = this.args.version;
-    const skipChangelog = 'skip-changelog' in this.flags;
+    const assumeYes = 'yes' in this.flags;
 
     let shouldContinue;
     const changelog = await getVersionChangelog(version);
@@ -38,7 +38,7 @@ class PromoteCommand extends BaseCommand {
       this.log(`\n---\n${changelog}\n---\n`);
 
       shouldContinue =
-        skipChangelog ||
+        assumeYes ||
         (await this.confirm(
           'Would you like to continue promoting with this changelog?'
         ));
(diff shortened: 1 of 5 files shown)


packages/cli/src/oclif/commands/promote.js
```diff
@@ -117,7 +123,14 @@ class PromoteCommand extends BaseCommand {
   }
 }
 
-PromoteCommand.flags = buildFlags();
+PromoteCommand.flags = buildFlags({
+  commandFlags: {
+    'skip-changelog': flags.boolean({
+      description:
+        'Suppress prompts asking to confirm a corresponding changelog entry for the version you are attempting to promote. ',
```
eliangcs, line 130: `skip-changelog` sounds like we're skipping changelog, even if there's one, while the actual behavior is to automatically answer "yes" to any prompts.jj

I suggest we name this flag `--yes` and `-y`, like how [`apt-get`](https://linux.die.net/man/8/apt-get) does it:

```suggestion
    yes: flags.boolean({
      char: 'y',
      description:
        'Automatically answer "yes" to any prompts. Useful if you want to avoid interactive prompts to run this command in CI.',
    }),
```

eliangcs (commented):
Functionality wise, it works great! But I suggest we change the flag from `--skip-changelog` to `--yes` to be more explicit about what it does. Let me know what you think. Thanks!

packages/cli/src/oclif/commands/promote.js
gregilo: @eliangcs I'm cool with approach 👍

eliangcs (approved):
👍

PR #580, opened 2022-10-07
[PDE-3458] feat(schema): Add upsert support

rnegron:
# Description

This PR adds "search and update" (or "upsert") action support for integrations developed via the Zapier Platform CLI.

The difference between this new "search and update" vs. the existing "search or update" is that if an item was found in the search step,  "search or update" would just return that unchanged item. Using "search and update" on the other hand, that found item would be updated with the specified data. The new "search and update" action would also raise an error if more than 1 item was found, whereas the existing "search or update" would just return the first match.

# Changelog

## Add
- Added `/SearchAndCreatesSchema`, which is an alias to `/SearchOrCreatesSchema` and only changes the `id` and `description` properties.
- Added search-and-create style example for `/SearchOrCreateSchema` and `/SearchOrCreatesSchema`
- Added functional constraints to `searchOrCreateKeys` that validates the new fields in `/SearchOrCreateSchema`
- Added functional constraint that verifies that only one of `searchOrCreates` and `searchAndCreates` is used at once
- Added tests for the `searchOrCreateKeys` functional constraint, including tests for the new constraints.

## Change
- Update `/AppSchema` to add `searchAndCreates` key
- Update the `/SearchOrCreateSchema` to include the following fields
  - `update`
  - `updateInputFromSearchOutput`
  - `searchUniqueInputToOutputConstraint` 
- Update `schema.md` documentation
- Updated `exported-schema.json`


packages/schema/lib/schemas/SearchOrCreateSchema.js
```diff
@@ -30,6 +32,21 @@ module.exports = makeSchema(
         description: 'The key of the create that powers this search-or-create',
         $ref: KeySchema.id,
       },
+      update: {
+        description:
+          'The key of the create that will be used if a search succeeds; having a value here converts this search-or-create into a search-and-create',
```
eliangcs, line 37: Let's add "EXPERIMENTAL" to the help text to discourage external partners from using it until it gets stable.

```suggestion
          'EXPERIMENTAL: The key of the create that will be used if a search succeeds; having a value here converts this search-or-create into a search-and-create',
```

Same goes for `updateInputFromSearchOutput` and `searchUniqueInputToOutputConstraint`.

packages/schema/lib/functional-constraints/searchOrCreateKeys.js
```diff
@@ -59,6 +122,149 @@ const validateSearchOrCreateKeys = (definition) => {
...
+    // Confirm searchOrCreate.updateInputFromSearchOutput contains objects with:
+    // keys existing in creates[update].operation.inputFields.key
+    // values existing in searches[search].operation.(outputFields.key|sample keys), if they are defined
+    if (searchOrCreateDef.updateInputFromSearchOutput && updateKey) {
+      // Note that _.each({key: value}) provides the following callback method signature: (value, key) => {}
+      _.each(
+        searchOrCreateDef.updateInputFromSearchOutput,
+        (searchOutputField, updateInputField) => {
+          // Confirm searchOrCreate.updateInputFromSearchOutput's key exists in creates[update].operation.inputFields.key
+          if (_.isEmpty(updateInputKeys)) {
```
eliangcs, line 174: `_.isEmpty(updateInputKeys)` is constant during the loop. So the code would be tidier and more efficient if we move it out of the loop.

Also, it'd be clearer to include `${updateKey}` in the error message.

```javascript
if (searchOrCreateDef.updateInputFromSearchOutput && updateKey) {
  const updateOptionHint = _.isEmpty(updateInputKeys) ? '(no "key" found in inputFields)' : `(options: 
 ${updateInputKeys})`;

  const searchOptionHint = `(options: ${[...allSearchOutputKeys].join(',')})`

  Object.entries(searchOrCreateDef.updateInputFromSearchOutput).forEach((updateInputField, searchOutputField) => {
    if (!updateInputKeys.includes(updateInputField)) {
      errors.push(
        new jsonschema.ValidationError(
          `must match a "key" from a creates.${updateKey}operation.inputFields ${updateOptionHint}`,
          searchOrCreateDef,
          '/SearchOrCreateSchema',
          `instance.searchOrCreates.${key}.updateInputFromSearchOutput`,
          'invalidKey'
        )
      );
    }
    
    if ((hasSearchOutputFields || hasSearchOutputSample) && !allSearchOutputKeys.has(searchOutputField)) {
      errors.push(
        new jsonschema.ValidationError(
          `must match a "key" from searches.${searchKey}.operation.(outputFields|sample) ${searchOptionHint}`,
          searchOrCreateDef,
          '/SearchOrCreateSchema',
          `instance.searchOrCreates.${key}.updateInputFromSearchOutput`,
          'invalidKey'
        )
      );
    }
  });
}
```

Same goes for the `searchUniqueInputToOutputConstraint` code block below.

eliangcs (commented):
I left some suggestions, mostly about tidying up the code.

There's a scenario we need to fix: When the app definition only has `searchAndCreates` and **not** `searchOrCreates`, the functional constraints aren't working. For example:

```javascript
const App = {
  // ...
  searchAndCreates: {
    'find_recipe': {
      key: 'find_recipe',
      display: { /* ... */ },
      search: 'find_recipe',
      create: 'add_recipe',
      update: 'BAD_KEY',  // <- should error
    },
  },
};
```

... only gives me one error saying `should not be used at the same time as its alias, searchAndCreates`. It misses another error saying `must match a "key" from a create` for `App.searchAndCreates.recipe.update`.

packages/schema/lib/functional-constraints/searchOrCreateKeys.js
```diff
@@ -3,6 +3,45 @@
...
+};
+
+const getSearchOutputSampleKeys = ({ definition, searchKey }) => {
+  const searchOutputSampleFields = _.get(
+    definition.searches,
+    `${searchKey}.operation.sample`,
+    {}
+  );
+
+  return _.keys(searchOutputSampleFields);
```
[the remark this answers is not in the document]
rnegron: Agreed!

rnegron:
@eliangcs The `should not be used at the same time as its alias, searchAndCreates` error was due to the `compileApp` function from `core/src/tools/schema` always adding a `searchOrCreates` key to `appRaw`, which then caused the alias validations to fail. I modified the `core` package to account for this by moving the `searchAndCreates` contents to `searchOrCreates`.

rnegron:
After further testing, I realized that the CLI should actually be sending a `searchAndCreates` key to the backend. So I further modified the `core` package to account for this by adding some conditionals around having a `searchAndCreates` key present in the definition.

The main issue I was running into was: When running `zapier validate`, the `compileApp` function runs before any validation takes place. So at that point, it may be possible to have both `searchOrCreates` and `searchAndCreates` defined in `appRaw`.

packages/schema/lib/functional-constraints/searchOrCreateKeys.js
```diff
@@ -3,20 +3,75 @@
 const _ = require('lodash');
 const jsonschema = require('jsonschema');
 
+const getFieldKeys = (definition, path) => {
+  const fields = _.get(definition, path, []);
+  return fields.map((field) => field.key);
```
eliangcs, line 8: I forgot an `inputField` or `outputField` [can be a function](https://github.com/zapier/zapier-platform/blob/e471e2bd8fdc7733cf9ea5a9d59053746def483d/packages/schema/docs/build/schema.md#fieldorfunctionschema) to support [custom fields](https://github.com/zapier/zapier-platform/blob/e471e2bd8fdc7733cf9ea5a9d59053746def483d/packages/cli/README.md#customdynamic-fields). So we need to filter those fields that are functions and don't have a `key`.

```suggestion
  return fields.map((field) => field.key).filter((k) => k);
```

eliangcs (commented):
This ends up being much more complicated than I expected. Thank you for keeping up good work so far!

I think your edits to the core package are good. 

I like the test coverage. I suggest we add one more regarding `searchAndCreates`:
- [ ] an app definition contains only `searchAndCreates` but has an invalid key in it (should error with the right message)

Once all the suggestions are addressed, I'll do one more round of manual testing. If the test looks good, this should be ready to merge.

packages/core/src/tools/schema.js
```diff
@@ -54,21 +55,26 @@ const convertResourceDos = (appRaw) => {
     }
 
     if (search && create && isVisible(search) && isVisible(create)) {
-      const searchOrCreate = {
-        // key: `${resource.key}SearchOrCreate`,
+      // searchAndCreates is an alias for searchOrCreates. Schema validation makes sure only one of them is defined.
+      const searchCreates = appRaw.searchAndCreates
+        ? appRaw.searchAndCreates
+        : appRaw.searchOrCreates;
```
eliangcs, line 61: There are two unused variables, `searchOrCreates` and `searchAndCreates`, in the beginning of this function. I think they probably should be used here instead?

packages/core/src/tools/schema.js
```diff
@@ -54,21 +55,26 @@ const convertResourceDos = (appRaw) => {
...
         },
         search: search.key,
         create: create.key,
       };
-      searchOrCreates[searchOrCreate.key] = searchOrCreate;
     }
   });
 
-  return { triggers, searches, creates, searchOrCreates };
+  return { triggers, searches, creates, searchOrCreates, searchAndCreates };
```
eliangcs, line 77: We may only want to add `searchOrCreates` _or_ `searchAndCreates`, not both. The backend can handle when both are defined, but I'd like it to enforce schema on write.

eliangcs (commented):
The logic in `convertResourceDos` may still need fixing. Not sure why tests aren't running on GitHub, but some test cases in `core/test/tools/schema.js` are failing on my local environment. I'll go ahead and fix it in this PR.

PR #582, opened 2022-10-14
[SUPPORTENG-385] Clarifying how fields accept file types

justin-lee-zapier:
# Description

I added some additional language to how Zapier processes `file` types in fields. It looks like our [user help docs](https://zapier.com/help/create/basics/different-field-types-in-zaps#step-4) provide some more additional details than our developer docs. So this will bring the developer docs in line with what we have that is public facing.

# Changelog

## Add
Additional information on how we process strings and that we make a GET request when we detect a URL.


eliangcs (approved):
@justin-lee-zapier the files under `docs/build/` are auto generated. What you'd want to edit instead is the `FieldSchema.js` file, for which I've gone ahead and pushed a commit to this PR. Please merge it if the change also looks good to you. Thanks!

justin-lee-zapier:
Ah! Okay, thanks for making that change @eliangcs . This looks good, merging now!

PR #589, opened 2022-11-09
[SUPPORTENG-482] Doc update: Update OAuth2 snippet to note key limitations

robertscl:
Here, I've updated the [OAuth2](https://github.com/zapier/zapier-platform/blob/master/packages/cli/README.md#oauth2) snippet to note that `client_id` and `client_secret` can't be used as keys for form input fields.

In testing, documented in the Slack threads linked in the accompanying Jira ticket, it appears `client_id` in particular conflicts with the similarly named environment variable, causing issues in the token request. This can be avoided by refraining from using these key names in input form fields in the OAuth2 config.

Note: 

- I see package-lock.json got changed here. To be transparent, I don't know if that's expected. Please let me know if I need to fix that and resubmit.


packages/cli/README.md
```diff
@@ -838,8 +838,10 @@ const authentication = {
   fields: [
     { key: 'subdomain', type: 'string', required: true, default: 'app' },
     // For OAuth2 we store `access_token` and `refresh_token` automatically
-    // in `bundle.authData` for future use. If you need to save/use something
-    // that the user shouldn't need to type/choose, add a "computed" field, like:
+    // in `bundle.authData` for future use. Note, `client_id` and `client_secret` 
+    // are reserved and cannot be used as keys for input form fields. 
+    // If you need to save/use something that the user shouldn't need to type/choose, 
+    // add a "computed" field, like:
```
eliangcs, line 844: Can you move this change to `README-source-md` and let the pre-commit hook auto generate `README.md` for us? Thanks!

packages/cli/package-lock.json
```diff
@@ -0,0 +1,13 @@
...
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "yarn": {
+      "version": "1.22.19",
+      "resolved": "https://registry.npmjs.org/yarn/-/yarn-1.22.19.tgz",
+      "integrity": "sha512-/0V5q0WbslqnwP91tirOvldvYISzaqhClxzyUKXYxs07yUILIs5jx/k6CFe8bvKSkds5w+eiOqta39Wk3WxdcQ=="
+    }
+  }
+}
```
eliangcs, line 13: `package-lock.json` should not be in this change. We should undo the change here.

robertscl:
@eliangcs, thank you for the suggestions! I believe I've made the necessary changes, but please let me know if anything is still incorrect.

eliangcs (approved):
Looks good. Thanks!

PR #590, opened 2022-11-22
docs(cli): More strongly recommend use of z.dehydrateFile for files

armchairlinguist:
We've seen a few cases where the use of plain `z.dehydrate` for file data has created file retrieval issues, so I've updated the section on File Dehydration to be a bit stronger on recommending the use of the dedicated function.

I also found we were listing quite a few things as "New in" which are no longer new, and changed that to "Added in". (Changes in versions 11+ are still listed as new.)


eliangcs (approved):
Looks good!

codebycaleb (approved):

PR #591, opened 2022-11-22
[SUPPORTENG-520] docs(cli): Update logging info

armchairlinguist:
We had some inconsistent information on types of logs and what's logged automatically; this PR removes the inconsistency and does some other light cleanup.


eliangcs (approved):
Looks good!

PR #610, opened 2023-01-23
PDE-3731 feat(cli): Add getAppRegistrationFieldChoices hook

gregilo:
This PR introduces a new hook `getAppRegistrationFieldChoices` that triggers during oclif's `init` hook (see [docs](https://oclif.io/docs/hooks)). The goal of this hook is to get enum field choices from `/api/platform/cli/apps/fields-choices` and provide them to `RegisterCommand` for use with forthcoming flags/prompts. It should not run on other commands.

This PR has no noticeable effect on current functionality.

### DIFF
--- a/packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
+++ b/packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
@@ -1,0 +1,17 @@
+const { callAPI } = require('../../utils/api');
+
+module.exports = async function (options) {
+  // We only need to run this for the register command
+  if (!options || !options.id || options.id !== 'register') {
+    return null;
+  }
+
+  const enumFieldChoices = {};
+  const formFields = await callAPI('/apps/fields-choices');
+
+  for (const fieldName of ['intention', 'role', 'app_category']) {
+    enumFieldChoices[fieldName] = formFields[fieldName];
+  }
+
+  this.config.enumFieldChoices = enumFieldChoices;
+};
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -99,7 +99,8 @@
         "./src/oclif/hooks/deprecated",
         "./src/oclif/hooks/updateNotifier",
         "./src/oclif/hooks/checkValidNodeVersion",
-        "./src/oclif/hooks/renderMarkdownHelp"
+        "./src/oclif/hooks/renderMarkdownHelp",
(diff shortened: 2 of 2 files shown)


packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
```diff
@@ -0,0 +1,17 @@
+const { callAPI } = require('../../utils/api');
+
+module.exports = async function (options) {
+  // We only need to run this for the register command
+  if (!options || !options.id || options.id !== 'register') {
+    return null;
```
kola-er, line 6: Good provision to return early for other commands/use-cases except for `register`.

docs/cli.html
```diff
@@ -381,7 +381,7 @@ <h2 id="convert">convert</h2>
       <blockquote>
 <p>Convert a Visual Builder integration to a CLI integration.</p>
 </blockquote><p><strong>Usage</strong>: <code>zapier convert INTEGRATIONID PATH</code></p><p>The resulting CLI integraiton will be identical to its Visual Builder version and ready to push and use immediately!</p><p>If you re-run this command on an existing directory it will leave existing files alone and not clobber them.</p><p>You&apos;ll need to do a <code>zapier push</code> before the new version is visible in the editor, but otherwise you&apos;re good to go.</p><p><strong>Arguments</strong></p><ul>
-<li>(required) <code>integrationId</code> | To get the integration/app ID, go to &quot;<a href="https://developer.zapier.com//app/developer&quot;">https://developer.zapier.com//app/developer&quot;</a>, click on an integration, and copy the number directly after &quot;/app/&quot; in the URL.</li>
+<li>(required) <code>integrationId</code> | To get the integration/app ID, go to &quot;<a href="https://developer.zapier.com&quot;">https://developer.zapier.com&quot;</a>, click on an integration, and copy the number directly after &quot;/app/&quot; in the URL.</li>
```
kola-er, line 384: I'm not sure this info is related to the added feature. Should this and other places it shows up be in a separate PR?

gregilo: @kola-er These are a result of the pre-commit hooks set up for the project. I'm not sure why there were changes - I could try rebasing my branch and seeing if these go away.

gregilo: Just rebased and they went away :)

kola-er (approved):

PR #612, opened 2023-01-25
[PDE-3678] Block promote, upload, and push for missing required app info

kola-er:

### DIFF
--- a/packages/cli/src/oclif/commands/promote.js
+++ b/packages/cli/src/oclif/commands/promote.js
@@ -7,6 +7,7 @@ const { buildFlags } = require('../buildFlags');
 const { callAPI } = require('../../utils/api');
 const { flattenCheckResult } = require('../../utils/display');
 const { getVersionChangelog } = require('../../utils/changelog');
+const checkMissingAppInfo = require('../../utils/check-missing-app-info');
 
 const serializeErrors = (errors) => {
   const opener = 'Promotion failed for the following reasons:\n\n';
@@ -27,6 +28,8 @@ const serializeErrors = (errors) => {
 class PromoteCommand extends BaseCommand {
   async perform() {
     const app = await this.getWritableApp();
+
+    checkMissingAppInfo(app);
 
     const version = this.args.version;
     const assumeYes = 'yes' in this.flags;
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -43,6 +43,8 @@ const {
   validateApp,
 } = require('./api');
 
+const checkMissingAppInfo = require('./check-missing-app-info');
+
 const { runCommand, isWindows } = require('./misc');
 
 const debug = require('debug')('zapier:build');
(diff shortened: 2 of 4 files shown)


packages/cli/src/utils/checkMissingAppInfo.js
```diff
@@ -0,0 +1,7 @@
+module.exports = async function (app: Object) {
```
gregilo, line 1: Does this function have to be `async`? I would expect every line in this function to run synchronously.

packages/cli/src/oclif/commands/promote.js
```diff
@@ -7,6 +7,7 @@ const { buildFlags } = require('../buildFlags');
 const { callAPI } = require('../../utils/api');
 const { flattenCheckResult } = require('../../utils/display');
 const { getVersionChangelog } = require('../../utils/changelog');
+const checkMissingAppInfo = require('../../utils/checkMissingAppInfo');
```
gregilo, line 10: Good call! This is much simpler than a custom hook and more readable.

packages/cli/src/utils/checkMissingAppInfo.js
gregilo, line 1: The other files in the `utils` directory use `kebab-case` for their file naming. Can you please update this file to match?

gregilo, line 1: I'm seeing an error on command line: 

We're not using TypeScript for this project and I don't think node has built-in type annotations for function parameters. Can you please remove the `: Object` in `(app: Object)`?

packages/cli/src/utils/checkMissingAppInfo.js
```diff
@@ -0,0 +1,7 @@
...
+  const requiredFields = ['title', 'description', 'app_category', 'intention', 'role'];
+  const missingRequiredFields = requiredFields.filter(field => app[field] == null);
+  throw new Error(
+    `Your integration is missing required info (${missingRequiredFields.join(', ')}). Please, run "zapier register" to add it.`
+  );
```
gregilo, line 6: This function always throws an `Error` (even if the app has all required fields). Can you please update this to check if any required fields are missing?

Here's a suggestion: 
```suggestion
  if (missingRequiredFields.length) {
    throw new Error(
      `Your integration is missing required info (${missingRequiredFields.join(', ')}). Please, run "zapier register" to add it.`
    );
  }
```

gregilo (changes requested):

packages/cli/src/utils/checkMissingAppInfo.js
kola-er: Of course! 😄

gregilo (approved):

gregilo:
@kola-er Looks good - nice work! Let's hold off on merging this until our `zapier register` updates are merged.

packages/cli/src/utils/build.js
```diff
@@ -497,6 +499,7 @@ const buildAndOrUpload = async (
   let app;
   if (upload) {
     app = await getWritableApp();
+    checkMissingAppInfo(app);
```
eliangcs, line 502: We probably don't want to do this in `build`. `zapier build` only generates the zip file and doesn't save the app. We should allow the developer to build the zip even if there's missing app info.

PR #618, opened 2023-02-06
PDE-3732 feat(cli): Implement individual field flags for register command

gregilo:
This PR adds prompts for required information when running `zapier register`. It also allows users to pass these values in via flags.


packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
```diff
@@ -1,13 +1,30 @@
 const { callAPI } = require('../../utils/api');
+const { BASE_ENDPOINT, IS_TESTING } = require('../../constants');
 
 module.exports = async function (options) {
   // We only need to run this for the register command
   if (!options || !options.id || options.id !== 'register') {
     return null;
   }
 
+  // Mock API call for register field choices when testing
```
kola-er, line 10: I'm assuming you are having hard doing this mock within the test cases. Right?

packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -204,6 +204,34 @@ describe('smoke tests - setup will take some time', function () {
...
+    it('zapier register should throw error for invalid category', () => {
+      (() => {
+        runCommand(context.cliBin, [
+          'register',
+          '--category',
+          'invalidCategory',
+        ]);
+      }).should.throw(/invalidCategory is not a valid value for category/);
+    });
+  });
```
kola-er, line 233: It would be nice to have for a valid value for each flag too.

packages/cli/src/oclif/commands/register.js
```diff
@@ -32,7 +138,42 @@ RegisterCommand.args = [
...
+  commandFlags: {
+    desc: flags.string({
+      char: 'D',
+      description:
+        'A sentence describing your app in 140 characters or less, e.g. "Trello is a team collaboration tool to organize tasks and keep projects on track."',
+    }),
+    url: flags.string({
+      char: 'u',
+      description:
+        'The homepage URL of your app. This will also be used as the email domain that would allow team members to join your integration as a Collaborator. Learn more.',
```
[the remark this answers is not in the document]
kola-er: Since the "Learn more" doesn't link to somewhere else, I guess it's unhelpful.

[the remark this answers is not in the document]
kola-er: If you agree, we would have to remove it from the changes to the docs too.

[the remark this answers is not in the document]
gregilo: @kola-er good catch! I'll remove.

packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
gregilo: @kola-er Correct. I think this is due to us [packing the CLI package](https://github.com/zapier/zapier-platform/blob/4af40fd0b3b7958bc96799095d6fa247b54d07ec/packages/cli/src/smoke-tests/smoke-tests.js#L89) in our `before` block of our smoke tests. For some reason, `nock` isn't able to intercept the http call. Maybe this is because the test cases are [running the `register` command externally](https://github.com/zapier/zapier-platform/blob/4af40fd0b3b7958bc96799095d6fa247b54d07ec/packages/cli/src/smoke-tests/smoke-tests.js#L210) so the API call is outside of `nock`'s scope/access?

My workaround was to implement this mock API call directly into this `getAppRegistrationFieldChoices` hook with the following side effects:
* mock API call isn't present in the test cases (it also isn't obvious, so I should add a comment if we continue with this workaround)
* fixture needs to be [included in `packages/cli/src/oclif` directory](https://github.com/zapier/zapier-platform/blob/4af40fd0b3b7958bc96799095d6fa247b54d07ec/packages/cli/src/oclif/fixtures/registerFieldChoices.js) (instead of in `smoke-tests` directory where it should be) for npm to pack the CLI up properly
* `nock` had to be moved from `devDependencies` to `dependencies` in package.json (see [commit](https://github.com/zapier/zapier-platform/pull/618/commits/5d483845aba7c9332c6d67b83e592607011096d5))
    * This slows down the pack / build process during testing (since we use the [`--production` flag when running `npm install`](https://github.com/zapier/zapier-platform/blob/4af40fd0b3b7958bc96799095d6fa247b54d07ec/packages/cli/src/smoke-tests/smoke-tests.js#L47))

Do you have any thoughts about this workaround? Or alternative approaches?

packages/cli/src/smoke-tests/smoke-tests.js
gregilo: @kola-er I mainly wanted to test the enum fields since I wound up implementing [manual flag validation](https://github.com/zapier/zapier-platform/blob/4af40fd0b3b7958bc96799095d6fa247b54d07ec/packages/cli/src/oclif/commands/register.js#L26) for them. Is there anything in particular you'd like us to be testing for the other fields?

packages/cli/src/oclif/commands/register.js
```diff
@@ -32,7 +138,42 @@ RegisterCommand.args = [
...
+    }),
+    url: flags.string({
+      char: 'u',
+      description:
+        'The homepage URL of your app. This will also be used as the email domain that would allow team members to join your integration as a Collaborator.',
+    }),
+    audience: flags.string({
+      char: 'a',
+      description:
+        'Are you building a public or private integration? Public integration development receives priority support and launch promotion from the Zapier team.',
```
eliangcs, line 157: I think what we're working on makes the second sentence no longer true, so:

```suggestion
        'Are you building a public or private integration?',
```

packages/cli/src/oclif/commands/register.js
```diff
@@ -22,6 +22,112 @@ class RegisterCommand extends ZapierBaseCommand {
...
+      appMeta.title = await this.prompt(
+        'What is the title of your integration?',
+        { required: true }
+      );
+    }
+
+    appMeta.description = this.flags.desc?.trim();
+    if (!appMeta.description) {
+      appMeta.description = await this.prompt(
+        'Please provide a sentence describing your app in 140 characters or less.',
```
eliangcs, line 71:
```suggestion
        'Provide a short description about what your app does.',
```

packages/cli/src/oclif/commands/register.js
```diff
@@ -32,7 +138,42 @@ RegisterCommand.args = [
...
   },
 ];
-RegisterCommand.flags = buildFlags();
+
+RegisterCommand.flags = buildFlags({
+  commandFlags: {
+    desc: flags.string({
+      char: 'D',
+      description:
+        'A sentence describing your app in 140 characters or less, e.g. "Trello is a team collaboration tool to organize tasks and keep projects on track."',
```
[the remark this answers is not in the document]
gregilo: @eliangcs Jorge and I [discussed enforcing this limit in the CLI last week](https://zapier.slack.com/archives/C5Z9BP4U9/p1675095825596589?thread_ts=1675095407.301199&cid=C5Z9BP4U9). I can add custom validation to check the character limit of the provided description. I agree that we may change the character limit in the future, but there is a desire to enforce the current limit now. Are you comfortable with me proceeding with a 140 character limit validator?

packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
```diff
@@ -1,13 +1,30 @@
 const { callAPI } = require('../../utils/api');
+const { BASE_ENDPOINT, IS_TESTING } = require('../../constants');
 
 module.exports = async function (options) {
   // We only need to run this for the register command
   if (!options || !options.id || options.id !== 'register') {
```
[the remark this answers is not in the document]
gregilo: @eliangcs I was trying to fetch the enum field options dynamically and provide them to the `options` directive (outlined in the [flag options in oclif's docs](https://cdn.zappy.app/0d0706249ddc17164425e12c7ee0f9cc.png)) so that `zapier register --help` would show the available options for `audience`, `category` and `role`  and we wouldn't have to write our own validation, but I couldn't figure out how to provide these options in a static context (the flags are [expected in a static context](https://github.com/zapier/zapier-platform/blob/596f2efe887cb01e62559a75487034f5c76a0549/packages/cli/src/oclif/commands/register.js#L142)). This is also why I had to write [my own validation with `_validateEnumFlags`](https://github.com/zapier/zapier-platform/blob/596f2efe887cb01e62559a75487034f5c76a0549/packages/cli/src/oclif/commands/register.js#L26).

My thought was that using a hook would allow me to asynchronously fetch these options from our backend and provide them to `buildFlags`, but I was mistaken.

Do you know of any ways that we could accomplish this?

packages/cli/src/oclif/commands/register.js
```diff
@@ -32,7 +138,42 @@ RegisterCommand.args = [
...
+    }),
+    url: flags.string({
+      char: 'u',
+      description:
+        'The homepage URL of your app. This will also be used as the email domain that would allow team members to join your integration as a Collaborator.',
+    }),
+    audience: flags.string({
+      char: 'a',
```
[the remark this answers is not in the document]
gregilo: @eliangcs I think my answer [here](https://github.com/zapier/zapier-platform/pull/618#discussion_r1102073277) also applies to this comment.

packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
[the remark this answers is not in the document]
eliangcs: Thanks for the additional context! Using a hook to fetch options dynamically makes sense. I was able to override `options` by doing something like this in the hook:

```js
if (!options || !options.id || options.id !== 'register') {
  return null;
}

let formFields;
try {
  formFields = await callAPI('/apps/fields-choices');
} catch (e) {
  this.error(
    'Unable to connect to Zapier API. Please check your connection and try again.'
  );
}

const cmd = options.config.findCommand('register');
if (cmd && cmd.flags) {
  if (cmd.flags.audience) {
    cmd.flags.audience.options = formFields.intention.map((x) => x.value);
  }
  if (cmd.flags.role) {
    cmd.flags.role.options = formFields.role.map((x) => x.value);
  }
  if (cmd.flags.category) {
    cmd.flags.category.options = formFields.app_category.map((x) => x.value);
  }
}
```

which results in the following help text:

```
$ zapier-dev register --help
Register a new integration in your account.

USAGE
  $ zapier register [TITLE]

ARGUMENTS
  TITLE  Your integrations's public title. Asked interactively if not present.

OPTIONS
  -D, --desc=desc
      A sentence describing your app in 140 characters or less, e.g. "Trello is a team collaboration tool to organize tasks and keep projects on track."

  -a, --audience=global|private
      Are you building a public or private integration? Public integration development receives priority support and launch promotion from the Zapier team.

  -c, --category=accounting|ads-conversion|analytics|app-builder|app-families|bookmarks|business-intelligence|calendar|call-tracking|cms|commerce|communication|cont
  acts|content-files|crm|customer-appreciation|customer-support|dashboards|databases|developer-tools|devices|documents|drip-emails|ecommerce|education|email|email-n
  ewsletters|event-management|fax|files|filters|fitness|forms|fundraising|gaming|hr|human-resources|images|internet-of-things|invoices|it-operations|it-operations-e
  ducation|lifestyle-entertainment|marketing|marketing-automation|news|notes|notifications|payment-processing|phone|printing|product-management|productivity|project
  -management|reviews|sales-crm|scheduling|security-identity|server-monitoring|signatures|social|social-marketing|spreadsheets|support|tax|team-chat|team-collaborat
  ion|time-tracking|todo-lists|transactional-email|transcription|url-shorten|video|video-calls|webinars|website-app-building
      How would you categorize your app? Choose the most appropriate option for your app's core features.

  -d, --debug
      Show extra debugging output.

  -r, --role=employee|contractor|user
      What is your relationship with the app you're integrating with Zapier?

  -s, --[no-]subscribe
      Get tips and recommendations about this integration along with our monthly newsletter that details the performance of your integration and the latest Zapier
      news.

  -u, --url=url
      The homepage URL of your app. This will also be used as the email domain that would allow team members to join your integration as a Collaborator.

DESCRIPTION
  After running this, you can run zapier push to build and upload your integration for use in the Zapier editor.

  This will change the ./.zapierapprc (which identifies this directory as holding code for a specific integration).

EXAMPLES
  zapier register
  zapier register "My Cool Integration"
```

packages/cli/src/oclif/commands/register.js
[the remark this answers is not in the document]
eliangcs: > Are you comfortable with me proceeding with a 140 character limit validator?

I'm good with that as long as the 140-character limit is done on the CLI client. If we want to validate that on the server side, I'd say we let the server give the error and the CLI client shouldn't mention anything about the limit.

packages/cli/src/oclif/commands/register.js
```diff
@@ -32,7 +138,42 @@ RegisterCommand.args = [
...
+      char: 's',
+      description:
+        'Get tips and recommendations about this integration along with our monthly newsletter that details the performance of your integration and the latest Zapier news.',
+      allowNo: true,
+    }),
+  },
+});
 RegisterCommand.examples = [
   'zapier register',
   'zapier register "My Cool Integration"',
```
eliangcs, line 179: Can we add some more examples here?

packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
```diff
@@ -1,13 +1,30 @@
...
+
   const enumFieldChoices = {};
-  const formFields = await callAPI('/apps/fields-choices');
+  let formFields;
+
+  try {
+    formFields = await callAPI('/apps/fields-choices');
+  } catch (e) {
+    this.error(
+      'Unable to connect to Zapier API. Please check your connection and try again.'
```
eliangcs, line 25: To make debugging easier:

```suggestion
      `Unable to connect to Zapier API. Please check your connection and try again. ${e}`
```

packages/cli/src/oclif/commands/register.js
gregilo: @eliangcs Good call!

packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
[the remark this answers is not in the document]
eliangcs: > I will set these 3 flags to the flag.string type and leave _validateEnumFlags in place for validation. Are you okay with this?

Yes! I also think this is the best option we have. Thanks for testing!

eliangcs (commented):
I added a few more suggestions in the review. To summarize what we've discussed, I think this is ready to merge once these are also addressed:
- write unit tests for the `register` command
    - use `nock` to mock `fields-choices` endpoint
    - remove test-related code from `register.js`
    - remove smoke tests
- add the 140-char limit check for the `--description` flag
- add some valid examples for the `register` command
- add `?formId=create` in the API call so `subscription` gets saved ([related Slack conversation](https://zapier.slack.com/archives/C5Z9BP4U9/p1676282504996119?thread_ts=1675723870.749489&cid=C5Z9BP4U9))

packages/cli/src/smoke-tests/smoke-tests.js
kola-er: Just to test for when the validation succeeds as we have (all) for when the validation fails.

packages/cli/src/oclif/hooks/getAppRegistrationFieldChoices.js
gregilo: @eliangcs That makes sense to me. I've moved my tests to a new unit test file for the `register` command in 5825c479848f5aec92d67a4809b692cf7493cbf5.

packages/cli/src/oclif/commands/register.js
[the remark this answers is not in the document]
gregilo: @eliangcs I started using a constant for the max allowed description length (in 3b577478095c8916a83eb643b3e96af126c60133) - hopefully this will help make it easier to update if we ever need to change this value.

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -159,6 +159,10 @@ class ZapierBaseCommand extends Command {
    * @param {object} opts `inquierer.js` opts ([read more](https://github.com/SBoudrias/Inquirer.js/#question))
    */
   async prompt(question, opts = {}) {
+    if (typeof opts.required !== 'undefined' && opts.required) {
```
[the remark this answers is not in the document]
gregilo: @eliangcs I changed how prompt validation works quite a bit in 4610f36b3bbbb71ddfd96b8309039f8c226c2541. This no longer applies.

packages/cli/src/oclif/commands/register.js
gregilo: @eliangcs I added more here - 286afb7d0bd75538601a69b5baa102bdb32c908f. Do you think these are thorough enough?

packages/cli/src/oclif/commands/register.js
```diff
@@ -22,6 +22,112 @@ class RegisterCommand extends ZapierBaseCommand {
...
   }
+
+  _validateEnumFlags() {
+    const flagFieldMappings = {
+      audience: 'intention',
+      role: 'role',
+      category: 'app_category',
+    };
+
+    for (const flag of Object.keys(this.flags)) {
```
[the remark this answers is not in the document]
gregilo: Fixed in 093aef973e5285ca1076367ece1733247237ecd4 - thank you for the suggestion!

packages/cli/src/smoke-tests/smoke-tests.js
gregilo: @kola-er Ohhhh I see, I misunderstood your first comment. I will add these tests first thing tomorrow!

gregilo: @kola-er Added in c9ebfe474189855794f098b60c331501ef0c2467

eliangcs (approved):
I made two additional changes in the test code:
- hide stdout/stderr
- replace `fs.existsSync` with `fs.statSync`

Now this looks really good. Nice work, thanks!

gregilo:
@eliangcs Nice! I was struggling to hide stdout/stderr for some reason on my local, so I'm glad you updated that. I'm going to merge this PR now.

PR #619, opened 2023-02-08
[SUPPORTENG-383] docs(schema): Add clarity on role of performList for testing REST Hooks

armchairlinguist:
Testing a REST Hook in the Zapier editor, even for a private app, is quite difficult without a performList: there's no exposure of a URL to send data to, so no real hook data can be sent/loaded. This change attempts to make the importance clear without cluttering the schema definition.


eliangcs (approved):
Thanks for improving the doc!

PR #620, opened 2023-02-09
[SUPPORTENG-394] docs(cli): Add information about how to return line items

armchairlinguist:
The question of how to correctly return line items comes up periodically, to the point where we maintain an internal reference. This addition will allow developers curious about how to return line items to better self-serve.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -529,7 +529,7 @@ You can find more details on the definition for each by looking at the [Trigger 
 > To add a trigger, search, or create to an existing integration, run `zapier scaffold [trigger|search|create] [noun]` to create the necessary files to your project. For example, `zapier scaffold trigger post` will create a new trigger called "New Post".
 ### Return Types
 
-Each of the 3 types of function should return a certain data type for use by the platform. As of core v1.0.11, there are automated checks to let you know when you're trying to pass the wrong type back. For reference, each expects:
+Each of the 3 types of function should return a certain data type for use by the platform. There are automated checks to let you know when you're trying to pass the wrong type back. For reference, each expects:
 
 | Method  | Return Type | Notes                                                                                                                |
 |---------|-------------|----------------------------------------------------------------------------------------------------------------------|
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -529,15 +529,51 @@ You can find more details on the definition for each by looking at the [Trigger 
 > To add a trigger, search, or create to an existing integration, run `zapier scaffold [trigger|search|create] [noun]` to create the necessary files to your project. For example, `zapier scaffold trigger post` will create a new trigger called "New Post".
 ### Return Types
 
-Each of the 3 types of function expects a certain type of object. As of core v1.0.11, there are automated checks to let you know when you're trying to pass the wrong type back. For reference, each expects:
+Each of the 3 types of function should return a certain data type for use by the platform. As of core v1.0.11, there are automated checks to let you know when you're trying to pass the wrong type back. For reference, each expects:
 
 | Method  | Return Type | Notes                                                                                                                |
 |---------|-------------|----------------------------------------------------------------------------------------------------------------------|
-| Trigger | Array       | 0 or more objects; passed to the [deduper](https://zapier.com/developer/documentation/v2/deduplication/) if polling  |
+| Trigger | Array       | 0 or more objects; passed to the [deduper](https://platform.zapier.com/docs/dedupe/) if polling  |
 | Search  | Array       | 0 or more objects. Only the first object will be returned, so if len > 1, put the best match first                   |
-| Action  | Object      | Return values are evaluated by [`isPlainObject`](https://lodash.com/docs#isPlainObject)                              |
+| Create  | Object      | Return values are evaluated by [`isPlainObject`](https://lodash.com/docs#isPlainObject)                              |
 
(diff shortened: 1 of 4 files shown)


packages/cli/README-source.md
```diff
@@ -529,16 +529,52 @@ You can find more details on the definition for each by looking at the [Trigger
...
 
 | Method  | Return Type | Notes                                                                                                                |
 |---------|-------------|----------------------------------------------------------------------------------------------------------------------|
-| Trigger | Array       | 0 or more objects; passed to the [deduper](https://zapier.com/developer/documentation/v2/deduplication/) if polling  |
+| Trigger | Array       | 0 or more objects; passed to the [deduper](https://platform.zapier.com/docs/dedupe/) if polling  |
 | Search  | Array       | 0 or more objects. Only the first object will be returned, so if len > 1, put the best match first                   |
-| Action  | Object      | Return values are evaluated by [`isPlainObject`](https://lodash.com/docs#isPlainObject)                              |
+| Create  | Object      | Return values are evaluated by [`isPlainObject`](https://lodash.com/docs#isPlainObject)                              |
```
RasHanz, line 538: Nice catches!

packages/cli/README-source.md
```diff
@@ -529,16 +529,52 @@ You can find more details on the definition for each by looking at the [Trigger
...
 
 When a trigger function returns an empty array, the Zap will not trigger. For REST Hook triggers, this can be used to filter data if the available subscription options are not specific enough for the Zap's needs.
 
+#### Returning Line Items (Array of Objects)
+
+In some cases, you may want to include multiple items in the data you return for Searches or Creates. To do that, return the set of items as an array of objects under a descriptive key. This may be as part of another object (for example, items in an invoice) or as multiple top-level items.
+
+For example, a Create Order action returning an order with multiple items might look like this:
```
RasHanz, line 546: This is a personal nitpick, but the "for example" feel too close to each other when reading this. Maybe, "e.g." in the parenthesis? I can see that being used in other parts of the doc.

packages/cli/README-source.md
```diff
@@ -529,16 +529,52 @@ You can find more details on the definition for each by looking at the [Trigger
...
+```
+
+While a Find Users search could return multiple items under an object key within an array, like this:
+
+```
+result = [{
+  users: [
+      { name: 'Zap Zaplar', age: 12, city: 'Columbia', region: 'Missouri' },
+      { name: 'Orange Crush', age: 28, city: 'West Ocean City', region: 'Maryland' },
+      { name: 'Lego Brick', age: 91, city: 'Billund', region: 'Denmark' },
```
RasHanz, line 569:
```
    _ 
   [_] 
 /;   :\
()'---' C
  | | |
  [=|=]
```

packages/cli/README-source.md
```diff
@@ -529,16 +529,52 @@ You can find more details on the definition for each by looking at the [Trigger
...
+    ],
+  }];
+```
+
+A standard search would return just the inner array of users, and only the first user would be provided as a final result. Returning line items instead means that the "first result" return is the object containing all the user details within it.
+
+Using the standard approach is recommended, because not all Zapier integrations support line items directly, so users may need to take additional actions to reformat this data for use in their Zaps. More detail on that at [Use line items in Zaps](https://zapier.com/help/create/basics/use-line-items-in-zaps). However, there are use cases where returning multiple results is helpful enough to outweigh that additional effort.
```
RasHanz, line 576: I like that link!

RasHanz (approved):
> Note: I just reviewed one pair of documents. I'm assuming the other two are necessary duplicates.

The only change I'd ask you to consider, @armchairlinguist, is the one about the "for example", but I won't be opposed if you decide not to include that.

Feel free to send another review request if you include it, and I'll approve it quickly.

packages/cli/README-source.md
armchairlinguist: Oh, good catch. I can replace the parenthetical one with "like"!

RasHanz (approved):
LGTM! (sorry it took me a bit)

armchairlinguist:
Thanks! I'll leave it open til Monday in case anyone on the eng team has feedback, as well.

packages/cli/README-source.md
```diff
@@ -529,16 +529,52 @@ You can find more details on the definition for each by looking at the [Trigger
 > To add a trigger, search, or create to an existing integration, run `zapier scaffold [trigger|search|create] [noun]` to create the necessary files to your project. For example, `zapier scaffold trigger post` will create a new trigger called "New Post".
 ### Return Types
 
-Each of the 3 types of function expects a certain type of object. As of core v1.0.11, there are automated checks to let you know when you're trying to pass the wrong type back. For reference, each expects:
+Each of the 3 types of function should return a certain data type for use by the platform. As of core v1.0.11, there are automated checks to let you know when you're trying to pass the wrong type back. For reference, each expects:
```
eliangcs, line 532: While you were there, can we remove the reference of v1.0.11 as it's ancient and we can assume no one is using platform-core older than v1.0.11?

```suggestion
Each of the 3 types of function should return a certain data type for use by the platform. There are automated checks to let you know when you're trying to pass the wrong type back. For reference, each expects:
```

eliangcs (approved):
Happened to see a reference of an old platform version. It's unrelated to your changes, but it'd be nice to remove it too. Otherwise, this PR looks good to me, thanks!

packages/cli/README-source.md
armchairlinguist: Absolutely! That was actually so old I wasn't even sure if it was referring to the same version set. Thanks for the catch!

eliangcs (approved):

PR #623, opened 2023-02-20
PDE-3242 - feat(schema) Add enablePkce to OAuth2Config

standielpls:
Adds a new option to the OAuth2Config: `enablePkce`. 
This option will be read by the Zapier Platform backend to use PKCE for the authorization/access token exchange request flow. If this option is not provided, it will default to use the existing OAuth2 flow.


gregilo:
@standielpls This is backwards compatible, right? As in, end-users shouldn't be affected when this is released? (sorry, I'm not super familiar with the schema package yet)

eliangcs (approved):
The schema change looks good to me! The new `enablePkce` field is optional, so the change is backward compatible.

I suggest we update the [OAuth2](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#oauth2) section in the docs too. But you can do it in a separate PR.

PR #627, opened 2023-02-22
PDE-3733 feat(cli): Implement --yes flag for register command

gregilo:
This PR adds the `--yes` flag for the register command as well as the ability to update an existing integration's app meta via `zapier register`.


packages/cli/src/oclif/commands/register.js
```diff
@@ -63,37 +98,77 @@ class RegisterCommand extends ZapierBaseCommand {
...
 
     appMeta.description = this.flags.desc?.trim();
     if (!appMeta.description) {
       appMeta.description = await this.prompt(
         `Please provide a sentence describing your app in ${MAX_DESCRIPTION_LENGTH} characters or less.`,
-        { required: true, charLimit: MAX_DESCRIPTION_LENGTH }
+        {
+          required: true,
+          charLimit: MAX_DESCRIPTION_LENGTH,
+          default: this.app,
```
eliangcs, line 153:
```suggestion
          default: this.app?.description,
```

packages/cli/src/oclif/commands/register.js
```diff
@@ -63,37 +98,77 @@ class RegisterCommand extends ZapierBaseCommand {
...
+        }
       );
     }
 
     appMeta.homepage_url = this.flags.url;
-    if (!appMeta.homepage_url) {
+    if (!appMeta.homepage_url && !this.flags.yes) {
```
eliangcs, line 159: I don't think we should skip `homepage_url` prompt when `-y` is enabled. Any reason why we may want to skip?

packages/cli/src/oclif/commands/register.js
```diff
@@ -63,37 +98,77 @@ class RegisterCommand extends ZapierBaseCommand {
...
+    if (this._hasAppFile()) {
+      console.info(colors.yellow('.zapierapprc file detected.'));
+      if (this.flags.yes) {
+        console.info(
+          colors.yellow('-y/--yes flag passed, updating current integration.')
+        );
+        appMeta.action = actionChoices[0].value;
+      } else {
+        appMeta.action = await this.promptWithList(
+          'Would you like to update your current integration or register a new one?',
```
eliangcs, line 123: The `--yes` help text says "assume yes for all yes/no prompts", but this isn't a clear yes-or-no question. Maybe we could rephrase the question and add an app ID to make it more clear?

```
          'Would you like to update your current integration (ID: {app_id})?',
```

And in the choices, we can say:

```
      { name: 'Yes, update current integration', value: 'update' },
      { name: 'No, register a new integration', value: 'register' },
```

eliangcs (commented):
Overall it looks good! Apart from my suggestions in the comments, can we add one or two test cases to cover the `--yes` scenario? Thanks!

packages/cli/src/oclif/commands/register.js
gregilo: @eliangcs Oops, don't know how I missed this one. Thanks!

packages/cli/src/oclif/commands/register.js
gregilo: @eliangcs I figured since it's optional, it should be skippable, but I'm okay not skipping it with `-y`.

gregilo: Updated in 915e630cc2cf6cf3b7958ddac772012dda29b6e0

packages/cli/src/oclif/commands/register.js
gregilo: @eliangcs Good call! Updated in 4594d1c5bd2ba95426e92e212a056ca95f0f3969

eliangcs (dismissed):
I made some changes I hope you don't mind:

- Update the help text to describe the new behavior more accurately
- Add a test for the `--yes` flag
   - Through that I found we were sending the entire app object, including read-only fields like `id`, `status`, and `slug`, to the update endpoint. So I made changes to make sure only necessary fields are sent.

If the changes look good to you too, feel free to merge!

packages/cli/src/oclif/commands/register.js
```diff
@@ -63,37 +95,82 @@ class RegisterCommand extends ZapierBaseCommand {
...
       );
     }
 
     appMeta.intention = this.flags.audience;
     if (!appMeta.intention) {
       appMeta.intention = await this.promptWithList(
         'Are you building a public or private integration?',
-        this.config.enumFieldChoices.intention
+        this.config.enumFieldChoices.intention,
+        { default: this.app?.intention }
```
kola-er, line 173: @gregilo Why don't we have `required: true` for intention? Perhaps, I'm missing something?

packages/cli/src/oclif/commands/register.js
```diff
@@ -104,33 +181,45 @@ class RegisterCommand extends ZapierBaseCommand {
...
       );
     }
 
     appMeta.app_category = this.flags.category;
     if (!appMeta.app_category) {
       appMeta.app_category = await this.promptWithList(
         'How would you categorize your app?',
-        this.config.enumFieldChoices.app_category
+        this.config.enumFieldChoices.app_category,
+        { default: this.app?.app_category }
```
kola-er, line 194: Ditto `required: true` missing.

packages/cli/src/oclif/commands/register.js
```diff
@@ -104,33 +181,45 @@ class RegisterCommand extends ZapierBaseCommand {
         this._getRoleChoicesWithAppTitle(
           appMeta.title,
           this.config.enumFieldChoices.role
-        )
+        ),
+        { default: this.app?.role }
```
kola-er, line 185: Ditto `required: true` missing.

packages/cli/src/oclif/commands/register.js
kola-er: Learned that we don’t need it for the `promptWithList` fields because users have to make a selection out of the choices presented and no way out.

kola-er (approved):
Good to go

eliangcs (approved):

PR #628, opened 2023-02-27
PDE-3822 feat(cli): Only require fields for private integrations via CLI

gregilo:
Currently, app meta fields/settings are only editable for private apps in the visual builder. Now that the CLI allows users to update these settings via CLI, we need to ensure that they can only update these settings for **private** apps.

[Jira story](https://zapierorg.atlassian.net/browse/PDE-3822)

### DIFF
--- a/packages/cli/src/oclif/commands/register.js
+++ b/packages/cli/src/oclif/commands/register.js
@@ -132,6 +132,13 @@ class RegisterCommand extends ZapierBaseCommand {
       this.startSpinner('Retrieving details for your integration');
       this.app = await getWritableApp();
       this.stopSpinner();
+
+      // Block non-private apps from updating settings
+      if (this.app?.intention !== 'private') {
+        throw new Error(
+          "You can't edit settings for this integration. To edit your integration details on Zapier's public app directory, email partners@zapier.com."
+        );
+      }
     }
 
     appMeta.title = this.args.title?.trim();
--- a/packages/cli/src/oclif/commands/register.js
+++ b/packages/cli/src/oclif/commands/register.js
@@ -8,6 +8,7 @@ const {
   callAPI,
   getLinkedAppConfig,
   getWritableApp,
+  isPublished,
   writeLinkedAppConfig,
 } = require('../../utils/api');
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/oclif/commands/register.js
```diff
@@ -132,6 +132,13 @@ class RegisterCommand extends ZapierBaseCommand {
       this.startSpinner('Retrieving details for your integration');
       this.app = await getWritableApp();
       this.stopSpinner();
+
+      // Block non-private apps from updating settings
+      if (this.app?.intention !== 'private') {
```
eliangcs, line 137: Should we use the `status` field instead?

```suggestion
      if (this.app?.status !== 'private') {
```

gregilo: @eliangcs Great catch - I was unaware of how `app.status` is being used. I checked [the logic we're using in VB](https://gitlab.com/zapier/team-developer-platform/dev-platform/-/blob/9fa28d8bacd04ebdad5937bd039c71aede4ede47/web/frontend/assets/app/entities/CliApp/CliApp.ts#L96) to limit users from updating app settings for published apps and decided to implement that logic in 576b463ae49a1007ce4014c28d62bb36e26adbfd.

Can you please re-review?

eliangcs: Thanks for checking the logic on the UI side! Looks good to me!

eliangcs (approved):

packages/cli/src/oclif/commands/register.js
gregilo: @eliangcs Thanks for reviewing! I saw that you pushed ba6ca25072f70d22c138e8e7d60e8ab9a26f2bbf, but decided to revert it - is there a more ideal solution than what I've pushed?

eliangcs: @gregilo I thought the `public_ish` field returned by the server is defined as `status in ('public', 'beta')` and we could use it, but then I found out `public_ish` is in fact defined as `status == 'public'`, so I reverted it.

PR #629, opened 2023-03-02
feat(docs) - Add documentation for PKCE OAuth2

standielpls:
Updates the Authentication to include a section on PKCE for OAuth2.

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -446,6 +446,8 @@ If you define `fields` to collect additional details from the user, please note 
 
 ### OAuth2 with PKCE
 
+*Added in v14.0.0.*
+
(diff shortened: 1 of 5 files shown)


packages/cli/README-source.md
```diff
@@ -443,6 +443,33 @@ Also, `authentication.oauth2Config.getAccessToken` has access to the additional
 
 If you define `fields` to collect additional details from the user, please note that `client_id` and `client_secret` are reserved keys and cannot be used as keys for input form fields.
 
+
+### OAuth2 with PKCE
```
eliangcs, line 447: May be worth mentioning which version supports PKCE:

```suggestion
### OAuth2 with PKCE

*Added in v14.0.0.*
```

eliangcs (approved):
I only have a suggestion on mentioning on which version we start to support PKCE. Otherwise, it looks good to me! Thanks!

eliangcs (approved):

PR #630, opened 2023-03-06
PDE-3664 feat(cli): Use uniform field names in check-missing-app-info util

gregilo:
Since the BE field names that the API is expecting do not match the names of the flags we're using in the CLI, we need to make sure we're using the proper name in the proper context. This commit ensures that we're using proper names in the CLI error message when check-missing-app-info fails.


eliangcs (approved):
Good catch!

PR #640, opened 2023-04-06
fix(cli): Fix convert missing key in .zapierapprc file

FokkeZB:
When you run `zapier convert`, the `.zapierapprc` file that it generates is missing the `key` property.

This PR fixes that, and includes a test to prevent regressions.


eliangcs (approved):
Thanks!

PR #644, opened 2023-05-06
[PDE-3979] chore(schema): mark Important field as deprecated, remove examples

kreddlear:
We're removing the `important` field per [this epic](https://zapierorg.atlassian.net/browse/PDE-3962). We don't want to introduce a breaking change just for this (it doesn't actually change any behavior, the `important` field already doesn't do anything in the Editor), so for now we're just marking the field as deprecated and removing references to it in example apps and tests.


eliangcs (approved):
👍

PR #646, opened 2023-05-15
feat(core): Add support for caching

kola-er:
- Add `cache` object with functions `get`, `set`, and `delete` to the `z` object
- Provide a way for devs to test their integration code that uses caching


kola-er:
Note that I'm still looking at providing some form of testing utility to make testing devs code using z.cache easy.

packages/core/src/tools/create-cache.js
```diff
@@ -0,0 +1,49 @@
...
+  const runValidationChecks = (rpc, key, value = null, ttl = null) => {
+    if (!rpc) {
+      throw new Error('rpc is not available');
+    }
+
+    if (!_.isString(key)) {
+      throw new Error('key must be a string');
+    }
+
+    if (!JSON.stringify(value)) {
```
eliangcs, line 18: `JSON.stringify()` is not a reliable way to tell if a value is JSON-encodable. For example:

```
> JSON.stringify(console)
'{}'
> JSON.stringify({func: function() {}})
'{}'
```

I think we need to do something like [this](https://stackoverflow.com/questions/30579940/reliable-way-to-check-if-objects-is-serializable-in-javascript).

Suppose the function is named `isJsonEncodable()`. If `value` is a primitive type, like a number or string, `isJsonEcodable(value)` simply returns true. If `value` is a nested type, like an object or array, we need to recursively iterate each element and call `isJsonEncodable(element)`.

packages/core/src/tools/create-cache.js
```diff
@@ -0,0 +1,49 @@
...
+      throw new Error('value must be JSON-encodable');
+    }
+
+    if (ttl != null && !_.isInteger(ttl)) {
+      throw new Error('ttl must be a number');
+    }
```
eliangcs, line 24: I think it's okay for `ttl` to be a non-integer number.

```suggestion
    if (ttl != null && typeof ttl !== 'number') {
      throw new TypeError('ttl must be a number');
    }
```

Or if you want to keep `ttl` an integer, we should change the error message:

```js
    if (ttl != null && !_.isInteger(ttl)) {
      throw new TypeError('ttl must be an integer');
    }
```

packages/core/src/tools/create-cache.js
```diff
@@ -0,0 +1,49 @@
...
+
+const createCacheObj = (input) => {
+  const rpc = _.get(input, '_zapier.rpc');
+  const runValidationChecks = (rpc, key, value = null, ttl = null) => {
+    if (!rpc) {
+      throw new Error('rpc is not available');
+    }
+
+    if (!_.isString(key)) {
+      throw new Error('key must be a string');
```
eliangcs, line 15:
```suggestion
      throw new TypeError('key must be a string');
```

packages/core/test/tools/create-cache.js
```diff
@@ -0,0 +1,73 @@
...
+    should(result).eql(false);
+  });
+
+  it('zcache_set: should throw error for values that are not JSON-encodable', async () => {
+    const key = 'random-key'
+    const value = () => { 'this is a function' };
+
+    await cache.set(key, value).should.be.rejectedWith('value must be JSON-encodable');
+  });
+
```
eliangcs, line 55: Can we add another non-JSON-encodable test case where `value` is `console`?

```js
  it('zcache_set: should throw error for values that are not JSON-encodable', async () => {
    const key = 'random-key'
    const value = console;

    await cache.set(key, value).should.be.rejectedWith('value must be JSON-encodable');
  });
```

eliangcs (commented):
Looks good overall! I left some suggestions in the comments.

packages/core/src/tools/ensure-json.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  for (const key in obj) {
+    if (!isJSONEncodable(obj[key])) {
+      return false;
+    }
+  }
+
+  return true;
+};
```
eliangcs, line 19: As I looked closer, I think it's necessary to provide some debug information in the error message for deeply-nested objects. So instead of returning a boolean, this function should be named `ensureJsonEncodable` and throw an error when the input value is not JSON-encodable. The debug info can just be a path showing which part of the object is unacceptable.

We should also handle circular referencing.

With that, I came up with this [example code](https://gist.github.com/eliangcs/ddff58ddda957905eac2dd85e6bc6a11) with some test cases. If that looks good to you, can you copy the function and the test cases to here? Thanks!

kola-er: Refactored, thanks for the code.

packages/core/src/tools/create-app-tester.js
```diff
@@ -29,7 +29,7 @@ const createAppTester = (appRaw, { customStoreKey } = {}) => {
 
   const randomSeed = genId();
 
-  return (methodOrFunc, bundle) => {
+  const appTester = (methodOrFunc, bundle, clearZcacheBeforeUse = null) => {
```
kola-er, line 32: This refactor of the `createAppTester` used for testing integrations allows devs to test the caching part of their integration code. Two things:
- It uses an object `zcacheTestObj` saved on this function `appTester` as the cache local storage. So the production cache won't be used during testing.
- It uses argument `clearZcacheBeforeUse` to know when to clear the cache in the test case. So devs have a way to ensure an empty cache is used for separate test cases.

A typical example of how devs would test caching is done here https://github.com/zapier/zapier-platform/pull/646/files#diff-e507cf9820f17d4bfa0d153e57ca25f22daf1b0f0e7dd7a44c86deb9a3f44846

packages/core/test/tools/create-cache.js
```diff
@@ -0,0 +1,75 @@
...
+  it('zcache_set: should set a cache entry when current throughput is within app\'s rate-limit', async () => {
+    const key = 'random-key'
+    const value = {entity:'Zapier', colors: ['Orange', 'black']};
+    mockRpcCall(JSON.stringify(true));
+
+    const result = await cache.set(key, value);
+    should(result).eql(true);
+  });
+
+  it('zcache_set: should not set a cache entry when current throughput is outside app\'s rate-limit', async () => {
```
standielpls, line 40: Correct me if I'm wrong, but are we using the previous "set" test case as a basis for this test case, otherwise, I'm having trouble understanding what the difference in behaviour is between the two. I recommend having isolated test cases if the above is true

standielpls (commented):
Looks good, just a comment about a test

packages/core/src/tools/create-rpc-client.js
```diff
@@ -8,11 +8,38 @@ const { genId } = require('./data');
...
+
+  if (method === 'zcache_delete') {
+    if (key in zcacheTestObj) {
+      delete zcacheTestObj[key];
+      return JSON.stringify(true);
+    }
+
+    return JSON.stringify(false);
+  }
+};
```
eliangcs, line 30: Maybe throw an error for unreachable code?

```suggestion
    throw new Error(`Unexpected method '${method}'`);
};
```

packages/core/src/tools/create-app-tester.js
eliangcs, line 32:
```suggestion
  const appTester = (methodOrFunc, bundle, clearZcacheBeforeUse = false) => {
```

eliangcs (approved):
I have two minor suggestions. Otherwise, the PR looks good to me! Nice work!

eliangcs:
There is some additional work we need to do, maybe in separate PRs:

- Documentation
  - How to use `z.cache`
  - How `z.cache` behaves in test code (like how `ttl` isn't supported)
- Add typings for `z.cache` in [`index.d.ts`](https://github.com/zapier/zapier-platform/blob/47d873ebcd24adf9403cce42c773a664e5d1127e/packages/core/index.d.ts)

packages/core/test/tools/create-cache.js
kola-er: Makes sense. I've combined the two test cases.

kola-er:
fyi, the typings for `z.cache` was added in the last commit. The documentation will be added in a separate PR. Needs another approval.

packages/core/index.d.ts
```diff
@@ -193,4 +194,10 @@ export interface ZObject {
     RefreshAuthError: typeof RefreshAuthError;
     ThrottledError: typeof ThrottledError;
   };
+
+  cache: {
+    get: (key: string) => Promise<any>;
+    set: (key: string, value: string, ttl?: integer) => Promise<boolean>;
```
eliangcs, line 200: Should `value` be `any` instead?

```suggestion
    set: (key: string, value: any, ttl?: integer) => Promise<boolean>;
```

kola-er: 🤦 thanks

eliangcs (approved):

packages/core/src/tools/create-cache.js
```diff
@@ -0,0 +1,46 @@
...
+
+    ensureJSONEncodable(value);
+  };
+
+  return {
+    get: async (key) => {
+      runValidationChecks(rpc, key);
+
+      const result = await rpc('zcache_get', key);
+      return JSON.parse(result);
```
eliangcs, line 31:
```suggestion
      const result = await rpc('zcache_get', key);
      return JSON.parse(result) if result else null;
```

PR #652, opened 2023-06-02
PDE-3946 feat(cli): Adds a new cache clear command

standielpls:
Adds a new cache command and subcommand clear.

### DIFF
--- a/packages/cli/src/oclif/commands/cache/clear.js
+++ b/packages/cli/src/oclif/commands/cache/clear.js
@@ -11,8 +11,12 @@ class ClearCacheCommand extends BaseCommand {
 
     const { majorVersion } = this.args;
 
-    let selectedMajorVersion =
-      majorVersion === '' ? null : Number(majorVersion);
+    let selectedMajorVersion = majorVersion ? Number(majorVersion) : null;
+    if (Number.isNaN(selectedMajorVersion)) {
+      throw new Error(
+        `Invalid major version '${majorVersion}'. Must be a number.`
+      );
+    }
 
     const majorVersions = [
       ...new Set(
@@ -31,7 +35,7 @@ class ClearCacheCommand extends BaseCommand {
     } else {
       if (!majorVersions.includes(selectedMajorVersion)) {
         throw new Error(
-          `This app does not have any versions on major version '${selectedMajorVersion}'. Valid versions are: ${majorVersions.join(
+          `This integration does not have any versions on major version '${selectedMajorVersion}'. Valid versions are: ${majorVersions.join(
             ', '
           )}`
         );
@@ -40,7 +44,7 @@ class ClearCacheCommand extends BaseCommand {
 
     if (
       !(await this.confirm(
-        `Are you sure you want to clear the app cache for major version '${cyan(
+        `Are you sure you want to clear all cache data for major version '${cyan(
           selectedMajorVersion
         )}'?`,
         true
@@ -54,10 +58,7 @@ class ClearCacheCommand extends BaseCommand {
     const { id: appId } = await getWritableApp();
     const url = `/apps/${appId}/major-versions/${selectedMajorVersion}/cache`;
 
-    await callAPI(url, {
-      url: url.startsWith('http') ? url : undefined,
-      method: 'DELETE',
-    });
+    await callAPI(url, { method: 'DELETE' });
 
     this.stopSpinner();
 
@@ -83,7 +84,7 @@ class ClearCacheCommand extends BaseCommand {
     });
 
     return await this.promptWithList(
-      'Which major version app cache would you like to delete?',
+      "Which major version's cache data would you like to delete?",
       majorVersionChoices,
       { default: currentMajorVersion }
     );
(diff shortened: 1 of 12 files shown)


packages/cli/src/oclif/commands/cache/clear.js
```diff
@@ -0,0 +1,111 @@
...
+class ClearCacheCommand extends BaseCommand {
+  async perform() {
+    this.startSpinner('Fetching versions...');
+    const { versions } = await listVersions();
+    this.stopSpinner();
+
+    const { majorVersion } = this.args;
+
+    let selectedMajorVersion =
+      majorVersion === '' ? null : Number(majorVersion);
```
eliangcs, line 15: I got this error when I ran `zapier cache:clear`:

```
$ zapier-dev cache:clear
✔ Fetching versions...
 ›   Error: This app does not have any versions on major version 'NaN'. Valid versions are: 2, 1
 ›
```

When executing `zapier cache:clear` without args, `majorVersion` is `undefined`. So after this line, `selectedMajorVersion` will be `Number(undefined)`, which is `NaN`.

To fix, we should:
1. turn any falsey value (`undefined`, `null`, `''`) to `null`.
2. check any string that can't be parsed into a number (such string makes `Number(s)` a `NaN`)

```suggestion
    let selectedMajorVersion = majorVersion ? Number(majorVersion) : null;
    if (Number.isNaN(selectedMajorVersion)) {
        throw new Error(`Invalid major version '${majorVersion}'. Must be a number.`);
    }
```

packages/cli/src/oclif/commands/cache/clear.js
```diff
@@ -0,0 +1,111 @@
...
+
+    if (selectedMajorVersion === null) {
+      selectedMajorVersion = await this._promptForMajorVersionSelection(
+        majorVersions,
+        currentVersion
+      );
+    } else {
+      if (!majorVersions.includes(selectedMajorVersion)) {
+        throw new Error(
+          `This app does not have any versions on major version '${selectedMajorVersion}'. Valid versions are: ${majorVersions.join(
```
eliangcs, line 34:
```suggestion
          `This integration does not have any versions on major version '${selectedMajorVersion}'. Valid versions are: ${majorVersions.join(
```

packages/cli/src/oclif/commands/cache/clear.js
```diff
@@ -0,0 +1,111 @@
...
+            ', '
+          )}`
+        );
+      }
+    }
+
+    if (
+      !(await this.confirm(
+        `Are you sure you want to clear the app cache for major version '${cyan(
```
eliangcs, line 43:
```suggestion
        `Are you sure you want to clear all cache data for major version '${cyan(
```

packages/cli/src/oclif/commands/cache/clear.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    this.startSpinner('Clearing cache');
+    const { id: appId } = await getWritableApp();
+    const url = `/apps/${appId}/major-versions/${selectedMajorVersion}/cache`;
+
+    await callAPI(url, {
+      url: url.startsWith('http') ? url : undefined,
+      method: 'DELETE',
+    });
```
eliangcs, line 60: Looks like the `url.startsWith('http')` will always be false and unnecessary?

```suggestion
    await callAPI(url, { method: 'DELETE' });
```

packages/cli/src/oclif/commands/cache/clear.js
```diff
@@ -0,0 +1,111 @@
...
+      return {
+        name: `${v}${
+          isCurrentMajorVersion ? ` (current version '${currentVersion}')` : ''
+        }`,
+        value: v,
+      };
+    });
+
+    return await this.promptWithList(
+      'Which major version app cache would you like to delete?',
```
eliangcs, line 86:
```suggestion
      "Which major version's cache data would you like to delete?",
```

packages/cli/src/oclif/commands/cache/clear.js
```diff
@@ -0,0 +1,111 @@
...
+ClearCacheCommand.args = [
+  {
+    name: 'majorVersion',
+    description:
+      '(Optional) The app cache will be deleted for this major version. If not provided, you must pick from a list of major versions for this app.',
+    required: false,
+  },
+];
+ClearCacheCommand.flags = buildFlags();
+ClearCacheCommand.description = `Clears the app cache for a major version.
```
eliangcs, line 102:
```suggestion
ClearCacheCommand.description = `Clear the cache data for a major version.
```

eliangcs (commented):
I suggested some changes on the wording:

- use "integration" instead of "app" so it's consistent with the other commands
- use "cache data" instead of "app cache"

Another suggestion: `zapier cache --help` prints:

```
$ zapier-dev cache --help
Clear the cache data for a major version.

USAGE
  $ zapier cache:COMMAND

DESCRIPTION

  This command clears the cache data for a major version of your integration.
  The job will be run in background and may take some time to complete.
  You can check `zapier history` to see the job status.
```

Can we fix the description so it's not specific to `cache:clear`? For example, `zapier team --help` give you this:

```
$ zapier-dev team --help
Add, remove, or get team members of your integration.

USAGE
  $ zapier team:COMMAND

COMMANDS
  team:add     Add a team member to your integration.
  team:get     Get team members involved with your integration.
  team:remove  Remove a team member from all versions of your integration.
```

This can be done by adding the command to [`topics` in `package.json`](https://github.com/zapier/zapier-platform/blob/efd0004f7c02f41d5bb603bf5f246731cb33d036/packages/cli/package.json#L108-L121).

eliangcs (approved):
Looks good, thanks! #657 should fix the failing test and unblock you from merging this.

PR #659, opened 2023-06-09
docs(examples) Update oauth2 example to use shorthands

FokkeZB:
Use HTTP shorthands instead of functions to match the [docs](https://platform.zapier.com/cli_docs/docs#oauth2) and promote simple over easy.


eliangcs (commented):
Changing an example in `example-apps` won't affect `zapier init --template <template_name>`. The auth code is generated by [`writeGenericAuth()`](https://github.com/zapier/zapier-platform/blob/c0da67300d1b2091e7b83c74fccac575bfeb371b/packages/cli/src/generators/index.js#L75). To change the `oauth2` project template, you need to:

1. update the code in [`auth-files-codegen.js`](https://github.com/zapier/zapier-platform/blob/c0da67300d1b2091e7b83c74fccac575bfeb371b/packages/cli/src/utils/auth-files-codegen.js#L27)
2. run `zapier init --template oauth2` and make sure the generated project works out of the box
3. copy the generated project to `example-apps/oauth2` (Projects in `example-apps` are only for easy browsing. They aren't used in the code.)

I have a concern: If we replaced functions with shorthands, we wouldn't have any example code left for function-style OAuth2. Maybe we should make another template like `oauth2-function` for function-style code?

FokkeZB:
O wow.. didn't know we were generating those templates at run-time.

In light of the technical platform vision we do want to lean towards definition more and code less, but perhaps this is better left to a broader initiative by Dev Platform them.

Closing.

PR #662, opened 2023-06-12
PDE-4011 - BREAKING CHANGE(schema) - Remove the important field

standielpls:
Following the deprecation of this field https://github.com/zapier/zapier-platform/pull/644, this PR removes the `important` field from the `BasicDisplaySchema`


rnegron (approved):
🙌🏻 
Screen Shot 2023-06-12 at 2 19 57 PM

eliangcs (approved):
I grepped the codebase and didn't find any other occurrences of `important` that we need to remove. So it looks good to me, thanks!

PR #665, opened 2023-06-14
chore - update-dependencies

standielpls:
mostly version bumps
ran `yarn upgrade-interactive --latest`

verified tests pass, and can init `zapier init app`


packages/cli/package.json
```diff
@@ -49,28 +49,28 @@
     "cli-table3": "0.6.3",
     "colors": "1.4.0",
     "debug": "4.3.4",
-    "fs-extra": "10.0.0",
+    "fs-extra": "11.1.1",
```
rnegron, line 52: Looks like this `fs-extra` package is the only major version change and[ it's related to requiring Node >14,](https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md#1100--2022-11-28) so we should be good! 👍🏻

rnegron (approved):

eliangcs (approved):

PR #674, opened 2023-07-05
fix(cli) Use same test template for convert as for scaffold, with jest instead of mocha

FokkeZB:
We promote jest: https://platform.zapier.com/cli_docs/docs#writing-unit-tests

But `zapier convert` still uses `mocha`.

This PR fixes that, and while at it:

- Removes unused `package.template.json`.
- Uses `init`'s `gitignore` instead of duplicate for `convert`.
- Replaces `create`, `search` and `trigger` specific test template with `scaffold`'s generic `test.template.js`.
- Drops `/* globals ...` from `test.template.js` as most will have those globals set in their eslint config, like we do in our monorepo, resulting in `no-redeclare` errors.
- Updates `jest` dependency for both `init` and `convert`.

`IP-137` [Fix zapier convert using mocha instead of jest](https://zapierorg.atlassian.net/browse/IP-137)


eliangcs (approved):
Thanks! I tested it quickly and it works great. Merging.

standielpls:
Noticing a failed test after this merged to main, investigating

https://github.com/zapier/zapier-platform/actions/runs/5483945149/jobs/9990859198

standielpls:
Fixed here: https://github.com/zapier/zapier-platform/pull/677

PR #681, opened 2023-07-26
feat(core) - Add support for locking

standielpls:
Add support for locking.
To help review the code, the locking usage would look like this:

this is the interface for devs
```js
lock(key: [string], value: string, callbackFn: Function, maxLockSec: number, scope: 'global' | 'user' | 'auth')
```

```js
const expensiveCall = () => z.request(requestOptions)
const response = await z.lock(['lyrics', params.songId], 10, expensiveCall, 'global')
```
if lock acquired - return results from `expensiveCall`, calls unlock
if lock is held by another process - throw ThrottledError
if lock failed for other reasons - throw generic Error

### DIFF
--- a/packages/core/src/tools/create-lock.js
+++ b/packages/core/src/tools/create-lock.js
@@ -1,0 +1,77 @@
+'use strict';
+
+const _ = require('lodash');
+const { ThrottledError } = require('../errors');
+
+const createLock = (input, cache) => {
+  const _unlock = async (key, scope) => {
+    await cache.delete(key, 'locking', scope);
+  };
+
+  const validateInput = (key, callbackFn, maxLockSec, scope) => {
+    if (!Array.isArray(key)) {
+      throw new TypeError('key must be an array');
+    }
+
+    if (!_.isFunction(callbackFn)) {
+      throw new TypeError('callbackFn must be a function');
+    }
+
+    if (!_.isInteger(maxLockSec)) {
+      throw new TypeError('maxLockSec must be an integer');
+    } else if (maxLockSec <= 0) {
+      throw new TypeError('maxLockSec must be greater than 0');
+    }
+
+    if (scope && scope !== 'user' && scope !== 'global' && scope !== 'auth') {
+      throw new TypeError(
+        `scope '${scope}' is invalid. It must be one of user, global, or auth`
+      );
+    }
+  };
+  const _lock = async (key, callbackFn, maxLockSec, scope) => {
+    validateInput(key, callbackFn, maxLockSec, scope);
+
+    const lockKey = key.join('-');
+
+    // Set the lock expiry as the value
+    const expirationTimeMs = Date.now() + maxLockSec * 10000;
+
+    const response = await cache.set(
+      lockKey,
+      expirationTimeMs,
+      maxLockSec,
+      'locking',
+      scope
+    );
+
+    // Did not acquire lock, try again later. The response value is the lock expiry
+    // time, so we can calculate the remaining time to wait.
+    if (typeof response === 'string') {
+      const remainingTimeSec = (Number(response) - Date.now()) / 10000;
+      throw new ThrottledError(
+        'Unable to acquire lock, trying again in ' +
+          remainingTimeSec +
+          ' seconds',
+        remainingTimeSec
+      );
+    }
+
+    // Lock acquired, will invoke callbackFn and return
+    let callbackResponse;
+    if (response) {
+      try {
+        callbackResponse = await callbackFn();
+      } finally {
+        await _unlock(lockKey, scope);
+      }
+      return callbackResponse;
+    }
+
+    throw new Error('Unable to acquire lock due to server issues.');
+  };
+  return (key, callbackFn, maxLockSec, scope = 'user') =>
+    _lock(key, callbackFn, maxLockSec, scope);
+};
+
+module.exports = createLock;
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -125,6 +125,14 @@ type DehydrateFunc = <T>(
   inputData: T
 ) => string;
 
+type LockFunc = <T>(
+  key: [string],
+  value: string,
+  maxLockSec: number,
+  callbackFn: () => Promise<T>,
+  scope: string
+) => Promise<T>;
+
 export interface ZObject {
   request: {
     // most specific overloads go first
@@ -200,4 +208,5 @@ export interface ZObject {
     set: (key: string, value: any, ttl?: number) => Promise<boolean>;
     delete: (key: string) => Promise<boolean>;
   };
+  lock: LockFunc;
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -58,7 +58,8 @@
     "aws-sdk": "^2.1397.0",
     "dicer": "^0.3.1",
     "fs-extra": "^11.1.1",
-    "mock-fs": "^5.2.0"
+    "mock-fs": "^5.2.0",
+    "timekeeper": "^2.2.0"
   },
   "optionalDependencies": {
     "@types/node": "^20.3.1"
--- a/packages/core/src/app-middlewares/before/z-object.js
+++ b/packages/core/src/app-middlewares/before/z-object.js
@@ -4,6 +4,7 @@ const _ = require('lodash');
 
 const createAppRequestClient = require('../../tools/create-app-request-client');
 const createCache = require('../../tools/create-cache');
+const createLock = require('../../tools/create-lock');
 const createDehydrator = require('../../tools/create-dehydrator');
 const createFileStasher = require('../../tools/create-file-stasher');
 const createJSONtool = require('../../tools/create-json-tool');
@@ -19,8 +20,9 @@ const hashing = require('../../tools/hashing');
 */
 const injectZObject = (input) => {
   const bundle = _.get(input, '_zapier.event.bundle', {});
+  const cache = createCache(input);
   const zRoot = {
-    cache: createCache(input),
+    cache,
     console: createLoggerConsole(input),
     cursor: createStoreKeyTool(input),
     dehydrate: createDehydrator(input, 'method'),
@@ -29,6 +31,7 @@ const injectZObject = (input) => {
     generateCallbackUrl: createCallbackHigherOrderFunction(input),
     hash: hashing.hashify,
     JSON: createJSONtool(),
+    lock: createLock(input, cache),
     require: (moduleName) => require(moduleName),
     stashFile: createFileStasher(input),
   };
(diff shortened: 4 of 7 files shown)


packages/core/src/tools/create-lock.js
```diff
@@ -0,0 +1,77 @@
...
+        callbackResponse = await callbackFn();
+      } finally {
+        await _unlock(lockKey, scope);
+      }
+      return callbackResponse;
+    }
+
+    throw new Error('Unable to acquire lock due to server issues.');
+  };
+  return (key, callbackFn, maxLockSec, scope = 'user') =>
```
kola-er, line 73:
```suggestion
  return (key, maxLockSec, scope = 'user', callbackFn) =>
```

Any way we can make the `callbackFn` be the last with the scope still having a default? This would make it simple to use for devs like the following;

```js
const response = await z.lock(key, maxLockSec, scope, () => {
  // expensive calls
});
```

kola-er:
Looks good. Left some comments.

eliangcs (commented):
I'll test the code further, but these are what I think about the code for now. Thanks!

### 1. Make `key` a string rather than an array

As a developer, I would prefer to format my own `key`. Code readability won't be a problem because we have template strings in JS:

```js
await z.lock(`task-${bundle.inputData.taskId}`, ...);
```

This way `z.lock()` can be kept simple and doesn't have to [hard code a delimiter](https://github.com/zapier/zapier-platform/blob/32922361a85430ea8452775dd3440e93614d3468/packages/core/src/tools/create-lock.js#L35). This would also prevent the developer from shooting their own foot by accidentally passing an object in the array that results in something like this:

```
> data = {user: {id: 1}}
{ user: { id: 1 } }
> [data.user].join('-')
'[object Object]'
```

@FokkeZB curious about your thoughts on this too. Thanks!

### 2. Use destructured parameters for parameters with a default

Instead of:

```js
function lock(key, option1 = 'default1', option2 = 'default2', ..., callback) { }
```

it's better to group all the options in an object:

```js
function lock(key, {option1 = 'default1', option2 = 'default2'} = {}, callback) {}
```

The latter uses the [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#destructured_parameter_with_default_value_assignment) syntax.

It's better because it'd be easier to add/remove arguments without worrying about breaking backward compatibility.

### 3. Wait time and lock TTL should be separate

I feel the names, `expirationTimeMs` and `maxLockSec`, are a little confusing. They are two different time lengths:

- `waitTime` (currently `expirationTimeMs`): the max time that `await z.lock()` can block before throwing `ThrottledError`.
- `lockTTL` (currently `maxLockSec`): the life time of the lock that we'll keep in the cache storage. Normally, the developer should release the lock earlier than the `lockTTL` by unlocking. But in case where the developer doesn't unlock, the lock should be removed automatically after `lockTTL`.

A developer may want to set different values to `waitTime` and `lockTTL`. So together with 2, maybe we can change the function signature to the following?

```typescript
interface LockOptions {
  waitSecs?: number;  // maybe defaults to 10 (seconds)
  lockTTL?: number;  // maybe defaults to 10 (seconds)
  scope?: string;  // defaults to 'user'
}

type LockFunc = <T>(
  key: [string],  
  callbackFn: () => Promise<T>,
  options?: LockOptions
) => Promise<T>;
```

`options` is the last so it can be optional. @kola-er you [suggested](https://github.com/zapier/zapier-platform/pull/681#discussion_r1278004666) we put `callbackFn` at the end, so I wonder what you think about this too. Thanks!

FokkeZB:
@casshill @jasonkotenko I'd love your feedback on the `z.lock()` interface as well.

Here's mine on the topics @eliangcs raised:

1. I agree a string key is more clear, standard and transparent.
2. I agree an options object is more clear and standard, but as they should be optional, and the callback isn't, I'd make them the last argument. That is also how I see most other libraries do it.
3. I agree there should be 2 options the wait/lock time, with the same unit, and I'd also keep the terms similar. I [suggested](https://coda.io/d/_di0MgBhlCWf/Support-Locking-in-the-CLI_sufKD#_luIXR) `maxWaitSec` and `maxLockSec`.

Other than that:

- I'd prefer if we'd error if we reach `maxLockSec`, otherwise nobody is aware they need to increase it.
- Are we also implementing the [configuration alternative](https://coda.io/d/_di0MgBhlCWf/Support-Locking-in-the-CLI_sufKD#_lus40), which:
  - Reduces the risk of running into Lambda timeouts, as when the lock wraps a slow process, the Lambda timeout would not be required to be at least double the time that may take, and the limit imposed by Celery's timeout of 180s is ~160s.
  - Prevents having that (more) expensive Lambda process sit idle.
  - Reduces risk of a Lambda operation having finished half-way, as nothing can have partially be executed before the lock failed to get acquired.

kola-er:
@eliangcs @FokkeZB The `expirationTimeMs` is not the same as `waitTime` or `waitSec`. Although the locking implementation in the monolith has a form of waiting before giving up on acquiring the lock in the same request to try again later, the plan is to not add it here. We think knowing the amount of time left for the lock to be released makes "waiting" unnecessary. So instead of waiting in a single request using `waitTime` or `waitSec`, the request fails immediately returning the expiration time of the current lock `expirationTimeMs` and[ the platform calculates the time left for the lock to be released from it and throws the ThrottledError with the time to wait for](https://github.com/zapier/zapier-platform/blob/32922361a85430ea8452775dd3440e93614d3468/packages/core/src/tools/create-lock.js#L50-L58) before retrying.

eliangcs:
> We think knowing the amount of time left for the lock to be released makes "waiting" unnecessary.

@kola-er thanks for the pointer! Sorry for misunderstanding the code! Please ignore the part where I wrote about the wait time then.

packages/core/src/tools/create-lock.js
```diff
@@ -0,0 +1,77 @@
...
+      );
+    }
+
+    // Lock acquired, will invoke callbackFn and return
+    let callbackResponse;
+    if (response) {
+      try {
+        callbackResponse = await callbackFn();
+      } finally {
+        await _unlock(lockKey, scope);
```
kola-er, line 66: >I'd prefer if we'd error if we reach maxLockSec, otherwise nobody is aware they need to increase it.

As suggested by @FokkeZB, we can log here when the current time is newer than `expirationTimeMs`.

FokkeZB:
> @eliangcs @FokkeZB The `expirationTimeMs` is not the same as `waitTime` or `waitSec` [..] instead of waiting in a single request using `waitTime` or `waitSec`, the request fails immediately returning the expiration time of the current lock `expirationTimeMs`

Ah, missed that. That makes sense and also mitigates my concern of this causing us to spend more time in Lambda and potentially running into timeouts.

One reason why I'd prefer the config mode still is that it better aligns with the technical vision

standielpls:
> One reason why I'd prefer the [config option](https://coda.io/d/_di0MgBhlCWf/Support-Locking-in-the-CLI_sufKD#_lus40) still is that it would work with the Platform UI and better fit the [technical vision](https://coda.io/d/Integration-Platform-Zone_dra_rAeNZ0-/Technical-Vision_suaPR) as that is driving to less code, or at least less code that is Zapier-specific and thus harder for AI to generate / assist with.

Thank you for sharing! @FokkeZB 
The `config` option pushes the locking logic from platform-core to the backend, which was the main discussion item when we were formulating a design. That is, we thought that by reducing the need to reuse shared code in the monolith will help us decouple from the monolith in the future. We haven't thought about this from a technical vision perspective. 

We discussed it here originally: https://zapier.slack.com/archives/C5Z9BP4U9/p1689157277734509

casshill:
We tried setting a low wait time for Google Sheets and it (caused an incident?) didn't work. We had to go with short wait times and multiple retries (via throttle errors). Not sure if this would be the same. 

I think the developer should just set one time not 2 (maxLockSec). The wait should be whatever works best for us. 

The config option is nice and simple. Would this restrict the lock to an entire perform task, not just a single function/http request? As long as we are fine to handle more/longer locks, as sometimes a perform function can do several things/make several requests and not all of them may need the lock

standielpls:
> We tried setting a low wait time for Google Sheets and it (caused an incident?) didn't work. We had to go with short wait times and multiple retries (via throttle errors). Not sure if this would be the same.
> 
> I think the developer should just set one time not 2 (maxLockSec). The wait should be whatever works best for us.

@casshill For GSheets, do we have custom logic around the retries? The current implementation will throw throttle errors if it can't acquire a lock, and will not wait. It's not a big lift to add a static wait time though.


> The config option is nice and simple. Would this restrict the lock to an entire perform task, not just a single function/http request? As long as we are fine to handle more/longer locks, as sometimes a perform function can do several things/make several requests and not all of them may need the lock

the config option will restrict to the entire perform task. the callback fn, is more flexible to lock specific http requests if there are many in a single perform

casshill:
>The current implementation will throw throttle errors if it can't acquire a lock, and will not wait. 

That sounds similar to Sheets. We originally had a fairly long wait before throwing a throttle error, but removed it in favour of a much shorter wait (3 sec). 
Sheets currently throws the throttle error with the retry time set for 15-25sec. The lock expiry is 55s

We did use throttles to handle the retries yes. That was the only available option we knew of at the time. 


What happens to the existing shouldLock config style option if we move forward with z.lock (instead?)

Perhaps if we went with a config option we could provide the function to be locked (instead of restricting to the whole perform run)?

standielpls:
> What happens to the existing shouldLock config style option if we move forward with z.lock (instead?)

Are you referring to [shouldLock config style option](https://zapier.github.io/zapier-platform-schema/build/schema.html#createschema)? It's currently a boolean and we won't make changes to that. 

The config style option in discussion is the alternative to the z.lock, both should provide the capability, but there some advantage of using config style for the future. This particular topic is being discussed [here](https://zapier.slack.com/archives/C04NRVD413K/p1690837078879209?thread_ts=1690547024.125629&cid=C04NRVD413K).

> Perhaps if we went with a config option we could provide the function to be locked (instead of restricting to the whole perform run)?

@kola-er I don't think we can support an individual function, wdyt?

casshill:
> Are you referring to [shouldLock config style option](https://zapier.github.io/zapier-platform-schema/build/schema.html#createschema)?
Yes. A few apps leverage that already. But it causes a huge error rate and is not fit for purpose. I don't believe it uses throttle errors, so most tasks waiting for the lock just timeout, making those Zaps unreliable. 

I'd recommend either fixing or removing that, ideally in parallel to the `z.lock` implementation.

standielpls:
@casshill we've decided to move away from `z.lock` and just support the config style option. I invite you to share your thoughts and feedback [here](https://coda.io/d/Team-Developer-Platform_di0MgBhlCWf/Support-Locking-in-the-CLI_sufKD#_lu2LX). Thank you

FokkeZB:
@standielpls thanks you for helping us learn fast through code, and realize we may need a different approach!

PR #682, opened 2023-08-01
docs(cli): Move up template explanation in README

xdhmoore:
Move up the explanation about templates to where user encounters them.


eliangcs (approved):
Good catch, thanks!

PR #689, opened 2023-08-04
SUPPORTENG-929_Adding detail on encoding default

marinahand:
as suggested here https://zapier.slack.com/archives/C04P5J5R8E4/p1689900123007759 

also added detail for callback testing, per https://zapier.slack.com/archives/C5Z9BP4U9/p1691168493239769?thread_ts=1691167523.978609&cid=C5Z9BP4U9


rnegron (approved):
Thanks for including the testing callback change @marinahand!

PR #690, opened 2023-08-04
fix(cli): Remove logic converting empty array to undefined in changelog metadata

kreddlear:
Currently, if either an App type update or an Issue type update is present, the promotion process expects **both** to be present. If only one is present, the other is converted to `undefined` instead of an empty array. However, there is a `join` attempt on both values -- so if one is undefined, the code returns something like `TypeError: issueFeatureUpdates is not iterable`.

This blocks the developer from promoting and doesn't clearly state what they need to do to fix it, which is not ideal.

It looks like this logic was originally added to keep empty fields from being included in the changelog -- however, I was not able to find instances where that would actually happen. Either way, empty fields would be preferable to blocking developers from promoting if they only include one of an issue type update and an app type update.

### DIFF
--- a/packages/cli/src/tests/utils/changelog.js
+++ b/packages/cli/src/tests/utils/changelog.js
@@ -23,6 +23,9 @@ describe('changelog utils', () => {
           '5. New action! create/add_contact\n' +
           'However, we also addressed fixed open issues!\n' +
           '- Fix #123 and an issue with create/send_message\n\n' +
+          '## 2.0.1\n\n' +
+          '1. Action fix only\n' +
+          '2. Another action fix\n\n' +
           '## new and improved version! (v2.0.0)\n\n' +
           '* Fix some bugs.\n' +
           '* Major docs fixes.\n\n' +
@@ -42,8 +45,8 @@ describe('changelog utils', () => {
     it('should be forgiving on the markdown format', async () => {
       const { appMetadata, issueMetadata, changelog } =
         await changelogUtil.getVersionChangelog('0.0.1', appDir);
-      should(appMetadata).equal(undefined);
-      should(issueMetadata).equal(undefined);
+      should(appMetadata.length).equal(0);
+      should(issueMetadata.length).equal(0);
       changelog.should.equal('initial release\n\njust for internal testing');
     });
 
@@ -74,8 +77,8 @@ describe('changelog utils', () => {
     it('should not return metadata if it is not found', async () => {
       const { appMetadata, issueMetadata, changelog } =
         await changelogUtil.getVersionChangelog('1.0.0', appDir);
-      should(appMetadata).equal(undefined);
-      should(issueMetadata).equal(undefined);
+      should(appMetadata.length).equal(0);
+      should(issueMetadata.length).equal(0);
       changelog.should.equal('* Removing beta "label".\n* Minor docs fixes.');
     });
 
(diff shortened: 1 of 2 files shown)


pragmatic-zac (approved):

eliangcs (commented):
Good catch!

Since `appMetadata` and `issueMetadata` can no longer be `undefined`, the places that make such assumption should be changed too.

Here's one I found: [Line 120-129](https://github.com/zapier/zapier-platform/blob/66bb6770471b736d92e7d851e64c9e832c6fe01e/packages/cli/src/oclif/commands/promote.js#L120-L129) in `promote.js`:

```js
      if (
        !appFeatureUpdates &&
        !issueFeatureUpdates &&
        !appBugfixes &&
        !issueBugfixes
      ) {
        this.log(
          `No metadata was found in the changelog. Remember, you can associate the changelog with issues or triggers/actions.\n\n${metadataPromptHelper}`
        );
      }
```

should be changed to

```js
      if (
        _.isEmpty(appFeatureUpdates) &&
        _.isEmpty(issueFeatureUpdates) &&
        _.isEmpty(appBugfixes) &&
        _.isEmpty(issueBugfixes)
      ) {
        this.log(
          `No metadata was found in the changelog. Remember, you can associate the changelog with issues or triggers/actions.\n\n${metadataPromptHelper}`
        );
      }
```

There are others in `promote.js` like [line 93](https://github.com/zapier/zapier-platform/blob/66bb6770471b736d92e7d851e64c9e832c6fe01e/packages/cli/src/oclif/commands/promote.js#L93) and [116](https://github.com/zapier/zapier-platform/blob/66bb6770471b736d92e7d851e64c9e832c6fe01e/packages/cli/src/oclif/commands/promote.js#L116).

kreddlear:
Thank you for the feedback @eliangcs ! I updated those lines so that they'll check if the array is empty, instead of if the value is truthy or not. I checked around and don't see other problematic checks on that, so hopefully we are all good here 🤞 

Question for you -- do you think this is the best way of fixing this? Or, would changing the check and the `join` be better here? I went with this way because I could see others expecting this to be an array and creating similar bugs in the future, but open to alternatives, although I know this is a simple situation!

packages/cli/src/oclif/commands/promote.js
```diff
@@ -113,15 +113,15 @@ ${metadataPromptHelper}`);
           .filter(({ app_change_type }) => app_change_type === 'BUGFIX')
           .map(({ issue_id }) => `#${issue_id}`);
 
-      if (appBugfixes || issueBugfixes) {
+      if (!_.isEmpty(appBugfixes) || !_.isEmpty(issueBugfixes)) {
         this.log(`Bug fixes: ${[...appBugfixes, issueBugfixes].join(', ')}`);
       }
```
standielpls, line 118: Alternatively, does something like this fix the issue? @kreddlear 

```js
if (appBugfixes || issueBugfixes) {
  this.log(`Bug fixes: ${[...(appBugfixes ?? []), ...(issueBugfixes ?? [])].join(', ')}`);
}
```

kreddlear: @standielpls yes I think so! I was just playing with something like that, I just remembered the `??` operator in JS exists 😅 I asked ChangHung something similar above, but do you think that would be a cleaner solution? I'm leaning towards yes since it's a one-line change and would mean all those `isEmpty` checks don't have to apply.

standielpls: I think you're right! The `isEmpty` check is mostly important if we change from a `undefined` to `[]` so we should be able to omit those

standielpls (approved):
I didn't test this locally but I think this solution should work!

eliangcs (approved):
Assuming you tested it, this is cleaner. Thanks!

PR #693, opened 2023-08-14
PDE-4140 feat(cli): Integration title length must be at least 2 characters

pragmatic-zac:
This PR adds a minimum length constraint for an integration title when adding via zapier register.

### DIFF
--- a/packages/cli/src/oclif/commands/register.js
+++ b/packages/cli/src/oclif/commands/register.js
@@ -33,7 +33,10 @@ class RegisterCommand extends ZapierBaseCommand {
       );
     }
 
-    if ('title' in this.args && this.args.title.length < MIN_TITLE_LENGTH) {
+    if (
+      this.args.title !== undefined &&
+      this.args.title.length < MIN_TITLE_LENGTH
+    ) {
       throw new Error(
         `Please provide a title that is ${MIN_TITLE_LENGTH} characters or more.`
       );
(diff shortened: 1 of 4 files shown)


packages/cli/src/oclif/commands/register.js
```diff
@@ -241,7 +253,7 @@ RegisterCommand.args = [
   {
     name: 'title',
     description:
-      "Your integrations's public title. Asked interactively if not present.",
+      "Your integration's public title. Asked interactively if not present.",
```
rnegron, line 256: Good catch!

packages/cli/src/oclif/commands/register.js
```diff
@@ -28,6 +33,12 @@ class RegisterCommand extends ZapierBaseCommand {
       );
     }
 
+    if ('title' in this.args && this.args.title.length < MIN_TITLE_LENGTH) {
```
rnegron, line 36: @pragmatic-zac When running `zapier register` (without an argument, thus triggering this flow), I am getting: `Error: Cannot read properties of undefined (reading 'length')`. Using the `--debug` flag shows that ` zapier:register args are { title: undefined }` so it seems that checking for just `'title' in this.args` is not enough 😢

rnegron (changes requested):

packages/cli/src/oclif/commands/register.js
pragmatic-zac: That's a really helpful shell alias, thank you!

Made a slight adjustment and this is fixed in my manual tests now. Ready for another look whenever you're able 😄

rnegron:
@pragmatic-zac I am seeing some failed CI tests on the latest commit. Tried re-running them but they failed again so they don't seem to be flaky tests.

PR #695, opened 2023-08-21
feat(schema) - Add lock config

standielpls:
Project: https://coda.io/d/Team-Developer-Platform_di0MgBhlCWf/Technical-Decision-CLI-Capability-Locking_su-UM#_luO3f
Ticket: https://zapierorg.atlassian.net/browse/PDE-4238

Adds the optional `lock` config to the `BasicOperationConfigSchema`. 

---

### Testing Locally
1. Created a CLI app. 
2. Added `lock` config that looks like this
 ```
  lock: {
      key: 'fetch_lyrics',
      timeout: 50,
      scope: ['account'],
    },
```
3. Run local `zapier` validate (using the schema in this PR)
4. See that it passes the local validation
```
Validating project locally
No structural errors found during validation routine.
This project is structurally sound!
```
(Note fails the API check bc the schema is not updated in the platform version)

### CODE REVIEW

File: packages/schema/exported-schema.json [hunk 1]
```diff
@@ -695,6 +695,29 @@
...
           }
+        },
+        "lock": {
+          "description": "**INTERNAL USE ONLY**. Zapier uses this config for internal operation locking.",
+          "type": "object",
+          "required": ["key"],
+          "properties": {
+            "key": {
+              "description": "The key to use for locking. This should be unique to the operation.",
+              "$ref": "/KeySchema"
```
eliangcs, line 706: The lock key uses the double curly template syntax, such as `{{bundle.inputData.spreadsheet_id}}:{{bundle.inputData.worksheet_id}}`. But `KeySchema` [doesn't allow](https://github.com/zapier/zapier-platform/blob/936f3e56fc0909f49323adfa3e008ea3ee1d95e7/packages/schema/lib/schemas/KeySchema.js#L10) curly braces. Maybe it can just be a non-empty string without restrictions?

eliangcs (commented):
Looks good to me except for the `KeySchema` restriction. Not sure if @robgolding @FokkeZB want to take a look too?

packages/schema/exported-schema.json
robgolding: Agreed, we do the same in `RedirectRequestSchema.url` which also supports curly syntax for variable replacement.

packages/schema/exported-schema.json
```diff
@@ -695,6 +695,30 @@
...
+          "type": "object",
+          "required": ["key"],
+          "properties": {
+            "key": {
+              "description": "The key to use for locking. This should be unique to the operation.",
+              "type": "string",
+              "minLength": 1
+            },
+            "scope": {
+              "description": "The level at which an app's access is restricted to. \n              'user' - Locks based on user ids. \n              'auth' - Locks based on unique auth ids\n              'account' - Locks for all users under a single account. \n              You may also combine scopes.",
```
kola-er, line 710: Although I suggested removing `app` from the scope list, I think we should be transparent about what it means when we default to it when no scope is specified. And if referring to it in the description w/o being in the list could be confusing, perhaps we add it also.

kola-er (approved):

standielpls:
@kola-er @eliangcs I addressed the legacy scripting in this PR as well. Also, I wanted to make sure if we covered all the steps to make this "internal" on the zapier-platform side. And we'll continue to limit this on the backend

eliangcs (dismissed):
Looks good to me! @FokkeZB do you want to take a final look? Thanks!

packages/schema/lib/schemas/BasicOperationSchema.js
```diff
@@ -50,6 +50,33 @@ module.exports = makeSchema(
...
+            description: `The level at which an app's access is restricted to. By default, locks are scoped to the app. That is, all users of the app will share the same locks. If you want to restrict serial access to a specific user, auth, or account, you can set the scope to one or more of the following: 'user' - Locks based on user ids.  'auth' - Locks based on unique auth ids. 'account' - Locks for all users under a single account. You may also combine scopes.`,
+            type: 'array',
+            items: {
+              enum: ['user', 'auth', 'account'],
+              type: 'string',
+            },
+          },
+          timeout: {
+            description:
+              'The number of seconds to hold the lock before timing out. If not provided, will use the default set by the app.',
```
FokkeZB, line 75: We should make clear what happens on time out, because typically time outs result in errors, but here it means we no longer prevent other invokes to kick of the action with the same key and scope.

packages/schema/lib/schemas/BasicOperationSchema.js
```diff
@@ -50,6 +50,33 @@ module.exports = makeSchema(
...
+        required: ['key'],
+        properties: {
+          key: {
+            description:
+              'The key to use for locking. This should be unique to the operation.',
+            type: 'string',
+            minLength: 1,
+          },
+          scope: {
+            description: `The level at which an app's access is restricted to. By default, locks are scoped to the app. That is, all users of the app will share the same locks. If you want to restrict serial access to a specific user, auth, or account, you can set the scope to one or more of the following: 'user' - Locks based on user ids.  'auth' - Locks based on unique auth ids. 'account' - Locks for all users under a single account. You may also combine scopes.`,
```
FokkeZB, line 66: I find "The level at which an app's access is restricted to." and the term "access" confusing, but what follows very clear, so I suggest we drop that first sentence.

I also suspect/expect that `app` would always be included as scope, otherwise one integration could block parallel execution of another that happens to use the same scope and e.g. `account` scope. That isn't clear now.

FokkeZB (approved):
Left a few comments on the descriptions. Even though this is internal only for now, let's try to make this clear from the start. Perhaps involve the platform support for platform doc team?

packages/schema/docs/build/schema.md
```diff
@@ -1141,6 +1145,35 @@ A unique identifier for this item.
 
 -----
 
+## /LockObjectSchema
```
kola-er, line 1148: Gave the lock object its own schema to make the description of its fields visible in this file. Also, it's easier to provide examples and anti-examples here than nesting it in the BasicOperationSchema.

packages/schema/lib/schemas/LockObjectSchema.js
```diff
@@ -0,0 +1,56 @@
...
+module.exports = makeSchema({
+  id: '/LockObjectSchema',
+  description:
+    '**INTERNAL USE ONLY**. Zapier uses this config for internal operation locking.',
+  type: 'object',
+  required: ['key'],
+  properties: {
+    key: {
+      description:
+        'The key to use for locking. This should be unique to the operation. While actions of different integrations with the same key and scope will never lock each other out, actions of the same integration with the same key and scope will do. User data provided for the input fields can be used in the key with the use of the curly braces referencing. For example, to access the user data provided for the input field "test_field", use `{{bundle.inputData.test_field}}`. Note that a required input field should be referenced to get user data always.',
```
kola-er, line 14: @FokkeZB The descriptions for the lock object fields have been updated to cater to your feedback. Please, review it again. Thanks.

kola-er:
@eliangcs and @stan Please review the new changes. Thanks.

FokkeZB (approved):

eliangcs (approved):

PR #703, opened 2023-09-15
docs(cli): update z.cursor docs for how to handle reaching the end of the result set

jasonkotenko:
Small syntax improvement in the docs example - the previous example would throw an error if `response.nextPage` was `null` which would commonly be the case.


rnegron (approved):
looks good to me! I ran `yarn docs` in `packages/cli` to propagate the code snippet change in the compiled `html` files.

eliangcs (approved):

PR #716, opened 2023-10-19
SUPPORTENG-984 Correct the  'value' and 'sample' description in the  'FieldChoicesWithLabel' schema

Osarhomiyeke:


packages/schema/lib/schemas/FieldChoiceWithLabelSchema.js
```diff
@@ -11,13 +11,13 @@ module.exports = makeSchema({
...
       description:
-        'The actual value that is sent into the Zap. Should match sample exactly.',
+        'The actual value that is sent into the Zap. This is displayed as light grey text in the editor. Should match sample exactly.',
       type: 'string',
       minLength: 1,
     },
     sample: {
       description:
-        "Displayed as light grey text in the editor. It's important that the value match the sample. Otherwise, the actual value won't match what the user picked, which is confusing.",
+        "It's important that the value match the sample. Otherwise, the actual value won't match what the user picked, which is confusing.",
```
eliangcs, line 20:
```suggestion
        "A legacy field that is no longer used by the editor, but it is still required for now and should match the value.",
```

eliangcs (approved):

PR #721, opened 2023-11-06
feat(schema): Add steadyState property on FieldSchema

kola-er:
Adds an optional property `steadyState` on the `FieldSchema` to be used with outputFields of polling triggers.

Ticket: https://zapierorg.atlassian.net/browse/PDE-4362

### DIFF
--- a/packages/schema/lib/schemas/FieldSchema.js
+++ b/packages/schema/lib/schemas/FieldSchema.js
@@ -121,7 +121,7 @@ module.exports = makeSchema(
         type: 'boolean',
       },
       steadyState: {
-        description: 'Prevents triggering on new output until all values for fields with this property remain unchanged for 2 polls. Can be used to, e.g., not trigger on a new contact until the contact has completed typing their name.',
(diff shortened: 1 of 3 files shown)


packages/schema/lib/schemas/FieldSchema.js
```diff
@@ -120,6 +120,10 @@ module.exports = makeSchema(
           'Does the value of this field affect the definitions of other fields in the set?',
         type: 'boolean',
       },
+      steadyState: {
+        description: 'Prevents triggering on new output until all values for fields with this property remain unchanged for 2 polls. Can be used to, e.g., not trigger on a new contact until the contact has completed typing their name.',
```
standielpls, line 124: Do you think it would be worthwhile to include a note about how this only applies to "Polling Triggers"?

kola-er: Yes, it would be helpful. Will include it.

kola-er: Added

standielpls (approved):

eliangcs (approved):

PR #730, opened 2023-11-27
PDE-4495 fix(cli): `zapier convert` crashes due to syntax error in user's code and should not replace `source` in sample

eliangcs:
Fixes two bugs in the `zapier convert` command.

Before (current behavior):

1. When there's a syntax error in user's code (the code the developer wrote in Code Mode), `zapier convert` would fail because we didn't catch the error raised from Prettier, the code formatter used in `zapier convert`.
2. When a sample has a field named `source` (there's an example in the test code), `zapier convert` would incorrectly replace it with a function, which then would cause Prettier to error.

After (expected behavior):

1. When there's a syntax error in user's code, `zapier convert` will print an warning and carry on the conversion. It will leave the bad code as is without prettifying. 
2. When a sample has a field named `source`, `zapier convert` won't replace it with a function.

### DIFF
--- a/packages/cli/src/tests/utils/convert.js
+++ b/packages/cli/src/tests/utils/convert.js
@@ -258,16 +258,25 @@ const setupTempWorkingDir = () => {
 };
 
 describe('convert', () => {
-  let tempAppDir, readTempFile;
+  let tempAppDir, readTempFile, origConsoleWarn;
+  const warnings = [];
 
   beforeEach(() => {
     tempAppDir = setupTempWorkingDir();
     readTempFile = (fpath) =>
       fs.readFileSync(path.join(tempAppDir, fpath), 'utf-8');
+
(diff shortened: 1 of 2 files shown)


packages/cli/src/tests/utils/convert.js
```diff
@@ -386,5 +395,73 @@ describe('convert', () => {
...
+      warnings[0].should.containEql('hydrators.js');
+
+      const hydratorsFile = readTempFile('hydrators.js');
+      should(
+        hydratorsFile.includes('getMovieDetails = async (z, bundle)')
+      ).be.true();
+      should(hydratorsFile.includes('{ bad code }')).be.true();
+    });
+
+    it("should not replace 'source' in sample", async () => {
```
rnegron, line 449: Great tests! 💯

rnegron (approved):

PR #735, opened 2024-01-29
PDE-4677 fix(core): RawHttpResponse method types

### LINKED ISSUE #734: [Bug]: z.request when raw is true does not include buffer method on response object
### Bug Description

I'm not sure if this is an issue with your TypeScript type definitions or with the actual library, but contrary to the examples and error messages you receive, there is no `buffer()` method on `RawHttpResponse` or `BaseHttpResponse`.

[The following example is invalid, at least for TS](https://platform.zapier.com/reference/cli-docs#http-response-object):
image

### Reproduction Steps

```typescript
import { ZObject } from 'zapier-platform-core' // 15.5.1 per package.json

const file = await z.request({
    url: download.item,
    raw: true,
})

// src/Hydrators.ts(34,44): error TS2339: Property 'buffer' does not exist on type 'RawHttpResponse'.
const buffer = await file.buffer() // CHOKE
```

### Zapier Platform version

15.5.1

### Node.js version

18.19.0

### Your Operating System

macOS 14.2.1

### npm/yarn version

10.2.3

### App ID

198292

### More Details

_No response_

eliangcs:
Fixes #734, where the response of `z.request({ raw: true })` doesn't have correct method typings for TypeScript.

## Example TypeScript Code

```typescript
import { Bundle, ZObject } from 'zapier-platform-core';

const getJson = async (z: ZObject): Promise<object> => {
    url: 'https://httpbin.zapier-tooling.com/get',
    raw: true,
  });
  return response.json();
};

const getBuffer = async (z: ZObject): Promise<Buffer> => {
    url: 'https://httpbin.zapier-tooling.com/image/png',
    raw: true,
  });
  return response.buffer();
};

const getText = async (z: ZObject): Promise<string> => {
    url: 'https://httpbin.zapier-tooling.com/status/418',
    raw: true,
    skipThrowForStatus: true,
  });
  return response.text();
};

const getContent = async (z: ZObject): Promise<Buffer> => {
    url: 'https://httpbin.zapier-tooling.com/robots.txt',
    raw: true,
  });
  return response.content;
};
```

## Current Behavior

When running the `tsc` transpiler, `response.buffer()`, `response.json()`, and `response.text()` report errors. `response.content`, on the other hand, _should_ report an error, but passes instead.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -115,9 +115,10 @@ export interface HttpResponse extends BaseHttpResponse {
 }
 
 export interface RawHttpResponse extends BaseHttpResponse {
-  content: Buffer;
-  json: Promise<object | undefined>;
   body: NodeJS.ReadableStream;
+	buffer(): Promise<Buffer>;
+	json(): Promise<object>;
+	text(): Promise<string>;
 }
 
 type DehydrateFunc = <T>(


packages/core/index.d.ts
```diff
@@ -115,9 +115,10 @@ export interface HttpResponse extends BaseHttpResponse {
 }
 
 export interface RawHttpResponse extends BaseHttpResponse {
-  content: Buffer;
```
eliangcs, line 118: `RawHttpResponse.content` isn't a valid property and shouldn't be here. When used, it would [throw a runtime error](https://github.com/zapier/zapier-platform/blob/cba83715c07a57f8357c476a53d4c0b69d15e4e1/packages/core/src/http-middlewares/after/prepare-response.js#L30-L38).

rnegron (approved):

PR #736, opened 2024-01-29
[PDE-4118] fix(core): Use node-fetch types in BaseHttpResponse

### LINKED ISSUE #667: [Bug]: BaseHttpResponse.headers incorrectly typed
### Bug Description

At https://github.com/zapier/zapier-platform/blob/c0da67300d1b2091e7b83c74fccac575bfeb371b/packages/core/index.d.ts#L104 `BaseHttpResponse.headers` is typed as `{ [key: string]: string }` but I've found that this is a `node-fetch` `Headers` class object instead: https://github.com/node-fetch/node-fetch/blob/7b86e946b02dfdd28f4f8fca3d73a022cbb5ca1e/%40types/index.d.ts#L40

### Reproduction Steps

1. After getting a `response` with `z.request()`, try to read a header using `response.headers['content-type']` and find it returns `undefined`.
2. Try to use `response.headers.get('content-type')` and find TypeScript raises a type error.
3. Ignore it and find it does return the expected value (meaning the type is wrong).

### Zapier Platform version

14.1.1

### Node.js version

v18.15.0

### Your Operating System

13.4 (22F66)

### npm/yarn version

1.22.19

### App ID

185850

### More Details

_No response_

rnegron:
Fixes #667. As noted in the bug report, using `.headers.get()` returns a TypeScript error `TS2349: This expression is not callable. Type 'String' has no call signatures.`:

Screenshot 2024-01-29 at 1 06 39 PM

But testing it out (using vanilla JavaScript) shows that `.headers.get()` does work (first console.log is `response.headers['content-type']`, second is `response.headers.get('content-type')`:

```
console.log
  undefined

  at perform (lib/creates/movie.js:15:13)

console.log
  application/json; charset=utf-8

  at perform (lib/creates/movie.js:16:13)
```

### DIFF
--- a/docs/cli.html
+++ b/docs/cli.html
@@ -345,7 +345,7 @@ a code {
 </blockquote><p><strong>Usage</strong>: <code>zapier build</code></p><p>This command does the following:</p><ul>
 <li><p>Creates a temporary folder</p>
 </li>
-<li><p>Copies all code (excluding files in .gitignore) into the temporary folder</p>
+<li><p>Copies all code into the temporary folder</p>
 </li>
 <li><p>Adds an entry point: <code>zapierwrapper.js</code></p>
 </li>
--- a/docs/cli.html
+++ b/docs/cli.html
@@ -345,7 +345,7 @@ a code {
 </blockquote><p><strong>Usage</strong>: <code>zapier build</code></p><p>This command does the following:</p><ul>
 <li><p>Creates a temporary folder</p>
 </li>
-<li><p>Copies all code into the temporary folder</p>
+<li><p>Copies all code (excluding files in .gitignore) into the temporary folder</p>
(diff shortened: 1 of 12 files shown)


kreddlear (dismissed):

eliangcs (approved):
Looks good!

I didn't know `@types/node-fetch` exist. I wonder if we could just replace a lot of things, like `HttpRequestOptions`, `BaseHttpResponse`, and `RawHttpResponse` with node-fetch's.

rnegron:
@eliangcs I looked at replacing some of the other custom types with `node-fetch` equivalents but there were some subtle differences between them, such as custom methods (like `throwForStatus`)  and properties (like `removeMissingValuesFrom`).

PR #737, opened 2024-01-29
[PDE-4679] fix(cli): better handling of broken symlinks while copying files to temp directory during build process

### LINKED ISSUE #706: [Bug]: `zapier build` follows invalid symlinks when copying
### Bug Description

Hi team,

I'm running into an issue with my local development environment when running `zapier build`, it will fail on the first step:

```
✖ Copying project to temp directory
 ›   Error: ENOENT: no such file or directory, stat
 ›   'hi'
```

FWIW, this file is in my .gitingore.  It would be great if the filter for this step excluded those `.gitignore` files as well.  I see it does this on zipping.  Seems like this should occur much sooner.

To be honest, I am not sure why it needs to copy into a separate directory in the first place.

Thanks!

### Reproduction Steps

1. Create a symlink to a file that does not exist: `ln -s bye hi`
2. Run `zapier build`

### Zapier Platform version

15.3.0

### Node.js version

v18.17.1

### Your Operating System

macOS 13.5.2 (22G91)

### npm/yarn version

_No response_

### App ID

_No response_

### More Details

_No response_

rnegron:
TO-DO: This is still missing some tests.

Intended to fix #706. Modifies the `copyDir` method by using `lstat` to avoid crashing when reading a faulty symbolic link.

About `lstat` from the [manual pages](https://man7.org/linux/man-pages/man2/lstat.2.html): 

Screenshot 2024-01-30 at 11 25 44 AM

Example of new warning message:

Screenshot 2024-01-31 at 4 33 22 PM

### DIFF
--- a/packages/cli/src/utils/files.js
+++ b/packages/cli/src/utils/files.js
@@ -87,7 +87,7 @@ const copyFile = (src, dest, mode) => {
   Returns a promise that copies a directory recursively.
 
 	Options:
-		
+
 	- clobber: Overwrite existing files? Default is false.
 	- filter:
 			A function that returns true if the file should be copied. By default, it
--- a/packages/cli/src/utils/files.js
+++ b/packages/cli/src/utils/files.js
@@ -128,9 +128,26 @@ const copyDir = async (src, dst, options) => {
 
   const promises = files.map(async (file) => {
     const srcItem = path.resolve(src, file);
-    const srcStat = fse.lstatSync(srcItem);
+
+    let srcStat;
+    try {
+      srcStat = fse.statSync(srcItem);
+    } catch (err) {
+      // If the file is a symlink and the target doesn't exist, skip it.
+      if (fse.lstatSync(srcItem).isSymbolicLink()) {
+        console.warn(
+          colors.yellow(
+            `\n! Warning: symlink "${srcItem}" points to a non-existent file. Skipping!\n`
+          )
+        );
+        return null;
+      }
+
+      // otherwise, rethrow the error
+      throw err;
+    }
+
     const srcIsFile = srcStat.isFile();
-    const srcIsSymbolicLink = srcStat.isSymbolicLink();
 
     const dstItem = path.resolve(dst, file);
     const dstExists = fileExistsSync(dstItem);
@@ -138,17 +155,7 @@ const copyDir = async (src, dst, options) => {
       return null;
     }
 
-    if (srcIsFile || srcIsSymbolicLink) {
-      if (srcIsSymbolicLink && !fileExistsSync(srcItem)) {
-        console.warn(
-          colors.yellow(
-            `\n! Warning: symlink "${srcItem}" points to a non-existent file. Skipping!\n`
-          )
-        );
-        options.onSkip(dstItem);
-        return null;
-      }
-
+    if (srcIsFile) {
       if (dstExists) {
         if (!options.clobber) {
           options.onSkip(dstItem);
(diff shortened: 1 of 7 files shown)


packages/cli/src/utils/files.js
```diff
@@ -98,41 +98,41 @@ const copyDir = (src, dst, options) => {
...
-        }
+  await ensureDir(dst);
+  const files = await fse.readdir(src);
+
+  const promises = files.map(async (file) => {
+    const srcItem = path.resolve(src, file);
+
+    // skip srcItem if filter returns false
+    if (!options.filter(srcItem)) {
+      return Promise.resolve();
```
eliangcs, line 109: Since we're already in an async function, `Promise.resolve()` is unnecessary.

```suggestion
      return;
```

packages/cli/src/utils/files.js
```diff
@@ -98,41 +98,41 @@ const copyDir = (src, dst, options) => {
...
+    const dstItem = path.resolve(dst, file);
+    const stat = fse.statSync(srcItem);
+    const dstExists = fileExistsSync(dstItem);
+
+    if (dstExists && options.clobber) {
+      fse.removeSync(dstItem);
+    } else if (dstExists) {
+      if (!stat.isDirectory()) {
+        options.onSkip(dstItem);
+        return Promise.resolve();
```
eliangcs, line 121:
```suggestion
        return;
```

packages/cli/src/utils/files.js
```diff
@@ -98,41 +98,41 @@ const copyDir = (src, dst, options) => {
...
+      }
+    }
+
+    if (stat.isDirectory()) {
+      await ensureDir(dstItem);
+      return copyDir(srcItem, dstItem, options);
+    } else {
+      return copyFile(srcItem, dstItem, stat.mode).then(() => {
+        options.onCopy(dstItem);
       });
```
eliangcs, line 131: Should we refactor this one to use `await` too?

eliangcs (commented):
`zapier build` produces two files:
- `build/build.zip`
- `build/source.zip`.

`source.zip` already respects `.gitignore`. However, I don't think `build.zip` should exclude the files in `.gitignore`. When running `zapier build --skip-npm-install`, we [intentionally don't](https://github.com/zapier/zapier-platform/blob/f693a122582504a937d3f23d73a5ec40065f4a9c/packages/cli/src/utils/build.js#L350-L356) run `npm install` internally. Instead, it copies the files from the working directory's `node_modules`. So even though `node_modules` is in `.gitignore`, but we should include it in `builde.zip`.

I believe the problem with #706 is not about `.gitignore` but is about invalid symlinks. `zapier build` should show a better error message when it encounters a bad symlink. Let me know if I missed anything though!

rnegron:
@eliangcs 

> So even though node_modules is in .gitignore, we should include it in build.zip.

OK, this makes sense to me! I was wondering if just using `gitignore` was the correct approach or not. I will re-tool this PR to add some logic for detecting an invalid symlink and erroring appropriately.

rnegron:
Hmm, using `lstat` causes the `describe('build in workspaces')` tests to fail. Checking...

rnegron:
I was thinking that, in order to address the original issue, simply alerting the user that the problem was a broken symlink might be enough and we don't need to get too creative by differentiating between file or directories...

The question is then: should we skip the faulty symlink or error out of the build process?

eliangcs:
> The question is then: should we skip the faulty symlink or error out of the build process?

I think skipping with a very visible warning (like you're doing now) would be good.

This looks good to me. Your comment reminds me that we need to skip the symlinks that point to a workspace. I'll go ahead and merge this, so I can fix that in my PR. Thanks!

eliangcs (approved):

PR #738, opened 2024-01-30
PDE-4678 fix(cli): `zapier build` fails with npm workspaces

### LINKED ISSUE #648: [Bug]: zapier push fails with npm workspaces
### Bug Description

Executing `zapier push` fails with an error: 
```
Looks like you're missing a local installation of zapier-platform-core. Run `npm install` to resolve
```

### Reproduction Steps

1. in the package.json at the root of the monorepo, add a folder `/my-adapter` and include it in the list of workspaces (see [npm workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces))
2. from the monorepo root, install zapier packages: `npm install zapier-platform-cli zapier-platform-core --workspace my-adapter`
3. implement the code for a zapier adapter in `/my-adapter`
4. `cd ./my-adapter && npm run build && npx zapier push` throws an error complaining about missing local installation

### Zapier Platform version

14.0.0

### Node.js version

18.15.0

### Your Operating System

macos

### npm/yarn version

9.6.3

### App ID

160973

### More Details

_No response_

eliangcs:
Fixes #648.

Some developers may choose to structure their project using [npm workpaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) or [yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/). An example project structure might look like:

```
(project root)
├─ package.json
└── packages/
    ├─ app-1/
    │  ├─ index.js
    │  └─ package.json
    └─ app-2/
        ├─ index.js
        └─ package.json
```

If you run `npm install` in the above example, `app-1` and `app-2` will share their dependencies in the project root's `node_modules`. Only the dependencies with version discrepancies will be placed into `app-1/node_modules` or `app-2/node_modules`.

The current implementation of `zapier build` doesn't support workspaces. It always assume it can always find `zapier-platform-core` in the working directory's `node_modules`.

This PR fixes that by finding `zapier-platform-core` in the working directory _and_ all the parent directories.

### DIFF
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -45,7 +45,7 @@ const {
 
 const checkMissingAppInfo = require('./check-missing-app-info');
 
-const { runCommand, isWindows } = require('./misc');
+const { runCommand, isWindows, findCorePackageDir } = require('./misc');
 
 const debug = require('debug')('zapier:build');
 
@@ -332,6 +332,7 @@ const _buildFunc = async ({
     osTmpDir,
     'zapier-' + crypto.randomBytes(4).toString('hex')
   );
+  debug('Using temp directory: ', tmpDir);
 
   maybeNotifyAboutOutdated();
 
@@ -342,9 +343,28 @@ const _buildFunc = async ({
   await ensureDir(constants.BUILD_DIR);
 
   startSpinner('Copying project to temp directory');
-  await copyDir(wdir, tmpDir, {
-    filter: skipNpmInstall ? (dir) => !dir.includes('.zip') : undefined,
-  });
+
+  const copyFilter = skipNpmInstall
+    ? (src) => !src.includes('.zip')
+    : undefined;
+
+  await copyDir(wdir, tmpDir, { filter: copyFilter });
+
+  if (skipNpmInstall) {
+    const corePackageDir = findCorePackageDir();
+    const nodeModulesDir = path.dirname(corePackageDir);
+    const workspaceDir = path.dirname(nodeModulesDir);
+    if (wdir !== workspaceDir) {
+      // If we're in here, it means the user is using npm/yarn workspaces
+      await copyDir(nodeModulesDir, path.join(tmpDir, 'node_modules'), {
+        filter: copyFilter,
+        onDirExists: (dir) => {
+          // Don't overwrite existing sub-directories in node_modules
+          return false;
+        },
+      });
+    }
+  }
 
   let output = {};
   if (!skipNpmInstall) {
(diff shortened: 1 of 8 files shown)


eliangcs:
@rnegron this is still work-in-progress as I'd like to add some test cases. But you might want to take a look first, because I also modified `copyDir()` as you did in #737.

rnegron:
@eliangcs 

> But you might want to take a look first, because I also modified copyDir() as you did in https://github.com/zapier/zapier-platform/pull/737.

Thanks for letting me know! Let's use this PR for those initial `async/await` changes, since I will likely modify it further to add some `fs.lstatSync(...).isSymbolicLink()` related logic!

packages/cli/src/utils/files.js
```diff
@@ -83,60 +83,87 @@ const copyFile = (src, dest, mode) => {
...
 };
 
-// Returns a promise that copies a directory.
-const copyDir = (src, dst, options) => {
-  const defaultFilter = (dir) => {
-    const isntPackage = dir.indexOf('node_modules') === -1;
-    const isntBuild = dir.indexOf('.zip') === -1;
+/*
+  Returns a promise that copies a directory recursively.
+
```
rnegron, line 88: 👍🏻  Always appreciate these descriptive comments

rnegron (dismissed):

packages/cli/src/utils/build.js
```diff
@@ -383,9 +414,7 @@ const _buildFunc = async ({
     path.join(tmpDir, 'zapierwrapper.js'),
     zapierWrapperBuf.toString()
   );
-  endSpinner();
 
-  startSpinner('Building app definition.json');
```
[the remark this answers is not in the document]
eliangcs: It was a mistake. Added it back in [52fba09](https://github.com/zapier/zapier-platform/commit/52fba09b195add7cd30fa7ed19efeb8775fbf31c). Thanks!

eliangcs:
@rnegron I merged your PR #737 and added some code to exclude workspace symlinks, such as `node_modules/app-1` and `node_modules/app-2`, from the `build.zip`. Can you re-review? Thanks a lot!

packages/cli/src/utils/build.js
```diff
@@ -332,24 +353,69 @@ const _buildFunc = async ({
...
+            return false;
+          }
+          const stat = fse.lstatSync(src);
+          if (stat.isSymbolicLink()) {
+            const realPath = path.resolve(
+              path.dirname(src),
+              fse.readlinkSync(src)
+            );
+            for (const workspace of workspaces) {
+              if (minimatch(realPath, workspace)) {
```
rnegron, line 398: @eliangcs Could we add a comment here explaining what this `minimatch` line is doing? I believe it's `true` if the `src` symlink resolves (points to) to a workspace path, in which case we want to filter out (`return false`) this particular `src`. So that we don't copy over any symlinks that point to workspace folders.

rnegron (approved):

packages/cli/src/utils/build.js
eliangcs: You interpretation is correct. Since the `workspaces` array in the root `package.json` can contain glob patterns, like `packages/*`, I use minimatch to help us do glob pattern matching. I'll add a comment in a separate PR. Thanks!

PR #739, opened 2024-01-30
[PDE-4682] chore(cli): Add @clif/core dependency, upgrade @oclif/plugin-help

rnegron:
As mentioned in https://github.com/zapier/zapier-platform/pull/556#issuecomment-1171944967, simply upgrading `@oclif/plugin-help` will break CLI functionality due to certain helper methods we were using in `ZapierBaseCommand` being removed. In fact, they were moved over to the `@oclif/core` library.

This PR adds the `@oclif/core` dependency, which is also the [first step in migrating over](https://github.com/oclif/core/blob/main/guides/PRE_CORE_MIGRATION.md) from deprecated helper libraries.

A visual difference in how text is formatted can be observed before and after this PR. I'm looking into why this formatting change occurred to see if we can maintain the previous formatting style.

Before (wider terminal):

Screenshot 2024-01-30 at 3 55 57 PM

After (wider terminal): 

Screenshot 2024-01-30 at 3 55 47 PM

Before (smaller terminal):

Screenshot 2024-01-30 at 3 56 07 PM

After (smaller terminal):

Screenshot 2024-01-30 at 3 56 17 PM

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -40,8 +40,9 @@
   "dependencies": {
     "@oclif/command": "1.8.27",
     "@oclif/config": "1.18.10",
+    "@oclif/core": "1.26.0",
     "@oclif/plugin-autocomplete": "0.3.0",
-    "@oclif/plugin-help": "3.2.2",
+    "@oclif/plugin-help": "3.2.20",
     "@oclif/plugin-not-found": "1.2.4",
     "adm-zip": "0.5.10",
     "archiver": "5.3.1",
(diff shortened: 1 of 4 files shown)


rnegron:
Opening for review because I'm not sure it's worth the time to implement changes to make the `renderList` display as before. The display does look uglier... but I think the tradeoff of closing #556 is worth it. 

The reason why the display changes is because the `renderList` implementation in `@oclif/core` is different: https://github.com/oclif/core/blob/10b7dc0dd207906df208b0c5b85e309106dd8a5f/src/cli-ux/list.ts#L15, it does not include additional properties like the one from `@oclif/plugin-help` (which no longer exists after the upgrade).

eliangcs (approved):
@rnegron I found instead of importing `renderList` from `@oclif/core/lib/cli-ux/list`, we should it from `@oclif/help/lib/list`. That way we can keep using the same `renderList` with previous options. Turns out `@oclif/plugin-help` was not gone but was renamed to `@oclif/help`.

I created #743 to fix. Otherwise, this PR looks good to me. Feel free to find someone else to approve. Thanks!

PR #742, opened 2024-02-02
PDE-4678 fix(cli): add comments and fix a path issue in `utils/build.js`

eliangcs:
Following up with this [comment](https://github.com/zapier/zapier-platform/pull/738#discussion_r1475061941), this PR adds some comments in the workspace handling code in `utils/build.js`.

Also replaces `path.join(workspaceRoot, relpath)` with `path.resolve(workspaceRoot, relpath)` just in case someone uses an absolute path in `<package_json>.workspaces` and `relpath` is an absolute path.

### DIFF
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -319,6 +319,8 @@ const maybeRunBuildScript = async (options = {}) => {
   }
 };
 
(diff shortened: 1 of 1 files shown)


rnegron (approved):
Thank you for circling back on this!

PR #752, opened 2024-03-01
[PLATSUP-1022] Document `allowGetBody` option in `z.request()` options

iscorer:
Document `allowGetBody` option as discussed and confirmed in this Slack thread: 

[Hey team - any reasons why we shouldn’t document the allowGetBody option on z.request() that was originally implemented here and is mentioned in the CHANGELOG?](https://zapier.slack.com/archives/C04P5J5R8E4/p1709091636494479)


eliangcs:
Thanks! Merging.

PR #754, opened 2024-03-18
PDE-4685 feat(core,schema): allow to set `primary` for outputFields

eliangcs:
After this PR, a developer can define which output field(s) to be the primary key of an output item. For example:

```js
{
  triggers: {
    recipe: {
      operation: {
        outputFields: [
          { key: 'authorId', primary: true },
          { key: 'name', primary: true }
        ]
      }
    }
  }
}
```

will tell Zapier to [dedupe](https://platform.zapier.com/build/deduplication) polling trigger results using the unique hash of `(authorId, name)`.

If no output fields are set as `primary`, the field with the key `id` will be as the primary key for deduplication. This is the same as the current behavior.


kola-er (dismissed):
Thanks for the thorough job.

packages/core/src/checks/trigger-has-id.js
```diff
@@ -10,9 +10,46 @@ const isTrigger = require('./is-trigger');
...
+    for (const field of outputFields) {
+      if (!field) {
+        continue; // just in case
+      }
+      if (field.primary) {
+        if (field.key === 'id') {
+          return true;
+        } else {
+          hasPrimary = true;
+          break;
```
kola-er, line 48: What if the loop is yet to process the `id` field with primary set? I don't think the outputFields is always sorted to start with the `id` field.

eliangcs: You're right. Reverted the change and added a test in 7f5ef179. Thanks!

kola-er (approved):

PR #757, opened 2024-03-21
feat(schema): Support "key" in throttle schema and the scope of "action"

kola-er:
feat(schema): Support "key" in throttle schema and the scope of "action"

### DIFF
--- a/packages/schema/docs/build/schema.md
+++ b/packages/schema/docs/build/schema.md
@@ -2067,7 +2067,7 @@ Key | Required | Type | Description
 
 ## /ThrottleObjectSchema
 
-Zapier uses this configuration to apply throttling when the limit for the window is exceeded.
+Zapier uses this configuration to apply throttling when the limit for the window is exceeded. **NOTE:** the final key used for the throttling is formed as a combination of the key, window, and scope. While actions of different integrations with the same key, window, and scope will never share the same limit, actions of the same integration with the same key, window, and scope will do when "action" is not in the scope.
 
 #### Details
 
@@ -2080,7 +2080,7 @@ Key | Required | Type | Description
 --- | -------- | ---- | -----------
 `window` | **yes** | `integer` | The timeframe, in seconds, within which the system tracks the number of invocations for an action. The number of invocations begins at zero at the start of each window.
 `limit` | **yes** | `integer` | The maximum number of invocations for an action, allowed within the timeframe window.
-`key` | no | `string` | The key to throttle with in combination with the scope. This should be unique to the operation. While actions of different integrations with the same key and scope will never share the same limit, actions of the same integration with the same key and scope will do when "action" is not in the scope. User data provided for the input fields can be used in the key with the use of the curly braces referencing. For example, to access the user data provided for the input field "test_field", use `{{bundle.inputData.test_field}}`. Note that a required input field should be referenced to get user data always.
+`key` | no | `string` | The key to throttle with in combination with the scope. User data provided for the input fields can be used in the key with the use of the curly braces referencing. For example, to access the user data provided for the input field "test_field", use `{{bundle.inputData.test_field}}`. Note that a required input field should be referenced to get user data always.
 `scope` | no | `array`[`string` in (`'user'`, `'auth'`, `'account'`, `'action'`)] | The granularity to throttle by. You can set the scope to one or more of the following: 'user' - Throttles based on user ids.  'auth' - Throttles based on auth ids. 'account' - Throttles based on account ids for all users under a single account. 'action' - Throttles the action it is set on separately from other actions. By default, throttling is scoped to the action and account.
 `overrides` | no | `array`[`object`] | EXPERIMENTAL: Overrides the original throttle configuration based on a Zapier account attribute.
 
--- a/packages/schema/docs/build/schema.md
+++ b/packages/schema/docs/build/schema.md
@@ -2067,7 +2067,7 @@ Key | Required | Type | Description
 
 ## /ThrottleObjectSchema
 
-Zapier uses this configuration to apply throttling when the limit for the window is exceeded. **NOTE:** the final key used for the throttling is formed as a combination of the key, window, and scope. While actions of different integrations with the same key, window, and scope will never share the same limit, actions of the same integration with the same key, window, and scope will do when "action" is not in the scope.
+Zapier uses this configuration to apply throttling when the limit for the window is exceeded. **NOTE:** The final key used for the throttling is formed as a combination of all the configurations; key, window, limit, and scope. To share a limit across multiple actions in an integration, each should have the same configuration set without "action" in the scope.
 
 #### Details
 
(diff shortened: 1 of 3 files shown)


packages/schema/docs/build/schema.md
```diff
@@ -2080,14 +2080,24 @@ Key | Required | Type | Description
...
 `limit` | **yes** | `integer` | The maximum number of invocations for an action, allowed within the timeframe window.
-`scope` | no | `array`[`string` in (`'user'`, `'auth'`, `'account'`)] | The granularity to throttle by. You can set the scope to one or more of the following: 'user' - Throttles based on user ids.  'auth' - Throttles based on auth ids. 'account' - Throttles based on account ids for all users under a single account. By default, throttling is scoped to the account.
+`key` | no | `string` | The key to throttle with in combination with the scope. This should be unique to the operation. While actions of different integrations with the same key and scope will never share the same limit, actions of the same integration with the same key and scope will do when "action" is not in the scope. User data provided for the input fields can be used in the key with the use of the curly braces referencing. For example, to access the user data provided for the input field "test_field", use `{{bundle.inputData.test_field}}`. Note that a required input field should be referenced to get user data always.
```
standielpls, line 2083: Is it worth mentioning what happens if the key is not provided?

packages/schema/lib/schemas/ThrottleObjectSchema.js
```diff
@@ -19,11 +19,17 @@ module.exports = makeSchema({
         'The maximum number of invocations for an action, allowed within the timeframe window.',
       type: 'integer',
     },
+    key: {
+      description:
+        'The key to throttle with in combination with the scope. This should be unique to the operation. While actions of different integrations with the same key and scope will never share the same limit, actions of the same integration with the same key and scope will do when "action" is not in the scope. User data provided for the input fields can be used in the key with the use of the curly braces referencing. For example, to access the user data provided for the input field "test_field", use `{{bundle.inputData.test_field}}`. Note that a required input field should be referenced to get user data always.',
```
standielpls, line 24: Can the user _not_ provide a key?

standielpls: What happens in the scenario where a user uses the same key in two actions, but has different values?

For example:
```
update_row: {
  ...
  throttle: {
    key: 'row_level_action',
    limit: 50,
    window: 60,
    scope: ['account']
  }
}

delete_row: {
  ...
  throttle: {
    key: 'row_level_action',
    limit: 1000,
    window: 60,
    scope: ['account']
  }    
}
```

packages/schema/docs/build/schema.md
FokkeZB: Might be helpful, though implied with "in combination with the scope"

FokkeZB: That may be helpful, although implied with "in combination with the scope".

Perhaps even more helpful would be to explain `key` and `scope` together in the description of the schema instead of here individually in the props.

We could explain that throttling will be shared by all actions with the same `key` and that the values referenced by `scope` are mixed into it, and that `scope` defaults to `['account']`.

Pseudo code may also help here:

```ts
const usedKey = (scope || ['account'])
  .map(s => values[s])
  .concat(key, app)
  .join('-');
```

Same likely applies to `LockObjectSchema`.

packages/schema/lib/schemas/ThrottleObjectSchema.js
FokkeZB: Good question!

Looking at https://gitlab.com/zapier/zapier/-/blob/fb1bdfa5dbb019ddac4a599e2d303c13cd3a712d/web/backend/zapier/useful/throttle_tools.py#L133-135

- `window` is mixed into the `key`, so if those don't match they won't affect each other. That should be called out in the above description (comment).
- `limit` is applied per action, so if in your example `window` would be identical, then within that window `update_row` would throttle at 50, but `delete_row` would happily continue up to 1000. 🤔

FokkeZB (approved):
Approving, but left a comment to improve docs, And @standielpls raised an excellent point that also should be documented or perhaps considered as change (dropping `window` from the key?)

packages/schema/docs/build/schema.md
kola-er: Updated the descriptions by moving the implication of same key, window, and scope to the schema description.

packages/schema/lib/schemas/ThrottleObjectSchema.js
standielpls: I see - I think there's 3 ways to go about this then:
1. Remove `window` from key as Fokke suggested
2. Add `limit` to the key as Kola suggested
3. Add a validation assertion to this schema: if a throttle key is defined, the limit and key must be equal and when the user runs `zapier validate` they will see a validation error.

1 and 2 carry the risk of unintentional behaviour, where a user may report that they thought their action will throttle at X/Y but is actually doing A/B. But we can be very loud about documenting this behaviour to try and mitigate this.

While 3 carries more work and effort, I think this will lead to implementing the original intent behind adding a throttle key. 

I'm fine with any - if 1/2, then I suggest adding a clear message and maybe a snippet of the above example
@FokkeZB @kola-er wdyt?

kola-er: Aside from supporting the new `key`, the main goal is to support the sharing of the same limit across multiple actions without causing unintended behavior. The scenario you painted was helpful to see the possibility of unintended behavior. However, I think validating would be restrictive and could be limiting the devs for some other intended behaviors. As for the option of removing the `window` from being mixed into the final key, it would still result to both limits of `50` and `1000` being used but not both in the same timeframe.

Adding the `limit` to the final key, in addition to the `window` that is already mixed into it, we can have a description with less confusion like this;

```
The final key used for the throttling is formed as a combination of all the configurations; key, window, limit, and scope. To share a limit across multiple actions, each should have the same configuration set without "action" in the scope.
```

This way, devs know exactly how they need to set the configuration to share a limit across actions without unintended behavior.

kola-er: Btw, this PR can be merged and the case of unintended behavior can be treated in a separate release to not delay the [15.6.0 release](https://github.com/zapier/zapier-platform/pull/758).

standielpls: That seems reasonable, thanks for discussing this!

standielpls (dismissed):

packages/schema/lib/schemas/ThrottleObjectSchema.js
standielpls: I am happy to move forward with adding the `limit` value to the final key

kola-er: Okay, will include it in the backend code and update the docs here too. Thanks @standielpls

kola-er: @standielpls Updated the description to note that all the configurations are combined to form the final key in the last commit.

standielpls (approved):

packages/schema/lib/schemas/ThrottleObjectSchema.js
FokkeZB: This makes sense. Thanks all.

Also see https://gitlab.com/zapier/zapier/-/merge_requests/56351#note_1830368072

PR #761, opened 2024-03-26
PDE-4844 - feat(schema) Add new "retry" in the throttle.overrides field

standielpls:
Adds a new field `retry` to the throttle.overrides schema to allow users to configure the effect throttling has on the action's tasks.


eliangcs (approved):

kola-er (commented):
Looks good. Let's also add the changes to the CLI docs [here](https://github.com/zapier/zapier-platform/blob/2e206c36cdaf39f3f1d42776b35382a9d32c1d33/packages/cli/README-source.md?plain=1#L1172) and [here](https://github.com/zapier/zapier-platform/blob/2e206c36cdaf39f3f1d42776b35382a9d32c1d33/packages/cli/snippets/throttle-configuration.js#L28-L44)

eliangcs (approved):

PR #767, opened 2024-04-05
PDE-4878 docs(cli): Add missing 'to' in cli docs

pragmatic-zac:
Tiny update for missing word in CLI docs.

Does not allow users enter data. -> Does not allow users **to** enter data.


rnegron (approved):
Thanks 😄

PR #768, opened 2024-04-05
[PDE-4869] feat(core) Raise max file size for upload streaming to 1GB

kola-er:


standielpls (approved):

packages/core/src/tools/create-file-stasher.js
```diff
@@ -295,6 +306,8 @@ const createFileStasher = (input) => {
       filename: _filename,
     } = await resolveToBufferStringStream(responseOrData);
 
+    isUploadMaxSizeExceeded(streamOrData, knownLength || length);
```
eliangcs, line 309: `knownLength || length` is repeated. Can we assign it to a variable:

```suggestion
    const finalLength = knownLength || length;
    isUploadMaxSizeExceeded(streamOrData, finalLength);
```

and reuse it below too?

```js
    return uploader(
      signedPostData,
      streamOrData,
      finalLength,   // <--
      filename || _filename,
      contentType || _contentType
    );
```

packages/core/src/tools/create-file-stasher.js
```diff
@@ -237,6 +235,19 @@ const uploader = async (
...
+const isUploadMaxSizeExceeded = (streamOrData, length) => {
+  let uploadMaxSize = NON_STREAM_UPLOAD_MAX_SIZE;
+  let uploadMethod = 'non-streaming';
+  if (streamOrData instanceof Readable) {
+    uploadMaxSize = UPLOAD_MAX_SIZE;
+    uploadMethod = 'streaming';
+  }
+
+  if (length && length > uploadMaxSize) {
+    throw new Error(`${length} is too big, ${uploadMaxSize} is the max for ${uploadMethod} of data.`);
```
eliangcs, line 247: That "of" sounds redundant. We might want to add the size unit too so it's more clear.

```suggestion
    throw new Error(`${length} bytes is too big, ${uploadMaxSize} is the max for ${uploadMethod} data.`);
```

packages/core/src/tools/create-file-stasher.js
```diff
@@ -237,6 +235,19 @@ const uploader = async (
   throw new Error(`Got ${response.status} - ${response.content}`);
 };
 
+const isUploadMaxSizeExceeded = (streamOrData, length) => {
```
eliangcs, line 238: This function name is good if it just returns a boolean. But it can actually throw an error. So I'd suggest we rename it to something like `ensureUploadMaxSizeNotExceeded`.

eliangcs (commented):
I did some tests using the debugger and the existing [test suite](https://github.com/zapier/zapier-platform/blob/5fcb3056d88a0d127aec276b53b6bf960c4d01b3/packages/core/test/tools/file-stasher.js). The code seems to work great! 👍 

I have some minor suggestions. Once they're fixed, I think this will be good to merge. Thanks!

packages/core/src/tools/create-file-stasher.js
kola-er: Done

packages/core/src/tools/create-file-stasher.js
kola-er: Done

kola-er:
Resolved the concerns and added some tests. Please, have another look. cc @eliangcs @standielpls

packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -247,7 +247,7 @@ describe('smoke tests - setup will take some time', function () {
         yarnInstall(corePackage.path, appDir);
 
         // should not throw an error
-        runCommand(context.cliBin, ['test', '--skip-validate'], {
+        runCommand(context.cliBin, ['test', '--debug', '--skip-validate'], {
```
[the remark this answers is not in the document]
kola-er: Will be removing it. Added it to find out what is causing the CLI smoke test to fail https://github.com/zapier/zapier-platform/actions/runs/8599623736/job/23562919199?pr=768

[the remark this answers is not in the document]
kola-er: No luck yet though

kola-er:
@eliangcs @standielpls  Any idea why the [CLI smoke test on `z.stashFile`](https://github.com/zapier/zapier-platform/actions/runs/8603216148/job/23575840492) is failing? Odd that it now fails when reverted to the [initial commit](https://github.com/zapier/zapier-platform/pull/768/commits/601108732b583e77ab6479dd5fe0c983bb0120c1) that passed before.

eliangcs (approved):
Smoke tests are passing now. Approving.

PR #772, opened 2024-04-16
fix(core): Handle null response_content before passing to findSensitiveValues

kola-er:

### DIFF
--- a/packages/core/test/logger.js
+++ b/packages/core/test/logger.js
@@ -335,6 +335,39 @@ describe('logger', () => {
     ]);
   });
 
+  it('should leave response content of null uncensored', async () => {
+    const event = {
+      method: 'authentication.sessionConfig.perform',
+    };
+    const logger = createlogger(event, options);
+
+    const { message, data } = prepareTestRequest({
+      resBody: JSON.stringify(null),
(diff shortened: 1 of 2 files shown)


packages/core/test/logger.js
```diff
@@ -335,6 +335,39 @@ describe('logger', () => {
...
+    const event = {
+      method: 'authentication.sessionConfig.perform',
+    };
+    const logger = createlogger(event, options);
+
+    const { message, data } = prepareTestRequest({
+      resBody: JSON.stringify(null),
+    });
+
+    logger(message, data);
```
eliangcs, line 348: I removed `|| {}`, trying to reproduce the same error:

```
TypeError: Cannot convert undefined or null to object
```

I found that we need to add an `await` here. Otherwise, the error would be silently swallowed.

```suggestion
    await logger(message, data);
```

eliangcs (commented):
The fix itself looks good! But to reproduce the exact same error, the test code needs a tweak.

packages/core/test/logger.js
kola-er: Thanks. I updated it @eliangcs

PR #785, opened 2024-05-14
feat(core): add runtime check for performBulk

eliangcs:
Adds a runtime check `perform-bulk-return-type` to ensure the result returned by the `performBulk` function follow the requirements:

1. The returned object must contain _all_ the IDs from the input `bundle.bulk`.
2. Every sub-object in the returned object has an `outputData` object _or_ an `error` string.

I found there was no integration test that ran all the [checks](https://github.com/zapier/zapier-platform/blob/01ced32396a4f96b10609b68539e11115b07af7f/packages/core/src/checks/index.js#L2) in the [checkOutput](https://github.com/zapier/zapier-platform/blob/01ced32396a4f96b10609b68539e11115b07af7f/packages/core/src/create-app.js#L37) middleware, so I added [one](https://github.com/zapier/zapier-platform/blob/01ced32396a4f96b10609b68539e11115b07af7f/packages/core/test/create-app.js#L317). Since I enabled the checks by adding `command: 'execute'` [here](https://github.com/zapier/zapier-platform/blob/01ced32396a4f96b10609b68539e11115b07af7f/packages/core/test/create-app.js#L22), some other test cases in [`create-app.js`](https://github.com/zapier/zapier-platform/blob/01ced32396a4f96b10609b68539e11115b07af7f/packages/core/test/create-app.js) started failing. I ended up refactoring many of the test cases.

### DIFF
--- a/packages/core/src/checks/perform-bulk-return-type.js
+++ b/packages/core/src/checks/perform-bulk-return-type.js
@@ -31,19 +31,24 @@ const performBulkEchoesIds = {
       let missingIdsStr = missingIds.slice(0, LIMIT).join(', ');
       const remainingCount = missingIds.length - LIMIT;
       if (remainingCount > 0) {
-        // No want to flood the user with too many IDs
+        // Don't want to flood the user with too many IDs
         missingIdsStr += `, and ${remainingCount} more`;
       }
       return [`Result object is missing these IDs as keys: ${missingIdsStr}`];
     }
 
     const errors = [];
-    for (const [id, item] of Object.entries(results)) {
+    for (const id of inputIds) {
+      const item = results[id];
+
       if (!_.isPlainObject(item)) {
         errors.push(`Result object member with ID '${id}' must be an object`);
-      } else if (!_.isPlainObject(item.outputData)) {
+      } else if (
+        !_.isPlainObject(item.outputData) &&
+        typeof item.error !== 'string'
+      ) {
         errors.push(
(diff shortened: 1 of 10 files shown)


packages/core/src/checks/perform-bulk-return-type.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    const inputIds = bundle.bulk
+      .map((b) => {
+        return b && b.meta ? b.meta.id : null;
+      })
+      .filter((id) => id);
+
+    const outputIds = Object.keys(results);
+    const missingIds = inputIds.filter((id) => !outputIds.includes(id));
```
eliangcs, line 27: @FokkeZB do you think if the output should contain _all_ the IDs in the input (`bundle.bulk`)? The current implementation assumes yes.

kola-er (approved):

eliangcs:
@kola-er just pushed [a4080da](https://github.com/zapier/zapier-platform/pull/785/commits/a4080daf61944deaccf05726a8b5fc04607a486e) so the result item must have either one of `outputData` _or_ `error`.

packages/core/src/checks/perform-bulk-return-type.js
eliangcs: Updated the requirements in the description. So now each item must have an `outputData` object _or_ an `error` string.

kola-er (approved):

PR #786, opened 2024-05-14
feat(core): Define types for performBulk

kola-er:


example-apps/typescript/src/test/creates.test.ts
```diff
@@ -16,4 +16,45 @@ describe('movie', () => {
...
+      idem2: {
+        outputData: { genre, id: '2', title: 'Title 2', year: 2021 },
+        error: null,
+      },
+      idem3: {
+        outputData: { genre, id: '3', title: 'Title 3', year: 2022 },
+        error: null,
+      },
+    });
+  });
```
kola-er, line 59: @eliangcs Is this the test you referred to in the ticket or a manual one?

kola-er: Also, the `Bundle` and the new `BulkBundle` are mixed up in the [failing tests](https://github.com/zapier/zapier-platform/actions/runs/9086671182/job/24972860812?pr=786) in this file. I'm confused about this.

eliangcs: @kola-er sorry for the confusion! A manual test should be enough: Manually create a TypeScript integration project, add a create with `performBulk`, and use the new `index.d.ts` to see it's working as expected.

eliangcs (commented):
I tested the new `index.d.ts` with this example:

```typescript
import {
  ZObject,
  BulkBundle,
  PerformBulkResult,
} from 'zapier-platform-core';

const performBulk = async (
  z: ZObject,
  bundle: BulkBundle<{ title: string; year: number }>
): Promise<PerformBulkResult> => {
  const movies = bundle.bulk.map((b) => {
    return {
      _idempotencyKey: b.meta.id,
      title: b.inputData.title,
      year: b.inputData.year,
    };
  });
    method: 'POST',
    url: 'https://example.com/api/movies/',
    body: {
      objects: movies,
    },
  });
  return movies.reduce(
    (result: PerformBulkResult, movie: { [x: string]: any }) => {
      result[movie._idempotencyKey] = {
        outputData: {
          title: movie.title,
          year: movie.year,
        },
      };
      return result;
    },
    {}
  );
};

export default {
  key: 'movie',
  noun: 'Movie',

  display: {
    label: 'Create Movie',
    description: 'Creates a new movie.',
  },

  operation: {
    performBulk,
    inputFields: [
      { key: 'title', required: true },
      { key: 'year', type: 'integer' },
    ],
    sample: {
      id: '1',
      title: 'example',
    },
  },
};
```

It works great! It was able to detect type issues. For example, when I replaced `outputData` with `outputData2`, my IDE would show:

> 'outputData2' does not exist in type 'PerformBulkResult`

which is good 👍 

Remove the new code you added in `example-apps/typescript` and then I think this is good to merge. Thanks!

packages/core/index.d.ts
```diff
@@ -204,3 +204,31 @@ export interface ZObject {
...
+  meta: {
+    id: string;
+    [x: string]: any;
+  };
+}
+
+export interface PerformBulkItem {
+  outputData: { [x: string]: any };
+  error?: string;
+}
```
eliangcs, line 219: To be consistent with #785, only _one_ of `outputData` or `error` is required. Maybe we could do this with two interfaces?

```typescript
export interface PerformBulkSuccessItem {
  outputData: { [x: string]: any };
}

export interface PreformBulkErrorItem {
  error: string;
}

export type PerformBulkItem = PerformBulkSuccessItem | PerformBulkErrorItem;
```

example-apps/typescript/src/test/creates.test.ts
kola-er: Thank you.

packages/core/index.d.ts
kola-er: I think these are the possible cases (`outputData` and `error`) or (`outputData`) or (`error`) for the and/or ;

```js
export interface PerformBulkSuccessItem {
  outputData: { [x: string]: any };
  error?: string;
}

export interface PreformBulkErrorItem {
  outputData?: { [x: string]: any };
  error: string;
}

export type PerformBulkItem = PerformBulkSuccessItem | PerformBulkErrorItem;
```

What do you think?

eliangcs: So `outputData` and `error` can coexist. That makes sense!

eliangcs (approved):

PR #787, opened 2024-05-15
docs(cli): Add docs for bulk-write support

kola-er:

### DIFF
--- a/packages/cli/README-source.md
+++ b/packages/cli/README-source.md
@@ -806,6 +806,24 @@ To define an Output Field for a nested field use `{{parent}}__{{key}}`. For chil
 [insert-file:./snippets/output-fields.js]
 ```
 
+## Bulk Create Actions
+
+*Added in v15.8.0.*
+
+Bulk Create allows you to create objects in bulk with a single or fewer API request(s). This is useful when you want to reduce the number of requests made to your server. When enabled, Zapier holds the data until the buffer reaches a size limit or a certain time has passed, then sends the buffered data using the `performBulk` function you define.
+
+To implement a Bulk Create, you define a [`bulk`](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#bulkobjectschema) configuration object and a `performBulk` function in the `operation` object. In the `bulk` config object, you specify how you want to group the buffered data using the `groupedBy` setting and the maximum number of items to buffer using the `limit` setting.
+
+The `performBulk` function should replace the `perform` function. Note that `perform` cannot be defined along with `performBulk`. Check out the [`create` action operation schema](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#basiccreateactionoperationschema) for details.
+
+Similar to the general `perform` function accepting two arguments, [`z`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#z-object) and [`bundle`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bundle-object) objects, the `performBulk` function accepts [`z`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#z-object) and [`bulkBundle`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bulkbundle-object) objects. They share the same `z` object, but the `bulkBundle` object is different from the `bundle` object. The `bulkBundle` object has an idempotency ID set at `bulkBundle.bulk[].meta.id` for each object in the bulk. `performBulk` would have to return the idempotency IDs to tell Zapier which objects were successfully written. Find the details about the `bulkBundle` object [here](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bulkbundle-object).
+
+Here is an example of a Bulk Create action:
+
+```js
+[insert-file:./snippets/bulk-create.js]
+```
+
 ## Z Object
 
 We provide several methods off of the `z` object, which is provided as the first argument to all function calls in your app.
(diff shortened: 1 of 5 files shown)


packages/cli/README-source.md
```diff
@@ -1081,6 +1081,48 @@ This is an object that contains the data you returned from the `performSubscribe
...
+
+Similar to the general `perform` function accepting 2 arguments [`z`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#z-object) and [`bundle`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bundle-object) objects, the `performBulk` function accepts [`z`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#z-object) and [`bulkbundle`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bulkbundle-object) objects. They share the same `z` object, but the `bulkbundle` object is different from the `bundle` object. The `bulkbundle` object has an idempotency ID set at `bulkbundle.bulk[].meta.id` for each invocation run in the bulk that would be used within the `performBulk` to tell Zapier backend which data items were successfully written. Find the details about the `bulkbundle` object [here](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bulkbundle-object).
+
+Here is an example of a bulk-write usage in a `create` action:
```
eliangcs, line 1120: Suggested revision:

```markdown
## Bulk Create Actions

*Added in v15.8.0.*

Bulk Create allows you to create objects in bulk with a single or fewer API request(s). This is useful when you get want to reduce the number of requests made to your server. When enabled, Zapier holds the data until the buffer reaches a size limit or a certain time has passed, then sends the buffered data using the `performBulk` function you define.

To implement a Bulk Create, you define a [`bulk`](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#bulkobjectschema) configuration object and a `performBulk` function in the `operation` object. In the `bulk` config object, you specify how you want to group the buffered data using the `groupedBy` setting and the maximum number of items to buffer using the `limit` setting.

The `performBulk` function should replace the `perform` function. Note that `perform` cannot be defined along with `performBulk`. Check out the [`create` action operation schema](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#basiccreateactionoperationschema) for details.

Similar to the general `perform` function accepting two arguments, [`z`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#z-object) and [`bundle`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bundle-object) objects, the `performBulk` function accepts [`z`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#z-object) and [`bulkBundle`](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bulkbundle-object) objects. They share the same `z` object, but the `bulkBundle` object is different from the `bundle` object. The `bulkBundle` object has an idempotency ID set at `bulkBundle.bulk[].meta.id` for each object in the bulk. `performBulk` would have to return the idempotency IDs to tell Zapier which objects were successfully written. Find the details about the `bulkBundle` object [here](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bulkbundle-object).

Here is an example of a Bulk Create action:
```

Why:

* In most public-facing docs, we say "Create instead of "Write", so I'd suggest we say "Bulk Create" instead of "Bulk-Write Create".
* In the beginning, I started with a high-level description on why you would want to use a Bulk Create (to reduce the number of requests made to the partner's server).
* There isn't a notion of an "invocation run" in the current docs. It may be confusing for partners, so I removed it.

Let me know what you think. Thanks!

packages/cli/README-source.md
```diff
@@ -1081,6 +1081,48 @@ This is an object that contains the data you returned from the `performSubscribe
...
+
+```js
+[insert-file:./snippets/bulk-write.js]
```
eliangcs, line 1123: Likewise, replacing `write` with `create`:

```suggestion
[insert-file:./snippets/bulk-create.js]
```

packages/cli/README-source.md
```diff
@@ -1081,6 +1081,48 @@ This is an object that contains the data you returned from the `performSubscribe
...
+
+#### `bulkbundle.bulk[].inputData`
+
+It is a user-provided data for a particular run of a `create` action in the bulk, as defined by the [`inputFields`](#input-fields).
+
+#### `bulkbundle.bulk[].meta`
+
+It contains an idempotency `id` provided to the `create` action to identify each run's data in the bulk data.
+
+## Bulk-Write in Create Actions
```
eliangcs, line 1112: I think this section is better placed before `## Output Fields` and after `### Nested & Children (Line Item) Fields`. It can still be a level-two heading, but I suggest we say `## Bulk Create Actions`.

packages/cli/snippets/bulk-write.js
```diff
@@ -0,0 +1,103 @@
+const performBulk = async (z, bulkBundle) => {
+  // Grab the line items, preserving the order
+  const rows = bulkBundle.bulk.forEach(({inputData}) => {
```
eliangcs, line 3: `.forEach()` returns `undefined`:

```
> a = [1, 2, 3]
[ 1, 2, 3 ]
> a.forEach((x) => x + 1)
undefined
```

Should we use `.map()` instead?

packages/cli/README-source.md
kola-er: That would make it be in between section `## Input Fields` and `## Output Fields`. Please, why so?

packages/cli/README-source.md
kola-er: Perfect, thanks.

kola-er: Updated

packages/cli/README-source.md
kola-er: Updated

packages/cli/snippets/bulk-write.js
kola-er: Updated

packages/cli/README-source.md
eliangcs: Sorry I had a typo there. It should be before `## Z Object` and after `### Nested & Children (Line Item)`. Something like:

kola-er: Done, thanks

eliangcs (approved):

PR #788, opened 2024-05-27
[PDE-5065] cli(fix): Add `shell: true` as a default option for Windows users when running spawn() command

### LINKED ISSUE #784: [Bug]: zappier push throws "spawn EINVAL" error
### Bug Description

When you execute `zapier push` on a Windows system with Nodejs version 18.20.2. It results in the below error
```
✔ Copying project to temp directory
✖ Installing project dependencies
 »   Error: spawn EINVAL
 »
 »   re-run this command with `--debug` for more info
 ```
The same error is thrown by `zapier test` or any other command that tries to execute npm as part of it.
As a workaround you can use an older version of node like 18.12.0

### Reproduction Steps

create an integration using zapier-cli on a windows device with Nodejs 18.20.2 (LTS) and execute `zapier test` or `zapier push`

### Zapier Platform version

15.7.2

### Node.js version

18.20.2

### Your Operating System

Windows 10 (win32-x64)

### npm/yarn version

10.5.0

### App ID

_No response_

### More Details

It is caused by [runCommand function](https://github.com/zapier/zapier-platform/blob/main/packages/cli/src/utils/misc.js#L27) due to changes in `child_process.spawn` which now requires shell option to be true for executing `.cmd` or `.bat` commands. This change was from [cve-2024-27980](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2#command-injection-via-args-parameter-of-child_processspawn-without-shell-option-enabled-on-windows-cve-2024-27980---high).

I would like to make a PR for this. Is it ok to just update the runCommand function to set shell as true?

### PR DESCRIPTION
**Pull request by rnegron**
This is related to `CVE-2024-27980`. See Node.js report for the full details: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2#command-injection-via-args-parameter-of-child_processspawn-without-shell-option-enabled-on-windows-cve-2024-27980---high

This fixes #784.

### DIFF
--- a/packages/cli/src/utils/misc.js
+++ b/packages/cli/src/utils/misc.js
@@ -25,7 +25,9 @@ const isWindows = () => {
 
 // Run a bash command with a promise.
 const runCommand = (command, args, options) => {
-  if (isWindows()) {
+  const isRunningUnderWindows = isWindows();
+
+  if (isRunningUnderWindows) {
     command += '.cmd';
   }
 
@@ -34,6 +36,11 @@ const runCommand = (command, args, options) => {
   }
 
   options = options || {};
+
+  if (isRunningUnderWindows) {
+    // See CVE-2024-27980
+    options.shell = true;
+  }
 
   debug('\n');
   debug(
--- a/packages/cli/src/utils/misc.js
+++ b/packages/cli/src/utils/misc.js
@@ -25,19 +25,15 @@ const isWindows = () => {
 
 // Run a bash command with a promise.
 const runCommand = (command, args, options) => {
-  const isRunningUnderWindows = isWindows();
-
-  if (isRunningUnderWindows) {
-    command += '.cmd';
-  }
-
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/cli/src/utils/misc.js [hunk 1]
```diff
@@ -25,7 +25,9 @@ const isWindows = () => {
 
 // Run a bash command with a promise.
 const runCommand = (command, args, options) => {
-  if (isWindows()) {
+  const isRunningUnderWindows = isWindows();
+
+  if (isRunningUnderWindows) {
```
standielpls, line 30: nit: Seems like we can combine the other `isRunningUnderWindows` conditional? By either moving `command += '.cmd'` down or moving `options` up?

rnegron: I like it! Thanks for the suggestion ✅

standielpls (approved):

PR #796, opened 2024-06-04
feat(schema) Add retry and filter to the root-level of throttle config

kola-er:
- Add "retry" and "filter" to the root-level of the throttle configuration
- Remove `throttle` for hook trigger.


packages/schema/lib/schemas/ThrottleObjectSchema.js
```diff
@@ -36,6 +36,16 @@ module.exports = makeSchema(
           type: 'string',
         },
       },
+      retry: {
+        description:
+          'The effect of throttling on the tasks of the action. `true` means throttled tasks are automatically retried after some delay, while `false` means tasks are held without retry. It defaults to `true`.',
+        type: 'boolean',
+      },
+      filter: {
+        description: `EXPERIMENTAL: Account-based attribute to override the throttle by. You can set to one of the following: "free", "trial", "paid". Therefore, the throttle scope would be automatically set to "account" and ONLY the accounts based on the specified filter will have their requests throttled based on the throttle overrides while the rest are throttled based on the original configuration.`,
```
eliangcs, line 45: 1. Can you share why `filter` is experimental while `retry` is not?
2. Regarding "the throttle scope would be automatically set to 'account'", what would happen if `scope` is not `account`?

kola-er: >Regarding "the throttle scope would be automatically set to 'account'", what would happen if scope is not account?

`filter` is an account-based config setting that uses ONLY `account` as the scope and disregards any explicit scope set.

>Can you share why filter is experimental while retry is not?

I assumed we wanted to open up the `retry` to the public due to the need it was requested for.

```
As we're planning on increasing the polling rate for all integrations, we're concerned about one or two apps which might become rate limited and thus, want to throttle them in advance to avoid issues...

Ultimately I think we need to avoid the possibility of stacking up retry attempts faster than they'll actually be processed, causing a pile that only gets bigger with time...
```

from https://zapier.slack.com/archives/C04P5J5R8E4/p1716513877072559

kola-er: @FokkeZB Please, did you intend to make the root-level `retry` experimental when you suggested it? https://zapier.slack.com/archives/D3YRTGHAM/p1717433047352299

FokkeZB: What's effect of marking it as experimental? Does that imply it's only for internal usage? As the override-level `filter` property isn't experimental either, I don't see why the root-level one should.

kola-er: @FokkeZB Yes, experimental implies internal usage.  The override-level `filter` is experimental indirectly because the `override` object itself is experimental.

FokkeZB: Ah, ok. I don't have a strong opinion on whether this one should be experimental.

eliangcs (dismissed):

kola-er:
@eliangcs Documented and added a functional constraint test on preventing devs from setting the `retry` field for polling triggers. Since there's no point in retrying polling tasks, the backend ignores it. Please, review the [latest commit](https://github.com/zapier/zapier-platform/pull/796/commits/98c7a2cf92a6d4439163b577b0fc776da3892091). Thanks.

packages/schema/lib/functional-constraints/pollingThrottle.js
```diff
@@ -0,0 +1,30 @@
...
+const _ = require('lodash');
+const jsonschema = require('jsonschema');
+
+const pollingThrottle = (definition) => {
+  const errors = [];
+  const actionType = 'triggers';
+
+  if (definition[actionType]) {
+    _.each(definition[actionType], (actionDef) => {
+      if (actionDef.operation && actionDef.operation.throttle && _.has(actionDef.operation.throttle, 'retry') && actionDef.operation.type === 'polling') {
```
eliangcs, line 12: `polling` is the default, so a polling trigger can have an undefined `operation.type`.

```suggestion
      if (
        actionDef.operation &&
        actionDef.operation.throttle &&
        _.has(actionDef.operation.throttle, 'retry') &&
        (!actionDef.operation.type || actionDef.operation.type === 'polling')
      ) {
```

eliangcs (commented):
Looks good except for the suggestion below. I'll approve once that's fixed.

Question: Why can't a polling trigger have `retry`? Does that mean it makes sense for a hook trigger to have `retry`?

packages/schema/lib/functional-constraints/pollingThrottle.js
eliangcs: Then should we also have a functional constraint that stops hook triggers from having `throttle`? @kola-er

FokkeZB: Yeah, probably block throttle entirely for any type of trigger.

It's not completely in-line with https://coda.io/d/Edge-Team_dCaLL2dK8Zu/No-More-Action-Types_suat9 as it should depend on _how_ an action is used, but pragmatically the correct thing to do for now.

kola-er: We can remove the `throttle` entirely from hook trigger, but we still need it for the polling trigger. As pointed out [here](https://zapier.slack.com/archives/C04P5J5R8E4/p1716513877072559), it still has a use-case.

>As we're planning on increasing the polling rate for all integrations, we're concerned about one or two apps which might become rate limited and thus, want to throttle them in advance to avoid issues.

kola-er: Done @eliangcs

packages/schema/lib/schemas/BasicHookOperationSchema.js
```diff
@@ -83,7 +83,6 @@ BasicHookOperationSchema.properties = {
   inputFields: BasicHookOperationSchema.properties.inputFields,
   outputFields: BasicHookOperationSchema.properties.outputFields,
   sample: BasicHookOperationSchema.properties.sample,
-  throttle: BasicHookOperationSchema.properties.throttle,
```
eliangcs, line 86: 👍

eliangcs (approved):

PR #803, opened 2024-06-11
[PDE-5063] feat(core): Add afterResponse middleware to check if HTTP redirect is made to disallowed domain

rnegron:
As mentioned in the [work ticket](https://zapierorg.atlassian.net/browse/PDE-5063), `node-fetch` doesn't support conditional erroring on redirects. In this MR, we add an "after" middleware to check if the `z.request` experienced a redirect, and error out if the redirection ended up in a disallowed domain (currently only disallowing localhost). We need to check the `resp.url` as the `req.url` does not change on redirect.

Changes:
* Install `nock` explicitly in `core` package
* Add a `throwForDisallowedHostnameAfterRedirect` middleware and use it by default in the `httpAfters` middleware list (after `logResponse`) 
* Update `prepare-response.js` in order to surface the `resp.redirected` and `resp.url` properties for the new middleware. An alternative here could be to move the new middleware even higher on the list.

An alternative approach here could be to set `redirect: 'manual'` as an [option](https://github.com/node-fetch/node-fetch?tab=readme-ov-file#manual-redirect) and then perform manual parsing of the headers [here](https://github.com/zapier/zapier-platform/blob/31bbfec850cfc26eb8e4ad2893b7b1e47022d2f0/packages/core/src/tools/request-client.js#L27).

The big upside to the alternative approach is that we could prevent the `disallowed` domain request from even happening, as opposed to just catching it like we do here. A potential downside is that it could potentially introduce breaking changes by not fully understanding how integrations are leveraging redirects (speaking for myself!) But ultimately, this change is just an attempt to help protect partners from end-user redirect abuse, which this this MR tackles.

### DIFF
--- a/packages/core/src/http-middlewares/after/throw-for-disallowed-hostname-after-redirect.js
+++ b/packages/core/src/http-middlewares/after/throw-for-disallowed-hostname-after-redirect.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { DisallowedRedirectError } = require('../../errors');
+const { Error } = require('../../errors');
 
 const disallowedRedirectHosts = [
   // Loopback addresses (IPv4)
@@ -24,8 +24,10 @@ function isDisallowedAfterRedirect(url) {
 }
 
 const throwForDisallowedHostnameAfterRedirect = (resp) => {
-  if (resp.redirected && isDisallowedAfterRedirect(resp.request.url)) {
-    throw new DisallowedRedirectError();
+  // Looking at the response URL instead of the request URL
+  // because the response URL can change after a redirect
+  if (resp.redirected && isDisallowedAfterRedirect(resp.url)) {
+    throw new Error('Redirecting to disallowed hostname');
   }
 
   return resp;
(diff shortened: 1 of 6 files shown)


packages/core/src/http-middlewares/after/throw-for-disallowed-hostname-after-redirect.js
```diff
@@ -0,0 +1,36 @@
...
+    // (being permissive just in case it affects backwards compatibility)
+    return false;
+  }
+}
+
+const throwForDisallowedHostnameAfterRedirect = (resp) => {
+  // Looking at the response URL instead of the request URL
+  // because the response URL can change after a redirect
+  if (resp.redirected && isDisallowedAfterRedirect(resp.url)) {
+    throw new Error('Redirecting to disallowed hostname');
```
rnegron, line 30: Should we throw a specific `Error` class here? I didn't notice a particularly appropriate one so we may have to create a new one.

eliangcs: I think a general `Error` class is good enough.

eliangcs (commented):
Looks good except for the order of the middlewares. Feel free to ask for someone else to approve once fixed.

Regarding the `redirect: manual` alternative, I thought about it but immediately gave up on that idea. There's a lot to cover in the [RFC](https://datatracker.ietf.org/doc/html/rfc7231#section-6.4). I don't think we'd want to re-implement the HTTP redirect that node-fetch already implements.

standielpls (approved):
Nice thanks!

PR #808, opened 2024-06-21
PDE-5125 - feat(core) Handle large response payloads

standielpls:
https://zapierorg.atlassian.net/browse/PDE-5125

This PR addresses a new event field: `autostashPayloadOutputLimit` that we can process.
If this value is set from the backend, we can conditionally stash the payload response. This is to work around the Lambda 6mb output limit. 
There are some changes as well to refactor some things around, as we will be reusing the `uploader()` functionality to stash the payload.

### DIFF
--- a/packages/core/src/tools/create-response-stasher.js
+++ b/packages/core/src/tools/create-response-stasher.js
@@ -31,8 +31,8 @@ const stashResponse = async (input, response, size) => {
       signedPostData,
       response.toString(), // accept JSON string to send to uploader.
       size,
-      crypto.randomUUID() + '.json',
-      'application/json'
+      crypto.randomUUID() + '.txt',
+      'text/plain'
     )
   );
 };
--- a/packages/core/src/tools/create-response-stasher.js
+++ b/packages/core/src/tools/create-response-stasher.js
@@ -2,7 +2,7 @@
 
 const _ = require('lodash');
 const uploader = require('./uploader');
-const { v4: uuidv4 } = require('uuid');
+const crypto = require('crypto');
 
 const withRetry = async (fn, retries = 3, delay = 100, attempt = 0) => {
   try {
@@ -31,7 +31,7 @@ const stashResponse = async (input, response, size) => {
       signedPostData,
       response.toString(), // accept JSON string to send to uploader.
       size,
-      uuidv4() + '.json',
+      crypto.randomUUID() + '.json',
       'application/json'
     )
   );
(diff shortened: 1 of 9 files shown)


eliangcs (approved):
I have a suggestion on the test code style. Otherwise, it looks good to me!

packages/core/src/tools/create-response-stasher.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+  const signedPostData = await rpc('get_presigned_upload_post_data');
+  return withRetry(
+    _.partial(
+      uploader,
+      signedPostData,
+      response.toString(), // accept JSON string to send to uploader.
+      size,
+      crypto.randomUUID() + '.txt',
+      'text/plain'
```
standielpls, line 35: In the previous commit I had these as 
```
crypto.randomUUID() + '.json',
'application/json'
```
But after a E2E test with z/z changes, I was unable to upload this data.
So I updated it to be text (like `create-file-stasher` supports) and it works. 
Perhaps something to do with xml nature of the S3 uploader

packages/core/test/app-middleware.js
```diff
@@ -65,4 +69,132 @@ describe('app middleware', () => {
...
+      );
+      input._zapier.rpc = rpc;
+
+      // set the payload autostash limit
+      // this limit is lower than res, so do not stash, let it fail
+      input._zapier.event.autostashPayloadOutputLimit = 8 * 1024 * 1024;
+
+      const output = await app(input);
+      output.should.not.have.property('resultsUrl');
+    });
```
kola-er, line 119: Seems the same with the next test case.

kola-er: Guess, this should be removed.

standielpls: Yes, good catch, this was extra

packages/core/src/tools/create-response-stasher.js
```diff
@@ -0,0 +1,40 @@
...
+    return withRetry(fn, retries, delay, attempt + 1);
+  }
+};
+
+// responseStasher uploads the data and returns the URL that points to that data.
+const stashResponse = async (input, response, size) => {
+  const rpc = _.get(input, '_zapier.rpc');
+
+  if (!rpc) {
+    throw new Error('rpc is not available');
```
pragmatic-zac, line 25: Asking for my own curiosity 🧠 here - when/how would rpc be unavailable?

standielpls: Usually it's fine, just sanity check. `create-rpc-client` looks for a few things, esp if the Lambda event doesn't have all the values it needs to create the client

pragmatic-zac (approved):

PR #809, opened 2024-06-24
feat(core): Allow relative paths via z.require() [PDE-5116]

kreddlear:
Based on the POC patch logic here: https://gitlab.com/zapier/zapier/-/merge_requests/57310/diffs#note_1923825583

`z.require('./constants')` should work relative to the app root rather than to `node_modules/zapier-platform-core/src/app-middlewares/before`.

### DIFF
--- a/packages/core/test/userapp/index.js
+++ b/packages/core/test/userapp/index.js
@@ -36,9 +36,7 @@ const ListRequire = {
     },
     operation: {
       perform: (z, bundle) => {
-        // need to specify the full /test/userapp path because
-        // it's not possible to mock process.cwd for require.resolve
-        // but in prod, that will return the app root directory
+        // in prod, process.cwd will return the app root directory
         const { BASE_URL } = z.require('./test/userapp/constants.js');
         return BASE_URL;
       },
--- a/packages/core/test/userapp/index.js
+++ b/packages/core/test/userapp/index.js
@@ -26,6 +26,26 @@ const List = {
   },
 };
 
+const ListRequire = {
+  key: 'listrequire',
+  noun: 'List',
+  list: {
+    display: {
+      description: 'Trigger on new thing in list.',
+      label: 'List',
+    },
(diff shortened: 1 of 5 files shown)


eliangcs:
@kreddlear the changes themselves look good. Can you add some tests to make sure it works as expected? Thanks!

You might want to create a separate PR for the `loadApp` change because it will need tests too and it'd be easier to review separately.

kreddlear:
@eliangcs Definitely, I started adding tests yesterday but didn't quite finish. I'll split this up to make it easier to review as well, and ping you and Zac directly when it's ready for review.

packages/core/test/userapp/index.js
```diff
@@ -26,6 +26,24 @@ const List = {
...
+  noun: 'List',
+  list: {
+    display: {
+      description: 'Trigger on new thing in list.',
+      label: 'List',
+    },
+    operation: {
+      perform: (z, bundle) => {
+        // in prod, process.cwd will return the app root directory
+        const { BASE_URL } = z.require('./test/userapp/constants.js');
```
kreddlear, line 40: @eliangcs @pragmatic-zac  👋 I'd appreciate your thoughts here - would this test be sufficient? I'd prefer to have it referencing `./constants.js` directly since that's how it would be in an actual app/app extension, but I tried various ways of mocking `process.cwd` (mostly using `mock-fs`) and ran into issues getting it to recognize the module 🤔 If this test should be made more robust, would `mock-fs` also be y'all's go-to, or is there another way I should explore of mocking that when `z.require()` is used?

eliangcs: Yes, I think this is sufficient. Thanks for adding the test! Approving.

I was wondering if there's a better way to locate the app root directory other than `process.cwd`, but I couldn't find one, so 🤷 .

eliangcs (approved):

PR #810, opened 2024-06-25
feat(core): [PDE-5137] Add allowlist for specific content types to log HTTP response/response for in patched HTTP client

rnegron:
When using `z.request()`, streaming data is obscured in logs via the `<streaming data>` (or `<probably streaming data>` when using the `raw` option). But in cases where `node-fetch` or another request library is used directly, we may still attempt to send binary data over to our `http-to-gelf` service. This is (at partly) responsible for an increase in costs associated with inbound data.

The idea behind this PR is: why even log binary data if it's unreadable in logs? 😅 
Here, I've adopted an allowlist approach based on `content-type` headers. But I'm open to feedback and ideas on alternatives!

See Slack for full context: https://zapier.slack.com/archives/C5Z9BP4U9/p1719341518902659?thread_ts=1716213680.917849&cid=C5Z9BP4U9


packages/core/test/tools/create-http-patch.js
```diff
@@ -0,0 +1,172 @@
+'use strict';
```
rnegron, line 1: This file is actually just `packages/core/test/tools/http-patch.js` renamed to be equivalent to the name of the file it's testing. I also removed a top-level `describe` and added two new tests to cover the changes in this PR.

rnegron:
@eliangcs Curious about your thoughts on this approach! 😄

packages/core/src/tools/create-http-patch.js
```diff
@@ -54,19 +57,34 @@ const createHttpPatch = (event) => {
...
+        const shouldIncludeResponseData =
+          ALLOWED_HTTP_DATA_CONTENT_TYPES.has(responseContentType);
+
         const sendToLogger = (responseBody) => {
           // Prepare data for GL
           const logData = {
             log_type: 'http',
-            request_type: 'devplatform-outbound',
+            // Using a custom request_type to differentiate from z.request() since `request_via_client` is not logged in GL
+            request_type: 'patched-devplatform-outbound',
```
eliangcs, line 75: > Using a custom request_type to differentiate from z.request()

The requests sent by `z.request()` is also logged using `sendToLogger`. So changing `request_type` to `patched-devplatform-outbound` here would also change it for `z.request()`. Let me know if I missed anything.

rnegron: 🤔 Wait, I think that's incorrect. I was thinking that request sent from `z.request()` used the `log-response` http after middleware (setting `request_type` [here](https://github.com/zapier/zapier-platform/blob/7d985a99f293712c152de1267ddf1efe03fe9f11/packages/core/src/http-middlewares/after/log-response.js#L30)) and that gets sent over to `http-to-gelf` via the `LogStream` in `create-logger` [here](https://github.com/zapier/zapier-platform/blob/eb1c70c655a731f0536127587cb83c0252ca1509/packages/core/src/tools/create-logger.js#L202).

This seems to be backed up by [this line](https://github.com/zapier/zapier-platform/blob/7d985a99f293712c152de1267ddf1efe03fe9f11/packages/core/src/tools/create-http-patch.js#L52) in `create-http-patch` which tries to discard that outgoing request (because it would double-log otherwise)

rnegron: Just ran a test with the code in this PR to verify the above, using the following simple `create`
```javascript
  const filePromise = z.request({
    url: bundle.inputData.fileToDownloadUrl,
    raw: true,
  });

  return { stashedUrl: z.stashFile(filePromise) };
};
```
It generated three HTTP logs:
* The initial RPC call to obtain a pre-signed S3 URL: [log](https://logs.internal.zapier.com/messages/integration_4602/40cacee5-34b6-11ef-984e-02819440192e)
* The `GET` request to download the file: [log](https://logs.internal.zapier.com/messages/integration_4602/40dc8221-34b6-11ef-ab9e-26c2d2856c29)
* The `POST` request to stash the file: [log](https://logs.internal.zapier.com/messages/integration_4602/4457c633-34b6-11ef-984e-02819440192e)

From the above, we can see that the middle `z.request` HTTP call keeps the `request_type` as `devplatform-outbound`, while the other two (RPC call from [here](https://github.com/zapier/zapier-platform/blob/main/packages/core/src/tools/create-file-stasher.js#L232),  S3 POST from [here](https://github.com/zapier/zapier-platform/blob/main/packages/core/src/tools/uploader.js#L46)) use `request_type`: `patched-devplatform-outbound`, which is good!

We can also see `<unsupported format>` in the request/response data for the first and last request. Actually, while I was there, I modified the above action to use `node-fetch` directly, because what we want to fix with this PR is the case where that middle request uses something other than `z.request()` which today sends a lot of binary data to Graylog (potentially crashing the Lambda in the worst case). Here is the [log](https://logs.internal.zapier.com/messages/integration_4602/d2e93b01-34b5-11ef-8be3-16c430b71d4e), showing that the `response_content` is removed. I actually could not recreate a comparison log for this situation (using `node-fetch` but without this PR's logic changes) because this "middle log" that shows the `GET` request happening when using `node-fetch` does not even show up in `http-to-gelf`... I highly suspect it hits the `LOGGER_TIMEOUT` and gets rejected or something? I can see the first and last log but not the middle one 😮 

In fact, I had to use a small test file (20mb instead of 50mb) for this last test because trying to use the same one as in the logs above simply crashes the Lambda function 😬 ([log](https://logs.internal.zapier.com/messages/integration_4602/e87abf41-34b8-11ef-9929-d2df6421715e))

eliangcs (approved):

PR #811, opened 2024-06-25
feat(core): Update loadApp to default to CLI definition and merge in external action definition as needed [PDE-5114]

kreddlear:
Allow `loadApp` to default to a CLI integration's definition when no definition override (for Platform UI apps) is passed, so that the partial definition of the external action gets merged in.


pragmatic-zac (approved):
LGTM, I appreciate the test too 💯

eliangcs (approved):
Looks good! 👍

PR #813, opened 2024-06-26
feat(core) Update extendAppRaw to override arrays as well as objects [PDE-5114]

kreddlear:
The `extendAppRaw` function in `createLambdaHandler` currently assumes it should always/only concatenate arrays in an app definition. However, for many cases we actually do want to override the array, not concatenate it - for example for inputFields coming in from an external action, we should fully overwrite the base inputFields with the partial inputFields.

### DIFF
--- a/packages/core/integration-test/integration-test.js
+++ b/packages/core/integration-test/integration-test.js
@@ -198,7 +198,13 @@ const doTest = (runner) => {
             noun: 'Foo',
             operation: {
               perform: { source: 'return [{id: 12345}]' },
-              inputFields: [{ key: 'name', type: 'string' }],
+              inputFields: [
+                { key: 'name', type: 'string' },
+                { key: 'testing', type: 'string' },
+              ],
+              sample: {
+                id: 123,
+              },
             },
           },
         },
@@ -209,11 +215,12 @@ const doTest = (runner) => {
       const definitionExtension = {
         creates: {
           foo: {
-            noun: 'Foobar',
+            noun: 'Foo',
             operation: {
+              perform: { source: 'return [{id: 12345}]' },
               inputFields: [{ key: 'message', type: 'string' }],
               sample: {
-                id: 678,
+                name: 'sample',
               },
             },
           },
@@ -273,6 +280,59 @@ const doTest = (runner) => {
       const event = {
         command: 'execute',
         method: 'creates.foo.operation.inputFields',
+        appRawOverride: [definitionHash, definitionExtension],
+        rpc_base: 'https://mock.zapier.com/platform/rpc/cli',
+        token: 'fake',
+      };
+
+      return runner(event).then((response) => {
+        response.results.should.eql([{ key: 'message', type: 'string' }]);
+      });
+    });
+
+    it('should handle array of [appRawOverrideHash, appRawExtension] and overrides with an operation trigger key', () => {
+      const definition = {
+        creates: {
+          operation: {
+            key: 'operation',
+            operation: {
+              perform: { source: 'return [{id: 12345}]' },
+              inputFields: [
+                { key: 'name', type: 'string' },
+                { key: 'testing', type: 'string' },
+              ],
+              sample: {
+                id: 123,
+              },
+            },
+          },
+        },
+      };
+
+      mocky.mockRpcCall(definition);
+
+      const definitionExtension = {
+        creates: {
+          operation: {
+            operation: {
+              perform: { source: 'return [{id: 123}]' },
+              inputFields: [{ key: 'message', type: 'string' }],
+              sample: {
+                name: 'sample',
+              },
+            },
+          },
+        },
+      };
+
+      const definitionHash = crypto
+        .createHash('md5')
+        .update(JSON.stringify(definition))
+        .digest('hex');
+
+      const event = {
+        command: 'execute',
+        method: 'creates.operation.operation.inputFields',
         appRawOverride: [definitionHash, definitionExtension],
         rpc_base: 'https://mock.zapier.com/platform/rpc/cli',
         token: 'fake',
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/create-lambda-handler.js
```diff
@@ -30,17 +30,15 @@ const isRequestOrFunction = (obj) => {
 
 const extendAppRaw = (base, extension) => {
   const keysToOverride = [
```
kreddlear, line 32: After looking over [the schema](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md), I added `inputFields` as well as some other fields here since those seemed most likely for users to want to override with external actions. Please let me know if there are others that seem like a good fit.

pragmatic-zac (dismissed):

packages/core/src/tools/create-lambda-handler.js
```diff
@@ -30,17 +30,17 @@ const isRequestOrFunction = (obj) => {
...
-    'test',
+    'inputFields',
+    'outputFields',
     'perform',
     'performList',
     'performSubscribe',
     'performUnsubscribe',
+    'sample',
+    'test',
+    'throttle',
```
eliangcs, line 41: `inputFields`, `outputFields`, `sample`, and `throttle` don't seem to do anything in this array. In the later `if`, we have this condition:

```
if (
  keysToOverride.indexOf(key) !== -1 &&
  _.isPlainObject(srcValue) &&
  _.isPlainObject(objValue) &&
  isRequestOrFunction(srcValue) &&
  isRequestOrFunction(objValue)
)
```

- `inputFields` and `outputFields` would pass `keysToOverride.indexOf(key) !== -1`, but they won't pass `_.isPlainObject()`.
- `sample` and `throttle` won't pass `isRequestOrFunction()`.

I verified it with:

```js
const base = {
  triggers: {
    recipe: {
      operation: {
        sample: {
          id: 1,
        },
      },
    },
  },
};

const extension = {
  triggers: {
    recipe: {
      operation: {
        sample: {
          name: 'Pancakes',
        },
      },
    },
  },
};

const result = extendAppRaw(base, extension);
console.log(JSON.stringify(result, null, 2));
```

The output is:

```json
{
  "triggers": {
    "recipe": {
      "operation": {
        "sample": {
          "id": 1,
          "name": "Pancakes"
        }
      }
    }
  }
}
```

which is not what we want -- we want to fully overwrite the `operation.sample` object.

Also consider the case where a trigger/create/search key is `sample` too:

```js
const base = {
  triggers: {
    sample: {
      operation: {
        sample: {
          id: 1,
        },
      },
    },
  },
};

const extension = {
  triggers: {
    sample: {
      display: {
        label: 'Sample Trigger',
      },
      operation: {
        sample: {
          name: "sample",
        },
      },
    },
  },
};
```

The expected output is:

```json
{
  "triggers": {
    "sample": {
      "display": {
        "label": "Sample Trigger"
      },
      "operation": {
        "sample": {
          "name": "sample"
        }
      }
    }
  }
}
```

I'm not sure if `_.mergeWith()` can do that. If not, we might have to write our own deep merge function.

kreddlear: Thanks for pointing that out @eliangcs ! I had made a note that `inputFields` wouldn't get past the first `if` statement and forgot to come back to it 🤦‍♀️

kreddlear:
@eliangcs I've added some logic to the existing `mergeWith` customizer function - would you mind taking a look to see if that looks like it would cover the desired use case here better? Since it looks like all the overrides we care about are stored within the `operation` within a given trigger, I focused on handling that rather than rewriting a new deep merge function. If it seems ultimately better to just rewrite it though, or if there are holes in this solution as well, please let me know and I'll rework it again. Thank you!

packages/core/integration-test/integration-test.js
```diff
@@ -234,10 +241,105 @@ const doTest = (runner) => {
...
+        rpc_base: 'https://mock.zapier.com/platform/rpc/cli',
+        token: 'fake',
+      };
+
+      return runner(event).then((response) => {
+        response.results.should.eql([{ key: 'message', type: 'string' }]);
+      });
+    });
+
+    it('should handle array of [appRawOverrideHash, appRawExtension] and overrides with an operation trigger key', () => {
```
eliangcs, line 293: The test code seems to be a `create` key instead of a `trigger` key.

```suggestion
    it('should handle array of [appRawOverrideHash, appRawExtension] and overrides with an operation key', () => {
```

eliangcs (commented):
`inputFields`, `outputFields`, `sample`, and `throttle` will never make the first `if` (line 50) condition true; they are dead code. This is going to be added as a JIT patch, so I'd like to see the code as clean and minimal as possible.

I came up with the following code, totally tailored to our needs. The only dependency is Lodash's `_.isPlainObject()`. No schema validation and no `_.mergeWith()`. It passes your tests too.

```js
const isDefinedPrimitive = (v) => {
  return (
    v === null ||
    typeof v === 'string' ||
    typeof v === 'number' ||
    typeof v === 'boolean'
  );
};

const shouldFullyReplace = (path) => path[path.length - 2] === 'operation';

const extendAppRaw = (base, extension, path) => {
  if (extension === undefined) {
    return base;
  } else if (isDefinedPrimitive(extension)) {
    return extension;
  } else if (Array.isArray(extension)) {
    return [...extension];
  }

  path = path || [];

  if (_.isPlainObject(extension)) {
    if (shouldFullyReplace(path)) {
      return extension;
    } else {
      const baseObject = _.isPlainObject(base) ? base : {};
      const result = { ...baseObject };
      for (const [key, value] of Object.entries(extension)) {
        const newPath = [...path, key];
        result[key] = extendAppRaw(baseObject[key], value, newPath);
      }
      return result;
    }
  }
  throw new TypeError('Unexpected type');
};
```

Let me know what you think.

eliangcs (approved):
Thanks!

PR #814, opened 2024-06-27
PDE-5158 fix(core) - Fix afterApp large response cacher bug

standielpls:
Checks if output.results is truthy before attempting to cache

### DIFF
--- a/packages/core/src/app-middlewares/after/large-response-cacher.js
+++ b/packages/core/src/app-middlewares/after/large-response-cacher.js
@@ -6,6 +6,9 @@ const responseStasher = require('../../tools/create-response-stasher');
 
 const largeResponseCachePointer = async (output) => {
   const response = cleaner.maskOutput(output);
+  if (!response.results) {
+    return output;
+  }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/app-middlewares/after/large-response-cacher.js
```diff
@@ -6,6 +6,9 @@ const responseStasher = require('../../tools/create-response-stasher');
 
 const largeResponseCachePointer = async (output) => {
   const response = cleaner.maskOutput(output);
+  if (!response.results) {
+    return output;
+  }
```
rnegron, line 11: Semantically, any concern with `response.results` being JavaScript "falsy"? Like an empty string.

rnegron (approved):
LGTM, left a comment about edge-cases just in case!

packages/core/src/app-middlewares/after/large-response-cacher.js
standielpls: Yeah! The only case we want to move forward is if there's a big payload, so if it's falsy, we'll just carry on and have the next middleware deal with the `output`

PR #816, opened 2024-06-28
PDE-5159 feat(core) - Increase hydration payload limit

standielpls:
Increases the hydration payload limit. In older Node versions, this was capped at 8kb due to the hydration URL that we need to call eventually. Since Node 16, this is now 16kb.

### DIFF
--- a/packages/core/src/tools/wrap-hydrate.js
+++ b/packages/core/src/tools/wrap-hydrate.js
@@ -5,7 +5,8 @@ const crypto = require('crypto');
 const { DehydrateError } = require('../errors');
 
 // https://nodejs.org/docs/latest-v16.x/api/http.html#httpmaxheadersize
-// base64 encoding is approx 4/3, so our limit should be:
+// Base64 encoding adds approx 4/3 to the original size
+// To account for encoding, we use the inverse to calc the max original size (3/4)
 // 16kb limit * 1024 * (3 / 4) = 12.228 kb max, minus some room for additional overhead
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/wrap-hydrate.js
```diff
@@ -4,9 +4,10 @@ const crypto = require('crypto');
 
 const { DehydrateError } = require('../errors');
 
-// Max URL length for Node is 8KB (https://stackoverflow.com/a/56954244/)
-// 8 * 1024 * (3 / 4) = 6144 max, minus some room for additional overhead
-const MAX_PAYLOAD_SIZE = 6000;
+// https://nodejs.org/docs/latest-v16.x/api/http.html#httpmaxheadersize
+// base64 encoding is approx 4/3, so our limit should be:
```
rnegron, line 8: I appreciate adding this context for where the 3/4 comes from!
```suggestion
// base64 encoding is approx 3/4, so our limit should be:
```

rnegron (dismissed):

eliangcs (dismissed):

packages/core/src/tools/wrap-hydrate.js
standielpls: I think the `4/3` is correct, if a str is 100 chars, it increases the length by 4/3 (+ 33%) = 133 chars
To find the limit of the original value
if the limit is 100, we have to take the inverse to 3/4 = 75 chars 
^ but I can make that more clear!

rnegron (approved):

PR #818, opened 2024-07-03
PDE-4968 feat(schema-to-ts) - Introduce the Schema-to-TS compiler

### LINKED ISSUE #8: Typescript bindings

tkcranny:
> [!WARNING]  
> This is rather extensive and being contributed without much knowledge of the DevPlatform's architecture or process. Any feedback or suggestions, no matter how sweeping will be kindly appreciated.

This introduces the new `schema-to-ts` project in the `packages/` directory that compiles typescript definitions from `packages/schemas` and outputs `packages/core/types/zapier.generated.d.ts`. This solution has been explored in the past but has abandoned until now, owing to the difficulties expressing function and promise syntax between JSON Schema and TypeScript. See #8 and #260 for further details.

This is a cleaned up and expanded version of a Hackweek Project from Team Integrations in Q2 2024 ([coda](https://coda.io/d/Q2-2024-Unstoppable-Unbreakable-Hackweek_dlf_DVBaAxJ/Integration-Platform-Developer-Tooling_suFd2)).

## How it works

- It is a TypeScript module itself, built with `yarn build`.
- It's run via `yarn generate-types`, which will run `dist/cli.js` (once built). This CLI has a number of flags and options if needed.
- It reads `../schema/exported-schema.json` for its input (by default).
- It outputs `../core/types/zapier.generated.d.ts` (by default).
- The generated code is referenced in a unified export with the existing types (previously from `index.d.ts`, now `zapier.custom.d.ts`), and are all intended to be bundled in the ZPC distribution on [npm](https://www.npmjs.com/package/zapier-platform-core) for downstream developers.
- It has `precommit` and `git-add` scripts in its `package.json` to build the generated types on commits via husky & lerna, similar to how `cli` and `schema` build their docs and output artefacts.
- Internally, we use the internals from the `json-schema-to-typescript` ([npm](https://www.npmjs.com/package/json-schema-to-typescript)) library. We choose to render our own TypeScript output with this intermediate result due to some unpleasant limitations and to correctly patch in real function definitions and links to public documentation.
  - See [this issue](https://github.com/bcherny/json-schema-to-typescript/issues/334) opened by a former Zapien's ticket on the same repo for the same reason.

## Potential Issues

- As this is an in situ tool to generate the types per-commit, I don't foresee it needing to built and pushed to NPM like the other four packages. Therefore its place in `packages/` or even this repo itself may not be the correct location.
- It is an esm package and uses TypeScript and [Vitest](https://vitest.dev/), so it's different yet again from the existing CJS packages.
- I know one dependency states it requires node 18 ([marked](https://github.com/markedjs/marked/blob/600df67e39c9d4f8e1dd2a0d1f1568c6e71ef3ce/package.json#L112), for reflowing the schema markdown docs), and some other parts of the DevPlatform state a max of v16, so node versions may yet prove problematic.
- It's non-standard and non-trivial to go from our JSON Schemas to TypeScript. While it has enormous benefit to the engineers, contractors, and partner developers that make integrations, I'm cognizant of it being disruptive and the potential future maintenance burden to drop in like this.

## See Also
- [The Team Integrations Hackweek Coda on Typing](https://coda.io/d/Q2-2024-Unstoppable-Unbreakable-Hackweek_dlf_DVBaAxJ/Integration-Platform-Developer-Tooling_suFd2)
- Closes #8 from 2016
- The (not-merged, closed) PR that attempted it first: #260
  - Comment on why `#260` was closed and `#8` was reopened in 2021: https://github.com/zapier/zapier-platform/issues/8#issuecomment-827972091
- David Brownman's issue thread from 2020 on the limitations of json-schema-to-typescript: bcherny/json-schema-to-typescript/issues/334


packages/core/types/zapier.generated.d.ts
```diff
@@ -0,0 +1,1746 @@
...
+ * This file was automatically generated by Zapier's schema-to-ts tool.
+ * DO NOT MODIFY IT BY HAND. Instead, modify the source JSON Schema
+ * files, and/or the schema-to-ts tool and run its CLI to regenerate
+ * these typings.
+ *
+ * zapier-platform-schema version: 15.10.0
+ *  schema-to-ts compiler version: 0.1.0
+ */
+
+import type { Bundle, ZObject } from './zapier.custom';
```
rnegron, line 11: It looks like these are only necessary to support the `PerformFunction`. I wonder, would it make sense to add that `PerformFunction` to the `custom` typings and prevent this import? We'd probably have to eject the entire `Function` object which references `PerformFunction`, but it seems possible from the types dependencies.

rnegron: I see that these are added manually in `transformers.ts`. Are they required to be in this generated declarations file?

packages/schema-to-ts/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "json-schema-to-typescript": "^13.1.2",
+    "marked": "^12.0.2",
+    "pino": "^9.1.0",
+    "pino-pretty": "^11.0.0",
+    "prettier": "^3.2.5",
+    "word-wrap": "^1.2.5"
+  },
+  "devDependencies": {
+    "typescript": "^5.5.2",
+    "vitest": "^1.5.0"
```
rnegron, line 31: I had never heard of `vitest` , looks interesting! I see that is has a Jest-compatible API, so I'm curious if there was any specific reason to pick vitest over Jest? We already use Jest in other packages it may be nice to stick with Jest for standardization's sake.

packages/schema-to-ts/tsconfig.json
```diff
@@ -0,0 +1,39 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
+    /* Projects */
+
+    /* Language and Environment */
+    "target": "ESNext" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
rnegron, line 8: Since we are not generating any JavaScript from this tool, is this key something that could be omitted?  Or is it a required key?

rnegron (commented):
@tkcranny So exciting to see that huge `zapier.generated.d.ts`! Great job with this effort, it's really impressive. I particularly appreciate the thorough MR description and code comments across all the new files 💯 

I have not dived into the details yet but I left some review comments; happy to keep the conversation going here on on Slack! My thinking is that this unlocks a lot of value for developers and has been a long time coming, so I'd like to support rolling this out in any way I can.

Responding to some of the points in the description:

> As this is an in situ tool to generate the types per-commit, I don't foresee it needing to built and pushed to NPM like the other four packages. Therefore its place in packages/ or even this repo itself may not be the correct location.

Agreed, I think `packages/` makes sense as the location for NPM-published packages. I'd suggest a new, top-level `tools` folder, or even making `schema-to-ts` a top-level folder. That said, I _do_ think this code should live in this monorepo.

> It is an esm package and uses TypeScript and [Vitest](https://vitest.dev/), so it's different yet again from the existing CJS packages.

Since this code is not something that we publish to NPM and stakeholders install on their machines, I don't feel the need to worry about esm vs. cjs in this case. As for `Vitest`, how difficult would it be to replace with Jest? Are there any benefits to keeping `vitest` (it would be the only package to use it).

> I know one dependency states it requires node 18 ([marked](https://github.com/markedjs/marked/blob/600df67e39c9d4f8e1dd2a0d1f1568c6e71ef3ce/package.json#L112), for reflowing the schema markdown docs), and some other parts of the DevPlatform state a max of v16, so node versions may yet prove problematic.

Definitely a fair concern, thanks for bringing this up. There are two places where we explicitly specify which Node versions we support:
1. The `engine` key in `package.json` ([example](https://github.com/zapier/zapier-platform/blob/440d6768327d954cd805223477188448b7ef8d43/packages/cli/package.json#L22)) <- This is for end-users installing the package from NPM and should be kept in sync with actual Node support.
2. The `.nvmrc` file ([example](https://github.com/zapier/zapier-platform/blob/pde-4967-compiled-types/packages/cli/.nvmrc)) <- This is for contributors to the packages and is more flexible.

It's nice to keep these in-sync but we are due for a `node >= 18` engine update soon anyway, so I wouldn't worry too much about adding an `.nvmrc` specifying Node>=18 to the `schema-to-ts` folder. 😄 

> It's non-standard and non-trivial to go from our JSON Schemas to TypeScript. While it has enormous benefit to the engineers, contractors, and partner developers that make integrations, I'm cognizant of it being disruptive and the potential future maintenance burden to drop in like this.

Emphasis on the "enormous benefit" part! I personally feel it's a worthy trade-off. One thing I am concerned about is maintenance from the point of view of the `json-schema-to-typescript` package; should that version be more strictly pinned (and a comment added to reference the pin)? since this implementation borrows from `json-schema-to-typescript` internals, we'd have to be careful about any non-trivial changes made in that package's internals.

Some other things I wanted to mention:
* a short `README.md` file similar to the `core` package one [here](https://github.com/zapier/zapier-platform/blob/main/packages/core/README.md) (which now that I look at it may need some updating 😅) could probably be added for parity with other packages. And then referenced from the internal docs section [here](https://github.com/zapier/zapier-platform?tab=readme-ov-file#docs).

* In terms of the developer experience of generating the typing declarations file, I saw the `precommit` script in `package.json` but it doesn't seem to `yarn` (install the package requirements) beforehand. I suppose for most contributions to this repo, changes would not need to be made to this package and therefore it should be almost transparent that the typing is auto-generated before the commit happens. What about adding a lifecycle step to the `core` package to ensure that `schema-to-ts` is ready to run on changes to that package? There are a [lot of available lifecycle options](https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-operation-order) for npm scripts 😵. 
  * Note that we do have a set of steps we strictly follow ([Coda doc](https://coda.io/d/Team-Developer-Platform_di0MgBhlCWf/Releasing-a-New-zapier-platform-Version_su5eD#_luL_l)) when releasing a new zapier-platform version, and if it simplifies the process, as a temporary workaround, I could see us adding a `generating updated types` step.

* End-to-end testing! I mean, making sure that the typings work as expected for end-users. Have you thought about this aspect? I believe this was covered in the original hackathon proposal; I just want to make sure we have a conversation around potential impacts for end-users here. Related, and potentially as a way to reduce the potential impact of this release, what do you think about initially releasing the generated typing as "opt-in" while we gather feedback from internal builders on the validity/stability of the approach? I'm not sure yet how that would work but I'll give it some more thought if you think it makes sense 🤔 

* Seems there are failing tests in CI related to this package; we'll likely have to perform some CI logic separation here to prevent leakage between this and other packages' tests. As you mentioned, some of the failures are related to `node v16`  which this package does not support.

tkcranny (commented):
> Agreed, I think `packages/` makes sense as the location for NPM-published packages. I'd suggest a new, top-level `tools` folder, or even making `schema-to-ts` a top-level folder. That said, I _do_ think this code should live in this monorepo.

Gotcha! I've moved `schema-to-ts` it to the root now, as it'd be real lonely in `./tools/` without anything else to join it 😆 Happy to move if advised.

> As for `Vitest`, how difficult would it be to replace with Jest? Are there any benefits to keeping `vitest` (it would be the only package to use it).

I elaborated on this on the [comment about Vitest](https://github.com/zapier/zapier-platform/pull/818#discussion_r1671249648); generally Vitest makes working with TypeScript tests a lot nicer because we don't need to bring in in babel and/or ts-jest configurations. It's a fast and compatible drop in replacement.

> One thing I am concerned about is maintenance from the point of view of the `json-schema-to-typescript` package; should that version be more strictly pinned (and a comment added to reference the pin)? since this implementation borrows from `json-schema-to-typescript` internals, we'd have to be careful about any non-trivial changes made in that package's internals.

Excellent pickup. They've even had a major version bump since hackweek, which needed some of those very same internal imports tweaked. I've bumped and pinned it specifically now.

> a short `README.md` file similar to the `core` package

Of course, added one now 👍

---

As for Lifecycle Hooks, **end-to-end** tests, and Fixing CI, I will have a think, chat to Chris, and make some changes imminently. Thanks so much for what you've reviewed already, it's already in better shape!

tkcranny:
I've made some considerable changes today. Notably they include:

- End-to-end tests of the generated types with a pretty standard integration structured app.
  - The main test file is in `core/types/index.test-d.ts`.
  - This is then tested by the tsd typescript definition checker ([GitHub](https://github.com/tsdjs/tsd)).
  - It is invoked via `yarn type-tests`, and subsequently via `yarn test` in the core package.
- I've [added](https://github.com/zapier/zapier-platform/pull/818/files#diff-66e823ff018a5c1055a1aae77d261bc74813718e4fe6fd107ac4697f90c86655R215-R225) `BeforeRequestMiddleware` and `AfterResponseMiddleware` types to `zapier.custom.d.ts` module.
  - These are then referenced and replace the plain `MiddlewaresSchema` for the respective types via a transformer for each.
- I've gotten a lot more tests working. Now it's just a matter of getting the `typescript` template to pass it's smoke test.
  - I'm having a lot of trouble getting more details out of the `zapier init` function this is running. I think that output is itself for some spawned subprocesses it's running.

eliangcs (commented):
@rnegron already gave a pretty detailed review, so I'm going to slack off and only test it from a user's perspective. :)

I like how you separate the tool as an independent package, so I don't feel the need to review the code in detail as long as it works. The generated type file is beautiful; it has all the comments, linking to schema docs.

I noticed there are a lot of changes in `yarn.lock` that merely replaced `registry.yarnpkg.com` with `registry.npmjs.org`. Can we avoid them?

tkcranny:
@rnegron:

> In terms of the developer experience of generating the typing declarations file, I saw the precommit script in package.json but it doesn't seem to yarn (install the package requirements) beforehand. I suppose for most contributions to this repo, changes would not need to be made to this package and therefore it should be almost transparent that the typing is auto-generated before a commit happens in zapier-platform-core (and zapier-platform-schema!) What about adding a lifecycle step to the core (and schema?) package to ensure that schema-to-ts is ready to run on changes to that package? There are a [lot of available lifecycle options](https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-operation-order) for npm scripts 😵.

I love this idea. I've tried a couple of things to this effect. What I've found by putting in lifecycle hooks into the Core/Schema packages to ensure schema-to-ts is installed though breaks both the node 16 tests, as well as the smoke-tests when the packages are built and used on their own without the schema-to-ts missing from the expected location.

Fortunately `yarn install` from the root of the repo _will_ prepare schema-to-ts's own `node_modules` in preparation for the transparent commit hooks that generate the output declarations.

Therefore the biggest issue I see is that people with this repo _already_ checked out will just need to run `yarn` from the root to get this tool to work. Fresh installs ought to be fine because of this too.

Is that acceptable for you?

eliangcs (approved):
Thanks for fixing the yarn.lock changes! Disclaimer: I didn't review the code in depth, but it seems to work well, so I'm approving. I'd see @rnegron's vote before merging.

rnegron (approved):

xavdid:
@tkcranny Congrats on getting this out the door! I was (apparently) still subscribed to the original issue so I got the email that it had been closed. This is an awesome approach; color me quite impressed. 🎉

PR #821, opened 2024-07-16
docs(schema-to-ts): [PDE-5969] Minor doc updates to reference schema-to-ts package

rnegron:
* ~~Added `schema-to-ts` to `workspaces` key in the root `package.json`, so that running `yarn` from the root of the repo installs `schema-to-ts` packages as well.~~
  * Without this, running `yarn` and then `yarn test` would result in errors around the `vitest` package missing as part of the `(cd schema-to-ts && yarn test)` part.
* Updated some docs to mention `schema-to-ts`

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -26,5 +26,6 @@ where:
     * schema
     * core
     * legacy-scripting-runner
+    * schema-to-ts
 
 -->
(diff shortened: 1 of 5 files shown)


rnegron:
Adding `schema-to-ts` in the `workspaces` key in `package.json` causes the Node.js 16 testing issue [mentioned here](https://github.com/zapier/zapier-platform/pull/818#issuecomment-2227710770) but running `yarn install` from the root of the repo does not install `schema-to-ts`'s dependencies.

The ideal solution is to drop Node.js 16 and add `schema-to-ts` to `workspaces`.. I'll add a ticket for this so we don't forget. In the meantime, I've updated the `CONTRIBUTING.md` file.

standielpls (approved):

PR #824, opened 2024-07-18
PDE-5184 fix(core) - Add base64 encoding before uploading to s3

standielpls:
Currently when using this autostashing payload feature, if the perform call returns non-ascii characters, it won't stash. This MR encodes the data before uploading and we can decode on the way out. 

Planning to release this on 15.11.1, so the backend should account for this by only decode when zpc@15.11.1.


kreddlear (approved):
This looks good to me! Is there a corresponding ticket for the backend decoding update?

PR #827, opened 2024-07-25
feat(cli): Handle subheadings when interpreting the changelog

the-csaba:
Hi,

This is a simple attempt to add support for the 'Keep a Changelog' format (https://keepachangelog.com). It is a popular changelog format, and the current parser returns an error if someone wants to use it.


eliangcs (approved):
Nice improvement, thanks! Merging.

PR #835, opened 2024-08-07
fix(schema-to-ts): Simplify Array types (IQQ-1490)

tkcranny:
This removes length-enforced tuples types with spread parameters that are difficult and annoying to write in separate declarations.

## Example:

```typescript
const choices: [FieldChoiceWithLabel, ...FieldChoiceWithLabel[] = [...]
```

becomes just:

```typescript
const choices: FieldChoiceWithLabel[] = [...]
```

cc @cvanderkolk

### DIFF
--- a/packages/core/types/zapier.generated.d.ts
+++ b/packages/core/types/zapier.generated.d.ts
@@ -185,7 +185,7 @@ export type RefResource = string;
  */
 export type FieldChoices =
   | { [k: string]: unknown }
-  | [string | FieldChoiceWithLabel, ...(string | FieldChoiceWithLabel)[]];
+  | (string | FieldChoiceWithLabel)[];
 
 /**
  * Defines a field an app either needs as input, or gives as output.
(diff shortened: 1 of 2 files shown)


eliangcs (approved):

packages/core/types/zapier.generated.d.ts
```diff
@@ -290,10 +290,8 @@ export interface Field {
    * sub-object for this field. Usually used for line items.
    *
    * @minItems 1
-   *
-   * @minItems 1
    */
-  children?: [Field, ...Field[]];
+  children?: Field[];
```
szchenghuang, line 294: Nice optimization!

packages/core/types/zapier.generated.d.ts
```diff
@@ -185,7 +185,7 @@ export type RefResource = string;
  */
 export type FieldChoices =
   | { [k: string]: unknown }
```
szchenghuang, line 187: I know this line is out of scope. Just thinking out loud here. It can be written as `Record<string, unknown>` alternatively.

PR #838, opened 2024-08-08
feat(cli): Introduce zapier pull command

pragmatic-zac:
Adds a new CLI command:
`zapier pull` 

This command pulls down the source integration from the remote Zapier server.
Users may want to use this if the integration has changed since the last push. If changes were made by Zapier, users must run `zapier pull` before making further production level changes. Users are still free to push to new versions. 

**How it works**
Users will be prompted if they want to proceed, warning that we may overwrite existing files if the differ from the remote source. 
For each conflicting file, users may evaluate if they want to overwrite it or not. Each file can be subjected to a diff of the changes as well. 
Once pull is completed, the user can make production level changes.


packages/cli/src/utils/files.js
```diff
@@ -200,9 +202,49 @@ const makeTempDir = () => {
   return workdir;
 };
 
+async function deleteUnmatchedFiles(src, dest, ig) {
```
pragmatic-zac, line 205: This function works! 

....until you have a gitignored directory (as you can see in the below TODO). So this'll need a refactor. My current thought is instead of taking a source and destination as params, these should be a source and destination list/array. Then they can be filtered, more closely matching the pattern we already have in `makeSourceZip` where we first list the files [then filter them](https://github.com/zapier/zapier-platform/blob/95680f62897064853fb147350d3a183b85509a2a/packages/cli/src/utils/build.js#L248) through gitignored and blacklisted.

packages/cli/src/oclif/commands/pull.js
```diff
@@ -0,0 +1,43 @@
...
+    await downloadSourceZip();
+    validateFileExists(constants.SOURCE_PATH);
+
+    const tmpDir = makeTempDir();
+    debug('Using temp directory for source unzip: ', tmpDir);
+
+    const zip = new AdmZip(constants.SOURCE_PATH)
+    zip.extractAllTo(tmpDir, true)
+
+    const gitIgnorePatterns = await getGitIgnorePatterns('example-target');
```
pragmatic-zac, line 29: Everywhere you see 'example-target' is just used for testing. We'll remove this prior to merge.

packages/cli/src/oclif/commands/pull.js
```diff
@@ -0,0 +1,55 @@
...
+const { listFiles } = require('../../utils/build');
+const { deleteUnignoredFiles } = require('../../utils/pull');
+const { buildFlags } = require('../buildFlags');
+
+class PullCommand extends ZapierBaseCommand {
+  async perform() {
+    if (
+      !(await this.confirm(
+        colors.yellow(
+          'This will overwrite your local integration files with the latest version. Continue?'
```
eliangcs, line 22: This command is destructive. I'd expect it to ask for confirmation on which file will be overwritten or deleted before proceeding.

For example, if you run `zapier init` twice with the same directory, you will get this:



It's done using [yeoman-generator](https://github.com/zapier/zapier-platform/blob/efe3dad7806b57e3a732a595ee744f2340f4e890/packages/cli/src/generators/index.js#L5). The library provides an ["in-memory" file system](https://yeoman.io/authoring/file-system#:~:text=An%20%E2%80%9Cin%20memory%E2%80%9D%20file%20system), which has helped us do the "ask for confirmation" thing. Can we leverage it here too?

packages/cli/src/utils/api.js
```diff
@@ -13,11 +13,13 @@ const AdmZip = require('adm-zip');
...
 const path = require('path');
 
-const { writeFile, readFile } = require('./files');
+const { writeFile, readFile, ensureDir } = require('./files');
 
 const { prettyJSONstringify, startSpinner, endSpinner } = require('./display');
 
 const { localAppCommand } = require('./local');
+const { promisify } = require('./promisify');
+const { pipeline } = require('stream');
```
eliangcs, line 22: Node.js v15+ has builtin [Streams Promises API](https://nodejs.org/api/stream.html#streams-promises-api), so we don't need `./promisify`.

```suggestion
const { pipeline } = require('stream/promises');
```

packages/cli/src/utils/api.js
```diff
@@ -364,6 +366,61 @@ const validateApp = async (definition) => {
...
+    method: 'GET',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json; charset=utf-8',
+      'User-Agent': `${constants.PACKAGE_NAME}/${constants.PACKAGE_VERSION}`,
+      'X-Requested-With': 'XMLHttpRequest',
+      'X-Deploy-Key' : credentials[constants.AUTH_KEY]
+    },
+  };
+  const response = await fetch(url, requestOptions);
```
eliangcs, line 396: There's a `callAPI()` function we can use to make API requests. So we don't need to deal with authentication or headers here. If `callAPI()` can't be used to download a binary file, feel free to change it so it supports that.

It's also a bit strange that we're expecting a zip file while the headers include things like `Accept: 'application/json'` and `X-Requested-With: 'XMLHttpRequest'`.

packages/cli/src/utils/api.js
```diff
@@ -364,6 +366,61 @@ const validateApp = async (definition) => {
...
+  }
+
+  const contentType = response.headers.get('content-type');
+  if (!contentType || !contentType.includes('application/zip')) {
+    throw new Error('The API did not return a ZIP file');
+  }
+
+  const sourceZipPath = constants.SOURCE_PATH;
+  const appDir = process.cwd();
+  const dirPath = path.resolve(appDir, constants.BUILD_DIR);
```
eliangcs, line 409: This can overwrite an existing `build/source.zip`, which isn't ideal. It'd be better if we download it to a temporary directory (using `os.tmpdir()`).

packages/cli/src/utils/pull.js
```diff
@@ -0,0 +1,49 @@
+// prob put this in files
+const fs = require('fs');
+const path = require('path');
+
+const constants = require('../constants');
+const { respectGitIgnore } = require('./build');
+
+const isBlacklisted = (filePath) => {
+  return constants.BLACKLISTED_PATHS.find((excluded) => {
```
eliangcs, line 9: While we're at it, let's rename the constant to something like `EXCLUDED_PATHS`.

eliangcs (commented):
Nice POC! To make it production ready, I think we should be more careful about overwriting or deleting files. I believe we can use [Yeoman](https://yeoman.io) to build a better CLI interface, as we already used it for `zapier init`.

packages/cli/src/utils/pull.js
standielpls: Ah nice I had the same thought.

eliangcs (commented):
I found that the current implementation only copies the top-level files. For example, my source zip file had a `src/triggers/newProject.ts` file, and `zapier pull` didn't copy it to the destination.

packages/cli/src/utils/api.js
```diff
@@ -90,16 +96,21 @@ const callAPI = (
       return fetch(_requestOptions.url, _requestOptions);
     })
     .then((res) => {
-      return Promise.all([res, res.text()]);
+      return Promise.all([res, isZipRes ? res.body : res.text()]);
     })
-    .then(([res, text]) => {
+    .then(([res, bodyOrText]) => {
+      const text = bodyOrText;
```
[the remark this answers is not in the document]
standielpls: :face_palm: ah yes that's right, I confused myself with the content header. Thanks for catching

standielpls:
> I found that the current implementation seems to only copy top-level files. For example, my source zip file had a file named src/triggers/newProject.ts, and zapier pull didn't copy it to the destination.

I think I address this in a new commit https://github.com/zapier/zapier-platform/pull/838/commits/bd6fe96598f636928cf8784438686cecb71dca35 - cant reproduce this.

packages/cli/src/generators/pull.js
```diff
@@ -0,0 +1,54 @@
...
+
+const maybeOverwriteFiles = async (gen) => {
+  const dstDir = gen.options.dstDir;
+  const srcDir = gen.options.srcDir;
+  for (const file of gen.options.sourceFiles) {
+    gen.fs.copyTpl(
+      gen.templatePath(path.join(srcDir, file)),
+      gen.destinationPath(path.join(dstDir, file)),
+      gen.options
+    );
```
eliangcs, line 15: 1. We only need `gen.fs.copy()` here because `copyTpl()` is for rendering a template.
2. `gen.templatePath()` and `gen.destinationPath()` are unnecessary because we're already dealing with absolute paths with `path.join()`.

```suggestion
    gen.fs.copy(path.join(srcDir, file), path.join(dstDir, file));
```

packages/cli/src/oclif/commands/pull.js
```diff
@@ -0,0 +1,52 @@
...
+      });
+    } finally {
+      removeDirSync(tmpDir);
+    }
+  }
+}
+
+PullCommand.flags = buildFlags();
+PullCommand.description =
+  'Pull the source code of the latest version of your integration from Zapier, overwriting your local integration files.';
```
eliangcs, line 50: We can set `skipValidInstallCheck` so `zapier pull` doesn't require developers to run `npm install` or `yarn` or equivalents.

```suggestion
  'Pull the source code of the latest version of your integration from Zapier, overwriting your local integration files.';
  PullCommand.skipValidInstallCheck = true;
```

packages/cli/src/generators/pull.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  prompting() {
+    const prompts = [
+      {
+        type: 'confirm',
+        name: 'confirm',
+        message: `Warning: You are about to overwrite existing files.
+
+Before proceeding, please make sure you have saved your work. Consider creating a backup or saving your current state in a git branch. During the process, you may abort anytime by pressing 'x'.
```
eliangcs, line 32: My editor found a trailing space:

```suggestion
Before proceeding, please make sure you have saved your work. Consider creating a backup or saving your current state in a git branch. During the process, you may abort anytime by pressing 'x'.
```

eliangcs (commented):
I think this should be good to merge after the following suggestions are addressed. Feel free to ask someone else for approval when I'm not around. Thanks!

packages/cli/src/utils/build.js
```diff
@@ -107,8 +106,8 @@ const requiredFiles = (cwd, entryPoints) => {
 };
 
 const listFiles = (dir) => {
-  const isBlacklisted = (filePath) => {
-    return constants.BLACKLISTED_PATHS.find((excluded) => {
+  const isBlocklisted = (filePath) => {
```
rnegron, line 109: nitpick: Isn't this duplicated in `utils/ignore.js` now?

packages/cli/src/utils/build.js
```diff
@@ -47,6 +45,7 @@ const {
 const checkMissingAppInfo = require('./check-missing-app-info');
 
 const { runCommand, isWindows, findCorePackageDir } = require('./misc');
+const { respectGitIgnore } = require('./ignore');
```
rnegron, line 48:
```suggestion
const { respectGitIgnore, isBlocklisted } = require('./ignore');
```

See below comment about duplication

PR #846, opened 2024-08-21
chore(types): Add `customOptions` field to z.request options

tkcranny:
This is a sanctioned object that can be used freely for extra middleware options etc. without needing to bend over backwards to support patching the `z.request` options

### DIFF
--- a/packages/core/types/zapier.custom.d.ts
+++ b/packages/core/types/zapier.custom.d.ts
@@ -95,6 +95,14 @@ export interface HttpRequestOptions {
   timeout?: number;
   url?: string;
   skipThrowForStatus?: boolean;
+
+  /**
+   * Custom options for the request. These do not affect the request
+   * itself, but can be used to pass data to middleware. For example,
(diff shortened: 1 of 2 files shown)


packages/core/types/zapier.custom.d.ts
```diff
@@ -95,6 +95,14 @@ export interface HttpRequestOptions {
...
   url?: string;
   skipThrowForStatus?: boolean;
+
+  /**
+   * Custom options for the request. These do not affect the request
+   * itself, but can be used to pass data to middleware. For example,
+   * this is recommended for things like `prefixErrorMessage` fields
+   * etc.
+   */
+  customOptions?: Record<string, any>;
```
eliangcs, line 105: This isn't an actual option provided by `z.request()`, so I'm not sure if we should include this. If we added this, we'd have to support it forever until we can make it a breaking change.

@tkcranny is there another way for you to get around this?

eliangcs (approved):
We discussed internally and settled on a rename to `middlewareData`.

PR #849, opened 2024-08-23
feat(core) - Add scopes as an option for zcache usage

standielpls:
Internal: Adds an scope option to z.cache. Currently it uses scope internally with user and auth only. We may want to exercise one or the other only so providing this flexibility from the interface.

### DIFF
--- a/packages/core/src/tools/create-cache.js
+++ b/packages/core/src/tools/create-cache.js
@@ -12,7 +12,7 @@ const createCache = (input) => {
     key,
     value = null,
     ttl = null,
-    scopes = null
+    scope = null
   ) => {
     if (!rpc) {
       throw new Error('rpc is not available');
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/create-cache.js
```diff
@@ -20,23 +26,32 @@ const createCache = (input) => {
       throw new TypeError('ttl must be an integer');
     }
 
+    if (
+      scopes != null &&
+      !scopes.every((scope) => scope === 'user' || scope === 'auth')
+    ) {
```
eliangcs, line 32: If `scopes` is a non-array object, the `TypeError` should be thrown as well. So:

```suggestion
    if (
      !Array.isArray(scopes) ||
      !scopes.every((scope) => scope === 'user' || scope === 'auth')
    ) {
```

eliangcs (dismissed):
Looks good to me except for a validation check. I assume we'll only merge this after the backend is ready?

standielpls:
:face_palm: My new tests all included mockRpcCall, even if they didn't make it pass the validation check, so I think the mockRpcCall was being used in other tests bc we never unmock.

rnegron (dismissed):

PR #854, opened 2024-08-30
refactor(core) Attempt to log http response size

standielpls:


eliangcs (approved):
There's a failing test because of the new log field. Otherwise, this looks good to me!

ksealmalaysia41-lang:
This logic bring the attacker down

PR #857, opened 2024-09-10
feat(core): Add node-fetch feature `signal` to request client [PDE-5367]

kreddlear:
As Cass reported [here](https://zapier.slack.com/archives/C5Z9BP4U9/p1725954040992739), `node-fetch` supports the `signal` option described here: https://www.npmjs.com/package/node-fetch#request-cancellation-with-abortsignal. Looks like this is supported after Node 8 and replaces `timeout` (although timeout still seems to be supported), according to these comments: https://github.com/node-fetch/node-fetch/tree/2.x?tab=readme-ov-file#options

However, we have not added that to the request options for `z.request`. This PR aims to add that per Fokke's suggestion that we may just need to add it to this picklist.


eliangcs (approved):

kreddlear:
Tested with a test integration in a Zap and confirmed the signal abort feature works properly ([screenshot](https://cdn.zappy.app/ac73673791b8e37696a1a9dd8142c42f.png)). Merging and will look into releasing.

PR #858, opened 2024-09-10
[PARTOPS-1224] docs(schema) - Add additional details to inputFormat property in fieldSchema

michellechu77:
Lefthook developers tried using the `inputFormat` field property for an action input field and the copy to clipboard field that renders in the Editor is not well suited for action step as it disables the ability to map a value to the field. Noting this in the docs explicitly. Slack thread with editor team [here](https://zapier.slack.com/archives/C02TQNXKEBF/p1725870227224329?thread_ts=1725653551.771619&cid=C02TQNXKEBF).


kreddlear (approved):
Your doc updates look great! What a weird field format option though 🤔

PR #861, opened 2024-09-12
feat(cli): Introduce zapier canary command

pragmatic-zac:
Adds a new command - `zapier canary` 🐦 

This command enables users to divert a portion of traffic from one app version to a different app version for a specified amount of time.

[Here's a gif](https://cdn.zappy.app/69b7f4f299034323fa2aa2a2b33b15a5.gif) of the command working...
- First, we create a canary sending 10% of traffic from version 1.0.0 to 1.0.1 for 120 seconds
- Then, we list all active canaries in a table
- Next, we delete the active canary
- Lastly, we re-list and see we've successfully deleted the canary


standielpls:
Nice this looks so promising! And great turnaround
my quick initial feedback is i was thinking more on the lines of sub-commands like `zapier env:get`, `zapier env:set` ([ref](https://github.com/zapier/zapier-platform/blob/e6e046588cfe8e6fcc75b6286608add0b8e321d8/packages/cli/src/oclif/commands/env/set.js#L50))

pragmatic-zac:
Ah gotcha, yeah I like the subcommands better! I'll swap those in instead

pragmatic-zac:
@standielpls just updated this to use subcommands so now we have `canary:create`, `canary:list`, and `canary:delete`. It's ready for a look whenever you have a chance 😃

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -0,0 +1,77 @@
...
+    if (!duration || duration <= 30 || duration > 24 * 60 * 60) {
+      this.error('Duration must be a positive number between 30 and 86400');
+    }
+  }
+}
+
+CanaryCreateCommand.flags = buildFlags({
+  commandFlags: {
+    versionFrom: flags.string({char: 'f', description: 'Version to route traffic from', required: true}),
+    versionTo: flags.string({char: 't', description: 'Version to canary traffic to', required: true}),
```
standielpls, line 61: Let's move the from/to versions as "arguments" - these are things the command should act on. ([migrate example](https://github.com/zapier/zapier-platform/blob/main/packages/cli/src/oclif/commands/migrate.js#L112))
Percent i think it's closer to a required argument as well, since this should be set by the user, I don't want to default to anything. 
Flags are treated more like options, and are things we can use to alter the behaviour (like duration).

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -0,0 +1,77 @@
...
+    this.stopSpinner();
+    this.log('Canary deployment created successfully.');
+  }
+
+  async findExistingCanary(versionFrom, versionTo) {
+    const activeCanaries = await listCanaries();
+    return activeCanaries.objects.find(c => c.from_version === versionFrom && c.to_version === versionTo) || null;
+  }
+
+  validateVersions(versionFrom, versionTo) {
```
standielpls, line 40: Should check if `versionFrom === versionTo` as well

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -0,0 +1,77 @@
...
+    versionTo: flags.string({char: 't', description: 'Version to canary traffic to', required: true}),
+    percent: flags.integer({char: 'p', description: 'Percent of traffic to route to new version', required: true}),
+    duration: flags.integer({char: 'd', description: 'Duration of the canary in seconds', required: true}),
+  },
+  opts: {
+    format: true
+  }
+});
+CanaryCreateCommand.description =
+  'Create a new canary deployment, diverting a specified percentage of traffic from one version to another for a specified duration.'
```
standielpls, line 70: I'd be a little bit more explicit here so it is clear to users what is going on.
```
Create a new canary deployment, diverting a specified percentage of traffic from one version to another for a specified duration. 

Only one canary can be active at the same time. You can run `zapier canary:list` to check. If you would like to create a new canary with different parameters, you can wait for the canary to finish, or delete it using `zapier canary:delete X Y`. 

Note: this is similar to `zapier migrate` but different in that this is temporary and will "revert" the changes once the specified duration is expired. 
```

standielpls: 🤔 I want to say something about breaking changes as well (1.X.Y to 2.X.Y) and advise against it, or to block it from letting the user do it entirely, what do you think?

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -0,0 +1,77 @@
...
+    const { versionFrom, versionTo, percent, duration } = flags;
+
+    this.validateVersions(versionFrom, versionTo);
+    this.validatePercent(percent);
+    this.validateDuration(duration);
+
+    const existingCanary = await this.findExistingCanary(versionFrom, versionTo);
+    if (existingCanary) {
+      const secondsRemaining = existingCanary.until_timestamp - Math.floor(Date.now() / 1000);
+      this.log(`A canary deployment already exists from version ${versionFrom} to version ${versionTo}, there are ${secondsRemaining} seconds remaining`);
```
standielpls, line 18: Maybe add a helper in there? 
```
If you would like to stop this canary now, run `zapier canary:delete {versionFrom} {versionTo}` 
```

packages/cli/src/oclif/commands/canary/delete.js
```diff
@@ -0,0 +1,52 @@
...
+    return activeCanaries.objects.find(c => c.from_version === versionFrom && c.to_version === versionTo) || null;
+  }
+
+  validateVersions(versionFrom, versionTo) {
+    this.throwForInvalidVersion(versionFrom);
+    this.throwForInvalidVersion(versionTo);
+  }
+}
+
+CanaryDeleteCommand.flags = {
```
standielpls, line 41: Similar to create, can we make these into arguments

packages/cli/src/oclif/commands/canary/delete.js
```diff
@@ -0,0 +1,52 @@
...
+
+    this.validateVersions(versionFrom, versionTo);
+
+    const existingCanary = await this.findExistingCanary(versionFrom, versionTo);
+    if (!existingCanary) {
+      this.log(`There is no active canary from version ${versionFrom} to version ${versionTo}`);
+      return;
+    }
+
+    const confirmed = await this.confirm(`Are you sure you want to delete the canary from ${versionFrom} to ${versionTo}?`);
```
standielpls, line 18: Nice

packages/cli/src/oclif/commands/canary/delete.js
```diff
@@ -0,0 +1,52 @@
...
+
+    this.startSpinner(`Deleting active canary from ${versionFrom} to ${versionTo}`);
+    await deleteCanary(versionFrom, versionTo);
+    this.stopSpinner();
+    this.log('Canary deployment deleted successfully.');
+  }
+
+  async findExistingCanary(versionFrom, versionTo) {
+    const activeCanaries = await listCanaries();
+    return activeCanaries.objects.find(c => c.from_version === versionFrom && c.to_version === versionTo) || null;
```
standielpls, line 32: Do we need `|| null` isn't .find() falsy if not found already ?

packages/cli/src/oclif/commands/canary/delete.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  validateVersions(versionFrom, versionTo) {
```
standielpls, line 35: same, check if version from and to are the same

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -0,0 +1,77 @@
...
+
+CanaryCreateCommand.flags = buildFlags({
+  commandFlags: {
+    versionFrom: flags.string({char: 'f', description: 'Version to route traffic from', required: true}),
+    versionTo: flags.string({char: 't', description: 'Version to canary traffic to', required: true}),
+    percent: flags.integer({char: 'p', description: 'Percent of traffic to route to new version', required: true}),
+    duration: flags.integer({char: 'd', description: 'Duration of the canary in seconds', required: true}),
+  },
+  opts: {
+    format: true
```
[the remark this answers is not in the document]
pragmatic-zac: Removed, thanks for catching that - it was a remnant from using a table for formatting

packages/cli/src/oclif/commands/canary/create.js
pragmatic-zac: 👍 I'd initially put them in flags because I thought it was little more descriptive, but I see what you mean. I moved everything to args

Do we want `duration` to be a flag though? Seems like a required arg to me, I'd prefer the user to be explicit about what they're wanting to do

packages/cli/src/oclif/commands/canary/delete.js
pragmatic-zac: Nope, removed

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -0,0 +1,97 @@
...
+
+    this.validateVersions(versionFrom, versionTo);
+    this.validatePercent(percent);
+    this.validateDuration(duration);
+
+    const activeCanaries = await listCanaries();
+    if (activeCanaries.objects.length > 0) {
+      const existingCanary = activeCanaries.objects[0];
+      const secondsRemaining = existingCanary.until_timestamp - Math.floor(Date.now() / 1000);
+      this.log(`A canary deployment already exists from version ${existingCanary.from_version} to version ${existingCanary.to_version}, there are ${secondsRemaining} seconds remaining.
```
pragmatic-zac, line 16: @standielpls one thing I'll call out here - since the API currently only allows one canary at a time, I've changed this code to block creation of a new one if _any_ canary exists. Initially it checked for a canary matching the `versionTo` and `versionFrom` args, but I think this is simpler and offers less potential for confusion

So tldr, only one canary can exist at a time

pragmatic-zac: I also don't super love the `activeCanaries.objects[0]` but the API returns the array right now, so 🤷

packages/cli/src/oclif/commands/canary/create.js
standielpls: sounds good to me

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -32,46 +32,66 @@ class CanaryCreateCommand extends ZapierBaseCommand {
...
+
 CanaryCreateCommand.description =
-  'Create a new canary deployment, diverting a specified percentage of traffic from one version to another for a specified duration.'
+  'Create a new canary deployment, diverting a specified percentage of traffic from one version to another for a specified duration. \n' +
+  '\n' +
+  'Only one canary can be active at the same time. You can run `zapier canary:list` to check. If you would like to create a new canary with different parameters, you can wait for the canary to finish, or delete it using `zapier canary:delete a.b.c x.y.z`. \n' +
+  '\n' +
+  'Note: this is similar to `zapier migrate` but different in that this is temporary and will "revert" the changes once the specified duration is expired. \n' +
+  '\n' +
+  '**Only use this command to canary traffic between non-breaking versions!**';
```
standielpls, line 89: Can we use the same pattern as migrate and use string literals 
```
CanaryCreateCommand.description = `Create a new canary deployment ....
You can run \`zapier canary:list\` 
`
```

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -32,46 +32,66 @@ class CanaryCreateCommand extends ZapierBaseCommand {
...
+
 CanaryCreateCommand.examples = [
-  'zapier canary:create --versionFrom=1.0.0 --versionTo=1.1.0 --percent=25 --duration=720',
-  'zapier canary:create -f 2.0.0 -t 2.1.0 -p 50 -d 300'
-]
+  'zapier canary:create 1.0.0 1.1.0 25 720',
```
standielpls, line 92: Idk if this is a problem or not, but the percent and duration numbers can get mixed up easily. It'd be bad if they wanted 100s of canary and 30% traffic, but it was in the wrong order, creating a canary that is 100% traffic for 30s. 

So I think it'd be nice to either 
1. Show the summary upfront but ask for confirmation
2. turn percent/duration into options (so it can be passed in like -p 25, -d 30 for example) 

I'm fine with either option

standielpls (commented):
NICE - just a couple more small things

standielpls (approved):
woooo!

PR #862, opened 2024-09-17
PDE-4994 chore: release 15.15.0

eliangcs:
I moved many of the older changelogs to a new directory `changelog/`, because the main `CHANGELOG.md` is getting too large.

The changelog for 15.15.0 is as follows.

## Changelog

_released `2024-09-18`_

This release introduces "[Buffered Create Actions](https://github.com/zapier/zapier-platform/blob/8353f32cba12da2a845419f0024b0029090437d9/packages/cli/README.md#buffered-create-actions)", currently only available for **internal** use. A Buffered Create allows you to create objects in bulk with a single or fewer API request(s). This is useful when you want to reduce the number of requests made to your server. When enabled, Zapier holds the data until the buffer reaches a size limit or a certain time has passed, then sends the buffered data using the `performBuffer` function you define.

### cli

None!

### core

- :test_tube: Add checks to support Buffered Create Actions ([#832](https://github.com/zapier/zapier-platform/pull/832))

### schema

- :test_tube: Add `performBuffer` and `buffer` to `operation` schema to support Buffered Create Actions ([#832](https://github.com/zapier/zapier-platform/pull/832))


standielpls (dismissed):
Wonder if we should leave a note in the main changelog about where to find older changelog files. But the changelog folder name pretty self-documenting, so it should be pretty simple to find.

eliangcs:
> Wonder if we should leave a note in the main changelog about where to find older changelog files.

@standielpls I did that by adding some HTML [anchors](https://github.com/zapier/zapier-platform/blob/62e9cd631b546aac73d43b67e4d8195c6c488e79/CHANGELOG.md?plain=1#L751-L846) at the bottom. So for example, when a user clicks a link https://github.com/zapier/zapier-platform/blob/62e9cd631b546aac73d43b67e4d8195c6c488e79/CHANGELOG.md#1000, they will be redirected to the Old Releases section at the bottom.

szchenghuang (approved):

szchenghuang (approved):

PR #874, opened 2024-10-01
fix(core): Throw a descriptive error for too-long encoded filenames [PDE-5415]

kreddlear:
S3 only accepts up to 2048 bytes in metadata. When the `Content-Disposition` header is built, if the filename needs to be encoded, it is included in the header both as-is and in its encoded form, which can overload the S3 metadata and cause a confusing error for users.

Filenames that are not encoded are much less likely to be an issue, since they will not be duplicated in the `Content-Disposition` header.

This PR throws a descriptive error if the encoded filename is very long and therefore likely to cause an S3 error, to make it easier for users and/or devs to handle this situation.


packages/core/src/tools/create-file-stasher.js
```diff
@@ -186,6 +190,24 @@ const ensureUploadMaxSizeNotExceeded = (streamOrData, length) => {
   }
 };
 
+// S3's max metadata size is 2KB
+// If the filename needs to be encoded, both the filename
+// and encoded filename are included in the Content-Disposition header
+const ensureMetadataMaxSizeNotExceeded = (filename) => {
+  const filenameMaxSize = ENCODED_FILENAME_MAX_LENGTH;
```
rnegron, line 197: nitpick: looks like we only use this variable once, as we directly use `ENCODED_FILENAME_MAX_LENGTH` for the `Error`, could we remove this variable?

rnegron (approved):

packages/core/test/tools/file-stasher.js
```diff
@@ -178,6 +182,23 @@ describe('file upload', () => {
...
   });
 
+  it('should fail a file with a too-long encoded text filename', async () => {
+    mockRpcGetPresignedPostCall('8888/new.txt');
+    mockUpload();
+
+    const file = Buffer.from('hello world this is a buffer of text');
+    // length 1026
+    const filename =
+      '太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了太長了';
```
eliangcs, line 192: 😂

eliangcs (approved):

PR #878, opened 2024-10-07
PDE-5323 feat(cli): `invoke auth refresh` subcommand

eliangcs:
Changes:

- add `invoke auth refresh` subcommand
- add `invoke` command to smoke tests
  - while I was there, I had to fix an issue with the `dynamic-dropdown` project template, where https://swapi.dev/api/people can have empty `species` if the person is a human.
- add `--local-port` flag for `auth start` subcommand so that the local port can be different from the one you specify in `--redirect-uri`, which is essential if you're doing port forwarding
- fix an issue where `--inputData @file.json` didn't really work
  - turns out if you're using `fs` Promises API (`require('node:fs/promises')`), you can't use `fs.createReadStream()`; you need to open the file first with `const fd = await fs.open()` and then `fd.createReadStream()`
- fix an issue where the rendered HTML command docs had redundant line breaks

The `zapier invoke auth refresh` command is available for `session` and `oauth2` auth types.

For `oauth2`, it invokes `authentication.oauth2Config.refreshAccessToken` using the existing auth data in the `.env` file. It gets the new auth data and appends it to `.env`.

For `session` auth, it's the same except for it invokes `authentication.sessionConfig.perform`.


packages/cli/src/oclif/commands/invoke.js
```diff
@@ -972,7 +1025,8 @@ class InvokeCommand extends BaseCommand {
           if (filePath === '-') {
             inputStream = process.stdin;
           } else {
-            inputStream = fs.createReadStream(filePath, { encoding: 'utf8' });
+            const fd = await fs.open(filePath);
+            inputStream = fd.createReadStream({ encoding: 'utf8' });
```
standielpls, line 1029: Interesting, is this an edge case? I definitely recalled testing the input file option and it worked ok

standielpls (approved):
I didn't get a chance to test the refresh option, but the code looks ok! Default to approving, if you want a full manual test for all the options, happy to help conduct one as well!

packages/cli/src/oclif/commands/invoke.js
eliangcs: I changed `require('node:fs')` to `require('node:fs/promises')` in the `auth start` [PR](https://github.com/zapier/zapier-platform/pull/872), so it was working and then was broken by the `auth start` PR.

PR #883, opened 2024-10-10
[PDE-5395] feat(schema): Add `meta` field support for FieldSchema

rnegron:
Adds support for static input fields to via a new `FieldMetaSchema`. Note that the `FieldMetaSchema` only allows for primitive values (no nested objects).


packages/core/types/zapier.generated.d.ts
```diff
@@ -323,6 +333,12 @@ export interface Field {
    * "https://{{input}}.yourdomain.com").
    */
   inputFormat?: string;
+
+  /**
+   * Allows for additional metadata to be stored on the field.
+   * Supports simple key-values only (no sub-objects or arrays).
+   */
+  meta?: FieldChoices;
```
zapzap, line 341: assuming this should be `FieldMeta` instead of `FieldChoices`?

cvanderkolk, line 341: assuming this should be `FieldMeta` instead of `FieldChoices`?

rnegron: Yes, good call! I am not sure why the tooling keeps generating `FieldChoices` 🤔 will take a look!

rnegron: Oh, it was an obvious typo 🤦🏻 Fixed!

kola-er (approved):

packages/core/types/zapier.generated.d.ts
```diff
@@ -187,6 +187,16 @@ export type FieldChoices =
...
   | (string | FieldChoiceWithLabel)[];
 
+/**
+ * Allows for additional metadata to be stored on the field. Only
+ * simple values are allowed (no objects or arrays)
+ *
+ * [Docs: FieldMetaSchema](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#FieldMetaSchema)
+ */
+export interface FieldMeta {
+  [k: string]: string | number | boolean | null;
```
FokkeZB, line 197: Can we drop `null`? The more strict the better.

```suggestion
  [k: string]: string | number | boolean;
```

packages/schema/exported-schema.json
```diff
@@ -149,6 +149,14 @@
         }
       ]
     },
+    "FieldMetaSchema": {
+      "id": "/FieldMetaSchema",
+      "type": "object",
+      "description": "Allows for additional metadata to be stored on the field. Only simple values are allowed (no objects or arrays)",
+      "additionalProperties": {
+        "type": ["string", "number", "boolean", "null"]
```
FokkeZB, line 157:
```suggestion
        "type": ["string", "number", "boolean"]
```

packages/schema/lib/schemas/FieldMetaSchema.js
```diff
@@ -0,0 +1,23 @@
...
+
+const makeSchema = require('../utils/makeSchema');
+
+module.exports = makeSchema({
+  id: '/FieldMetaSchema',
+  type: 'object',
+  description:
+    'Allows for additional metadata to be stored on the field. Only simple values are allowed (no objects or arrays)',
+  additionalProperties: {
+    type: ['string', 'number', 'boolean', 'null'],
```
FokkeZB, line 11:
```suggestion
    type: ['string', 'number', 'boolean'],
```

FokkeZB (commented):
I suggest dropping `null` - the more strict the better.

packages/core/types/zapier.generated.d.ts
FokkeZB, line 1: We should also update the TS definition for `bundle.meta` at https://github.com/zapier/zapier-platform/blob/1126671b795a6e9d20b8df47617a8ba4acf73fb8/packages/core/types/zapier.custom.d.ts#L34

FokkeZB (approved):

packages/schema/lib/schemas/FieldSchema.js
```diff
@@ -137,6 +139,11 @@ module.exports = makeSchema(
         // TODO: Check if it contains one and ONLY ONE '{{input}}'
         pattern: '^.*{{input}}.*$',
       },
+      meta: {
+        description:
+          'Allows for additional metadata to be stored on the field. Supports simple key-values only (no sub-objects or arrays).',
+        $ref: FieldMetaSchema.id,
+      },
```
kola-er, line 146:
```suggestion
      meta: {
        description:
          'Allows for additional metadata to be stored on the field. Supports simple key-values only (no sub-objects or arrays).',
        $ref: FlatObjectSchema.id,
      },
```

The [/FlatObjectSchema](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#flatobjectschema) does the work without the need for describing a separate `FieldMetaSchema`.

kola-er: Just came across it.

packages/schema/lib/schemas/FieldMetaSchema.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const makeSchema = require('../utils/makeSchema');
+
+module.exports = makeSchema({
+  id: '/FieldMetaSchema',
+  type: 'object',
+  description:
+    'Allows for additional metadata to be stored on the field. Only string, numeric or boolean values are allowed',
```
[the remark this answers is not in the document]
kola-er: Not necessary though

packages/schema/lib/schemas/FieldSchema.js
kola-er: Realizing that Fokke suggested we don't include `null` in the valid values, can we make the `FieldMetaSchema` to be similar to `FlatObjectSchema` at least? To have its properties displayed in a table like that of the `FlatObjectSchema`.

packages/schema/lib/schemas/FieldMetaSchema.js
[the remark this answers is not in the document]
rnegron: The `additionalProperties` property for the schema means that this can be statically validated! For example, check out the error from running `zapier validate` with the following inputField:
```js
 {
        key: "my_key",
        label: "My Key",
        type: "string",
        required: false,
        meta: {
          subObject: { greeting: "hello!" },
        },
}
```

image

packages/schema/lib/schemas/FieldMetaSchema.js
```diff
@@ -0,0 +1,27 @@
...
+  additionalProperties: {
+    type: ['string', 'number', 'boolean'],
```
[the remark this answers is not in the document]
rnegron: Great question; this is part of the [JSONSchema spec](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties) and allows us to enforce typing on object values. We can't use `properties` to constrain the keys of this `meta` object since a builder should be able to define keys as they like, but we can use `additionalProperties` to make sure the objects types are limited (and can't be other objects, for example).

Specifically, check out this section of the docs linked above: 
Screen Shot 2024-11-20 at 5 51 08 PM

packages/schema/lib/schemas/FieldSchema.js
rnegron: Interesting, I didn't know about the `anyOf` property in combination with `patternProperties`! Here's how it statically validates against a sub-object (same inputField as the one mentioned above):

Screen Shot 2024-11-20 at 5 58 32 PM

I've updated the PR to use this mechanism! Good call out, thanks!

rnegron: Here's how the Schema docs look now:
Screen Shot 2024-11-20 at 6 01 54 PM

packages/schema/lib/schemas/FieldMetaSchema.js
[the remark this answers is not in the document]
kola-er: Thanks

kola-er (approved):

kola-er (approved):

PR #890, opened 2024-10-14
[PDE-5214] chore(cli): Oclif upgraded to v4

szchenghuang:
This PR primarily upgrades Oclif from v1 to v4. The primary commit is 0e421a55d5fbfed459a630da94c13688b57205a6.

### Command flags are defined differently.
`flags` from `@oclif/command` is replaced with `Flags` from `@oclif/core`.

#### Before
```
const { flags } = require('@oclif/command');
flags.boolean({ ... });

```

#### After
```
const { Flags } = require('@oclif/core');
Flags.boolean({ ... });
```

#### Command args are defined differently.
It's an object keyed by arg names rather than an array previously.

#### Before
```
SomeCommand.args = [{
  name: 'somArgName',
  ...
}];
```

#### After
```
SomeCommand.args = {
  someArgName: {
    ...
  },
};
```

### `yarn outdated --depth=3`
Outdated Oclif dependencies should be gone.

 Other commits are supporting changes.


packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -1,6 +1,4 @@
-const { Command } = require('@oclif/command');
-const { stdtermwidth } = require('@oclif/help/lib/screen');
-const { renderList } = require('@oclif/help/lib/list');
```
szchenghuang, line 3: `renderList` has been moved from `@oclif/help` to `@oclif/core`, but it can't be imported anymore.

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -149,18 +151,10 @@ class ZapierBaseCommand extends Command {
...
+      console.log(formatter(rows, headers, showHeaders, hasBorder, style));
     }
   }
 
-  /**
-   * Print text in a list style.
-   * @param {string[][]} items
-   */
-  logList(items) {
-    this.log(renderList(items, { spacer: '\n', maxWidth: stdtermwidth }));
```
szchenghuang, line 161: Its only use case was logging a table-like list in the `validate` command. Since we no longer have `renderList`, `logList` is replaced by `logTable`.  The table-like list is really a table after all.

packages/cli/src/tests/baseCommand.js
```diff
@@ -22,22 +22,52 @@ class TestBaseCommand extends BaseCommand {
...
 }
-// logs both text and a table, which behave differently baed on the format flag
-class TestSampleCommand extends TestBaseCommand {
+
+class WithArgsCommand extends TestBaseCommand {
+  async perform() {
+    this.log(MESSAGE);
+  }
+}
+WithArgsCommand.args = {
```
szchenghuang, line 31: A test command to validate command arguments.

packages/cli/src/tests/baseCommand.js
```diff
@@ -22,22 +22,52 @@ class TestBaseCommand extends BaseCommand {
...
+  name: Args.string({
+    description: 'name is a required argument',
+    required: true,
+  }),
+  number: Args.string({
+    description: 'number is an optional argument',
+  }),
+};
+
+class WithFlagsCommand extends TestBaseCommand {
```
szchenghuang, line 41: A test command to validate command flags.

packages/cli/src/utils/analytics.js
```diff
@@ -20,7 +20,7 @@ const shouldSkipAnalytics = (mode) =>
   process.env.DISABLE_ZAPIER_ANALYTICS ||
   mode === ANALYTICS_MODES.disabled;
 
-const recordAnalytics = async (command, isValidCommand, args, flags) => {
+const recordAnalytics = async (command, isValidCommand, argNames, flags) => {
```
szchenghuang, line 23: `recordAnalytics` is called at [ZapierBaseCommand.js](https://github.com/zapier/zapier-platform/pull/890/files#diff-3b23b2f6b3353d797bd2412d62eea92f69d33a48ef839325edea158cb4c432b9R349).`Object.keys(this.args)` returns an array of argument names rather than the arguments per se.

packages/cli/src/oclif/buildFlags.js
```diff
@@ -1,25 +1,24 @@
...
-  format: flags.string({
+  debug: Flags.boolean({
+    char: 'd',
+    description: 'Show extra debugging output.',
+    // pull from env?
+  }),
+  format: Flags.string({
     char: 'f',
     options: Object.keys(formatStyles),
     default: 'table',
```
rnegron, line 15: Curious if [this change](https://github.com/oclif/core/blob/main/guides/PRE_CORE_MIGRATION.md#update-default-property-on-flag-definitions) around `default` build `async` now applies to even these literal values? 🤔

packages/cli/src/oclif/commands/versions.js
```diff
@@ -23,6 +25,47 @@ class VersionCommand extends BaseCommand {
...
+          version: `- ${colors.bold('Warnings')}`,
+          platform_version:
+            "Issues and recommendations that need human reviews by Zapier before publishing your integration. They don't block.",
+        },
+      ],
+      hasBorder: false,
+      style: { head: [], 'padding-left': 0, 'padding-right': 0 },
+    });
+
+    this.logList([
```
rnegron, line 51: I think this call should be deleted! 😄

packages/cli/src/oclif/ZapierBaseCommand.js
rnegron: I found another use case in `oclif/commands/versions`!

docs/cli.html
```diff
@@ -434,8 +434,8 @@ <h2 id="canarylist">canary:list</h2>
       <blockquote>
 <p>List all active canary deployments</p>
 </blockquote><p><strong>Usage</strong>: <code>zapier canary:list</code></p><p><strong>Flags</strong></p><ul>
-<li><code>-f, --format</code> | Change the way structured data is presented. If &quot;json&quot; or &quot;raw&quot;, you can pipe the output of the command into other tools, such as jq. One of <code>[plain | json | raw | row | table]</code>. Defaults to <code>table</code>.</li>
```
rnegron, line 437: these doc changes would go away after a merge from `main`? maybe?

packages/cli/package.json
```diff
@@ -19,12 +19,12 @@
     "/oclif.manifest.json"
   ],
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
```
rnegron, line 22: here's the "real" breaking change 😉 so we should rebase these changes into `zapier-platform-major-release-16.0.0-dev` instead of `main`

rnegron (commented):
Very exciting that all tests are passing (and there's new ones!) I ran a few commands locally as well and all worked as expected. Left a few final questions/comments

packages/cli/src/oclif/buildFlags.js
szchenghuang: Good question! It digs into the oclif package but it ends up [here](https://github.com/oclif/core/blob/main/src/interfaces/parser.ts#L63). In essence, type `T` there for `Flags.string` is `string` which is the default type for `default`.

packages/cli/src/oclif/commands/versions.js
szchenghuang: You're right. I forgot to remove it that I used for validation during development.

packages/cli/package.json
szchenghuang: Done!

szchenghuang:
> Very exciting that all tests are passing (and there's new ones!) I ran a few commands locally as well and all worked as expected. Left a few final questions/comments

@rnegron Thanks for taking a look. I addressed your feedback. Let me know if there are any changes you'd like to add.

rnegron (approved):
Thanks @szchenghuang! I'm sure this was a tricky migration (v1->v4!); thanks for the great work!

PR #894, opened 2024-10-18
feat(cli): Add OAuth to TS init output [IQQ-1860]

tkcranny:
This upgrades the `zapier init` TypeScript example (and the example reference application as well) to use a much closer to real-world OAuth2 example configuration with scopes and `API_URL` constants.

A minor tweak was made to the Before and After Middelware types to make the arguments provided no longer optional, as they are always present.

### DIFF
--- a/example-apps/typescript/src/authentication.ts
+++ b/example-apps/typescript/src/authentication.ts
@@ -1,0 +1,32 @@
+import type { Authentication } from 'zapier-platform-core';
+
+import { SCOPES } from './constants';
+
+export default {
+  type: 'oauth2',
+  test: { url: 'https://api.example.com/v2/token/authorized_by' },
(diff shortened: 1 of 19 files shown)


example-apps/typescript/src/authentication.ts
```diff
@@ -0,0 +1,42 @@
+import type { Authentication } from 'zapier-platform-core';
+
+import { SCOPES } from './constants';
+
+export default {
+  type: 'oauth2',
+  test: { url: 'https://auth-json-server.zapier-staging.com/me' },
```
kola-er, line 7: Suggest using the `API_URL` constant here and in the oauth2Config below.

tkcranny: Done 👍

PR #895, opened 2024-10-18
feat(types): Type z.request responses; deprecate `.json`

tkcranny:
This adds generic parameters to the `HttpResponse` and `RawHttpResponse` types that allow them to specify what they're returning. If not provided the default is `any`, similar to the previous behaviour.

```ts
interface User {
  id: numner,
  name: string,
}

const resp = await z.request<User>('https://example.com/users');
resp.data.firstName // Error: Type `firstName` doesn't exist on type User, did you mean `name`.
```

---

Because of [this note](https://github.com/zapier/zapier-platform/blob/main/packages/cli/README.md#bundletargeturl) in the CLI docs, I've also marked the `.json` attribute on `HttpResponse` objects as `@deprecated`. It will now show up in the editor struck out, and suggest using `.data` instead. Is this deprecating `.json` correct in your opinion?

Screen Shot 2024-10-18 at 17 32 53

### DIFF
--- a/packages/core/types/zapier.custom.d.ts
+++ b/packages/core/types/zapier.custom.d.ts
@@ -117,16 +117,16 @@ interface BaseHttpResponse {
   request: HttpRequestOptions;
 }
 
-export interface HttpResponse extends BaseHttpResponse {
+export interface HttpResponse<T = any> extends BaseHttpResponse {
   content: string;
-  data?: any;
-  json?: any;
+  data: T;
+  json?: T;
 }
 
-export interface RawHttpResponse extends BaseHttpResponse {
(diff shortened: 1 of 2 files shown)


kola-er (approved):

packages/core/types/zapier.custom.d.ts
```diff
@@ -138,16 +139,20 @@ type DehydrateFunc = <T>(
 export interface ZObject {
   request: {
     // most specific overloads go first
-    (
```
FokkeZB, line 141: @tkcranny I wonder if the default better be `unknown`, so that we don't effectively disable type checking for all downstream code.

PR #901, opened 2024-10-23
fix(core): [PDE-5470] Force exit process on EMFILE error inside checkMemory

rnegron:
See related Slack thread: https://zapier.slack.com/archives/C04P5J5R8E4/p1729720807731419?thread_ts=1729271444.276339&cid=C04P5J5R8E4

You can recreate the `EMFILE` error locally like:

```js
const fs = require('fs');
const path = require('path');

// Path to a file you want to repeatedly open (can be any existing file)
const filePath = path.resolve(__dirname, 'testfile.txt');

// Ensure the test file exists
fs.writeFileSync(filePath, 'This is a test file.');

const openFiles = [];

function openFileRepeatedly() {
  try {
    // Attempt to open a file repeatedly without closing it
    for (let i = 0; i < 10000000; i++) {
      const fd = fs.openSync(filePath, 'r');
      openFiles.push(fd);
      console.log(`Opened file descriptor ${fd}`);
    }
  } catch (err) {
    console.log(`Error code: ${err.code}`)
    console.error(`Error occurred: ${err.message}`);
  } finally {
    // Clean up and close all open file descriptors
    openFiles.forEach(fd => fs.closeSync(fd));
    console.log('Cleaned up opened file descriptors.');
  }
}

openFileRepeatedly();
```

### DIFF
--- a/packages/core/src/tools/create-lambda-handler.js
+++ b/packages/core/src/tools/create-lambda-handler.js
@@ -198,20 +198,7 @@ const createLambdaHandler = (appRawOrPath) => {
 
     // If we're running out of memory or file descriptors, force exit the process.
     // The backend will try again via @retry(ProcessExitedException).
-    try {
-      checkMemory(event);
-    } catch (err) {
-      if (err.code === 'EMFILE') {
-        console.error(
-          'Force killing process by Zapier for too many open file descriptors'
-        );
-
-        /* eslint no-process-exit: 0 */
-        process.exit(1);
-      } else {
-        throw err;
-      }
-    }
+    checkMemory(event);
 
     environmentTools.cleanEnvironment();
 
--- a/packages/core/src/tools/create-lambda-handler.js
+++ b/packages/core/src/tools/create-lambda-handler.js
@@ -196,13 +196,18 @@ const createLambdaHandler = (appRawOrPath) => {
       ZapierPromise.patchGlobal();
     }
 
-    // If we're running out of memory, exit the process. Backend will try again.
+    // If we're running out of memory or file descriptors, force exit the process.
+    // The backend will try again via @retry(ProcessExitedException).
     try {
       checkMemory(event);
     } catch (err) {
-      // Don't crash the invocation if we can't check memory usage.
       if (err.code === 'EMFILE') {
-        console.error('Too many open files, skipping memory usage check...');
+        console.error(
+          'Force killing process by Zapier for too many open file descriptors'
+        );
+
+        /* eslint no-process-exit: 0 */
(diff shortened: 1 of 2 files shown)


standielpls (dismissed):

rnegron:
Before merging, I'm considering if we should instead `process.exit(1)` here and force a container restart (thus clearing open file handles...) which is what the `checkMemory` function is actually trying to do but is not able to due to the early exception being raised.

This old AWS doc mentions that "crashing" is effectively a way to restart the Lambda container: https://aws.amazon.com/blogs/compute/container-reuse-in-lambda/ 
Screen Shot 2024-10-23 at 5 56 14 PM

eliangcs (commented):
The [`checkMemory()`](https://github.com/zapier/zapier-platform/blob/c02bdd928f8c3176007ee1c8b59f61254e672b01/packages/core/src/tools/memory-checker.js#L8) function doesn't open any files. So how is it possible for it to throw an error with code `EMFILE`? Or have you found a way to reproduce `EMFILE` using `checkMemory()`?

rnegron:
@eliangcs 

> I tested your script. The `EMFILE` error is thrown by the `openSync()` function. `checkMemory()` doesn't open any files, so it doesn't seem `checkMemory()` will ever throw an `EMFILE` error.

The example script I provided is just to show how we might capture the specific error (since `err.code === 'EMFILE'`). In that case, yes, the error is thrown via `openSync()`.

But surprisingly, we can observe `checkMemory()` opening file handles via `process.memoryUsage()` ([log gist](https://logs.internal.zapier.com/messages/integration_7075/daff0b11-9155-11ef-8667-aae4637293f7)), likely to read from `/proc/self/stat` or similar. See this Slack comment for more details: https://zapier.slack.com/archives/C04P5J5R8E4/p1729709367302319?thread_ts=1729271444.276339&cid=C04P5J5R8E4

Since `checkMemory()` happens outside the main Lambda Handler block, let's see if catching the error here (earlier) solves the issue!

rnegron:
@eliangcs Can do! I thought about that approach initially but noticed that `process.memoryUsage();` is called conditionally two times, so refactoring it outside _could_ add additional unnecessary calls in certain scenarios. But maybe it's offset by only having one call. Let me try!

rnegron:
OK, I think the review suggestion is good as-is! The conditionals are really about avoiding `console.logs` and not about avoiding `process.memoryUsage` calls. Tests pass so it's not a CI concern either 👍🏻

eliangcs (approved):
Thanks for the change!

PR #904, opened 2024-10-25
feat(core) - PDE-5490 calling lambda callback doesnt wait for event loop

standielpls:
In debugging some apps, it seems like sometimes lambdas may be hanging indefinitely until the timeout expires.
This causes the action to fail, even if all the app code has completed. 
Debugging further it looks like it's possible that when processing larger datasets, some open connections will be kept open and block the event loop from clearing. 

This heuristic is something we can add from Lambda
https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html
the gist is, if we set `context.callbackWaitsForEmptyEventLoop = false;` then once the callback is called, we exit, despite having other connections

https://zapierorg.atlassian.net/browse/PDE-5490

### DIFF
--- a/packages/core/src/tools/create-lambda-handler.js
+++ b/packages/core/src/tools/create-lambda-handler.js
@@ -191,8 +191,7 @@ const createLambdaHandler = (appRawOrPath) => {
     // using the callback; just putting it here to be explicit.
     // In some cases, the code hangs and never exits because the event loop is not
     // empty, so we can override the default behavior and exit after the app is done.
-    context.callbackWaitsForEmptyEventLoop =
-      event.waitForAsyncBeforeExit ?? true;
+    context.callbackWaitsForEmptyEventLoop = !event.skipWaitForAsync || true;
 
     // replace native Promise with bluebird (works better with domains)
     if (!event.calledFromCli) {
--- a/packages/core/src/tools/create-lambda-handler.js
+++ b/packages/core/src/tools/create-lambda-handler.js
@@ -192,7 +192,7 @@ const createLambdaHandler = (appRawOrPath) => {
     // In some cases, the code hangs and never exits because the event loop is not
     // empty.
     context.callbackWaitsForEmptyEventLoop =
-      event.callbackExitsAfterApp || true;
+      event.waitForAsyncBeforeExit ?? true;
 
     // replace native Promise with bluebird (works better with domains)
     if (!event.calledFromCli) {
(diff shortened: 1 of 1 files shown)


packages/core/src/tools/create-lambda-handler.js
```diff
@@ -264,6 +264,13 @@ const createLambdaHandler = (appRawOrPath) => {
           const compiledApp = schemaTools.prepareApp(appRaw);
 
           const input = createInput(compiledApp, event, logger, logBuffer, rpc);
+
+          // When the app is done, callback() will be called and the Lambda function will end.
+          // In some cases, perhaps due to fetching large data blobs, the Lambda function
+          // will hang waiting for the event loop to drain. Setting this to false will
+          // prevent that and return when the callback is called.
+          context.callbackWaitsForEmptyEventLoop = false;
```
rnegron, line 272: 🤔 I'm a bit nervous about enabling this by default. It seems like a good default for Lambda to wait until the event loop is empty before returning. Otherwise, I imagine we can introduce unpredictability that's quite hard to debug since some async tasks will complete sometimes and not other times.

We're actually explicitly setting this to `true` a bit above, [here](https://github.com/zapier/zapier-platform/blob/e8b49e027e98cef2b345e42d20707e520ab49f2f/packages/core/src/tools/create-lambda-handler.js#L192).

What about enabling this per-app to start with? At this point in the code we have access to the `event` payload from Lambda in which we could provide app-specific flags.

packages/core/src/tools/create-lambda-handler.js
```diff
@@ -189,7 +189,9 @@ const createLambdaHandler = (appRawOrPath) => {
     // Wait for all async events to complete before callback returns.
     // This is not strictly necessary since this is the default now when
     // using the callback; just putting it here to be explicit.
-    context.callbackWaitsForEmptyEventLoop = true;
+    // In some cases, the code hangs and never exits because the event loop is not
+    // empty, so we can override the default behavior and exit after the app is done.
+    context.callbackWaitsForEmptyEventLoop = !event.skipWaitForAsync || true;
```
rnegron, line 194: @standielpls I think this is always `true`, no? `false || true === true`, `undefined || true === true`.

```suggestion
    context.callbackWaitsForEmptyEventLoop =  true;
    if (event.skipWaitForAsync === false) {
        context.callbackWaitsForEmptyEventLoop = false;
     }
```

rnegron: Alternatively: `event.skipWaitForAsync !== true;`

standielpls: Whoops, yes you're right!

rnegron (approved):

PR #905, opened 2024-10-28
fix(core): HTTP patch breaks aws-sdk v3 (or other libraries) intermittently

eliangcs:
Fixes an issue where a late listener of a third-party library (like aws-sdk v3) attached to HTTP response stream could miss data chunks that were already consumed by our HTTP logger. The fix involves:

1. Replacing direct `response.on('data')` listener with a patched `emit` function so that we don't attach a listener earlier than any other listeners

2. Maintaining the original event emission behavior by calling the original `emit` with all arguments

3. Adding a test that verifies:
   - Late listeners receive complete data without missing bytes
   - Logger captures the complete response as expected

The changes also include:
- Adding `text/plain; charset=utf-8` to allowed content types

See also this [video](https://zapier.slack.com/archives/C5Z9BP4U9/p1730112195116519) (internal-only) to walk through my thought process.

### DIFF
--- a/packages/core/test/tools/create-http-patch.js
+++ b/packages/core/test/tools/create-http-patch.js
@@ -1,13 +1,20 @@
 'use strict';
 
 const EventEmitter = require('events');
+const https = require('node:https');
 
 const should = require('should');
 
-const { JSON_TYPE, XML_TYPE, BINARY_TYPE } = require('../../src/tools/http');
+const {
+  BINARY_TYPE,
+  JSON_TYPE,
+  TEXT_TYPE,
+  XML_TYPE,
+} = require('../../src/tools/http');
 const createAppTester = require('../../src/tools/create-app-tester');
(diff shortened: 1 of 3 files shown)


packages/core/test/tools/create-http-patch.js
```diff
@@ -152,6 +159,89 @@ describe('create http patch', () => {
...
+      (res) => {
+        // Make sure it's a response object
+        should(res.statusCode).eql(200);
+        should(res.headers['content-type']).containEql(TEXT_TYPE);
+
+        setTimeout(() => {
+          // Suppose this listener is some HTTP client library that consumes the
+          // response stream. We use setTimeout to recreate the scenario where
+          // the listener is attached AFTER the logger has already started
+          // consuming the stream.
```
rnegron, line 188: Nice test 💯

rnegron (approved):

PR #908, opened 2024-10-29
PDE-5489 fix(core): patch native fetch to write http logs

eliangcs:
Adds logging capabilities for outbound HTTP requests made via the native [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), similar to our existing logging for Node's built-in `http`/`https` modules.

We're doing this because some SDKs, such as [webflow-api](https://github.com/webflow/js-webflow-api), [use the native fetch](https://github.com/webflow/js-webflow-api/blob/472bac9a5d43b9a15c96b9d7bb96ebe36341b738/src/core/fetcher/getFetchFn.ts#L9) and their requests aren't logged.

### Changes

1. Added new `fetch-logger.js` module that wraps the global `fetch` function to log request/response details
2. Modified `create-lambda-handler.js` to patch `global.fetch` when HTTP logging is enabled
3. Added comprehensive tests for the fetch logger functionality

### Implementation details

The fetch logger wrapper:
- Logs requests and responses in the same format as our existing HTTP logger
- Handles various input types (URL strings, `URL` objects, `Request` objects) that `fetch` accepts
- Properly handles request body data in different formats (string, `URLSearchParams`, `FormData`)
- Skips logging for requests made by `z.request()` (by checking `User-Agent` header)
- Clones response before reading body to avoid consuming the original stream

### What will it look like?

```
webflow (main) ❯ zapier-dev invoke trigger sites --debug
  zapier:invoke argv is [ 'trigger', 'sites', '--debug' ] +0ms
  zapier:invoke args are { actionType: 'trigger', actionKey: 'sites' } +1ms
  zapier:invoke flags are {
  debug: true,
  isFillingDynamicDropdown: false,
  isLoadingSample: false,
  isPopulatingDedupe: false,
  limit: -1,
  page: 0,
  'non-interactive': false,
  timezone: 'America/Chicago',
  'redirect-uri': 'http://localhost:9000',
  'local-port': 9000
} +1ms
  zapier:invoke ------------ +0ms

Your code depends on v15.17.0 of zapier-platform-core, but your local copy is v15.18.0. You should probably reinstall your dependencies.

  zapier:invoke Action type: trigger +0ms
  zapier:invoke Action key: sites +0ms
  zapier:invoke Action label: Sites +0ms
  zapier:invoke staticInputFields: [] +0ms
⠋ Invoking triggers.sites.operation.inputFields  zapier:analytics skipping analytics +0ms
✔ Invoking triggers.sites.operation.inputFields
  zapier:invoke inputFields: [] +5ms
⠹ Invoking triggers.sites.operation.perform  zapier:invoke 200 GET https://api.webflow.com/v2/sites +1s
  zapier:invoke {
  zapier:invoke   log_type: 'http',
  zapier:invoke   request_type: 'patched-devplatform-outbound',
  zapier:invoke   request_url: 'https://api.webflow.com/v2/sites',
  zapier:invoke   request_method: 'GET',
  zapier:invoke   request_headers: 'Authorization: :censored:71:15968376db:\n' +
  zapier:invoke     'X-Fern-Language: JavaScript\n' +
  zapier:invoke     'X-Fern-SDK-Name: webflow-api\n' +
  zapier:invoke     'X-Fern-SDK-Version: 2.4.2\n' +
  zapier:invoke     'User-Agent: webflow-api/2.4.2\n' +
  zapier:invoke     'X-Fern-Runtime: node\n' +
  zapier:invoke     'X-Fern-Runtime-Version: 20.13.1',
  zapier:invoke   request_data: '',
  zapier:invoke   request_via_client: false,
  zapier:invoke   response_status_code: 200,
  zapier:invoke   response_headers: 'connection: keep-alive\n' +
  zapier:invoke     'content-length: 361\n' +
  zapier:invoke     'content-type: application/json; charset=utf-8\n' +
  zapier:invoke     'date: Wed, 30 Oct 2024 10:15:20 GMT\n' +
  zapier:invoke     'etag: W/"169-U5r9O4ikVvPAFGKPnnLy0ALiZaw"\n' +
  zapier:invoke     'retry-after: 60\n' +
  zapier:invoke     'x-ratelimit-limit: 60\n' +
  zapier:invoke     'x-ratelimit-remaining: 59\n' +
  zapier:invoke     'x-response-time: 274.785ms',
  zapier:invoke   response_content: `{"sites":[{"id":"671b4b7e3440e8bab7c132c4","workspaceId":"671b4b5ee4762fb852365cc1","displayName":"Chang-Hung's Test Site","shortName":"chang-hungs-test-site","previewUrl":null,"timeZone":"Asia/Taipei","createdOn":"2024-10-25T07:40:46.779Z","lastUpdated":"2024-10
-25T07:40:55.083Z","lastPublished":null,"parentFolderId":null,"customDomains":[],"locales":null}]}`
  zapier:invoke } +0ms
✔ Invoking triggers.sites.operation.perform
[
  {
    "id": "671b4b7e3440e8bab7c132c4",
    "workspaceId": "671b4b5ee4762fb852365cc1",
    "displayName": "Chang-Hung's Test Site",
    "shortName": "chang-hungs-test-site",
    "timeZone": "Asia/Taipei",
    "createdOn": "2024-10-25T07:40:46.779Z",
    "lastUpdated": "2024-10-25T07:40:55.083Z",
    "customDomains": []
  }
]
```


packages/core/src/tools/fetch-logger.js
```diff
@@ -0,0 +1,106 @@
...
+    if (requestInfo && !isZapierUserAgent(requestInfo.headers)) {
+      const responseContentType = response.headers.get('content-type');
+
+      logger(`${response.status} ${requestInfo.method} ${requestInfo.url}`, {
+        log_type: 'http',
+        request_type: 'patched-devplatform-outbound',
+        request_url: requestInfo.url,
+        request_method: requestInfo.method,
+        request_headers: requestInfo.headers,
+        request_data: requestInfo.data,
```
rnegron, line 90: @eliangcs Should we consider avoiding `streaming data` here? I recall streaming data for the patched `http` module being a likely cause of runaway ingress data increase, see [this Slack message](https://app.slack.com/client/T024VA8T9/C5Z9BP4U9) for context.

rnegron: It seems like `stringifyRequestData` handles this; just making sure to call this out explicitly just in case it was not considered 😄

packages/core/test/tools/fetch-logger.js
```diff
@@ -0,0 +1,207 @@
+/* global Request */
+
+const assert = require('node:assert');
```
rnegron, line 3: 👁️ Very nice, I've been interested in this built-in library! Should we go ahead and default to the `strict` version? `require('node:assert/strict')`

rnegron (dismissed):

packages/core/src/tools/fetch-logger.js
eliangcs: That's a good call. I [added](https://github.com/zapier/zapier-platform/pull/908/commits/82b50d5e0963ba852876cd94fff6e8d2a830a6e8) a comment in `stringifyRequestData` to call it out for posterity.

packages/core/test/tools/fetch-logger.js
eliangcs: Switched to `node:assert/strict` in [cb6bdd9](https://github.com/zapier/zapier-platform/pull/908/commits/cb6bdd9adc1d05f441254a35aec771e8cd14cf26). I actually caught  an issue with the non-strict `assert.equal()` where it sees `'foo'` and `['foo']` as equal. Using `node:assert/strict` and `assert.deepEqual()` fixed that, so thanks!

szchenghuang (approved):

PR #917, opened 2024-11-19
feat(core) Type bundle.meta.timezone

FokkeZB:
Types `bundle.meta.timezone`, which once https://gitlab.com/zapier/zapier/-/merge_requests/60106 is merged will be set regardless of `zapier-platform-core` version.

### DIFF
--- a/packages/core/types/zapier.custom.d.ts
+++ b/packages/core/types/zapier.custom.d.ts
@@ -43,6 +43,7 @@ export interface Bundle<InputData = { [x: string]: any }> {
     isTestingAuth: boolean;
     limit: number;
     page: number;
+    timezone: string | null;
     zap?: { id: string };
   };
   rawRequest?: Partial<{


kola-er:
Does it to be documented for public usage [here](https://docs.zapier.com/platform/build-cli/overview#bundle-meta)?

packages/core/types/zapier.custom.d.ts
```diff
@@ -43,6 +43,7 @@ export interface Bundle<InputData = { [x: string]: any }> {
     isTestingAuth: boolean;
     limit: number;
     page: number;
+    timezone: string | undefined;
```
kola-er, line 46:
```suggestion
    timezone: string | null;
```

Since the backend returns `None`.

FokkeZB: Thanks for catching that! Indeed `None` converts to `null` via JSON.

FokkeZB:
> Does it to be documented for public usage [here](https://docs.zapier.com/platform/build-cli/overview#bundle-meta)?

Yes, I'll update those docs as well.

kola-er (approved):

FokkeZB:
@kola-er this is good to merge!

FokkeZB:
Replaced by https://github.com/zapier/zapier-platform/pull/922

PR #921, opened 2024-11-20
style(core) Update the TS typing for z.cache.set to include args nx

kola-er:


packages/core/types/zapier.custom.d.ts
```diff
@@ -213,7 +213,7 @@ export interface ZObject {
 
   cache: {
     get: (key: string) => Promise<any>;
-    set: (key: string, value: any, ttl?: number) => Promise<boolean>;
+    set: (key: string, value: any, ttl?: number, nx?: boolean) => Promise<boolean|null>;
```
kola-er, line 216: The return value of `null` is for throttled requests, while the boolean `true/false` is for when the key/value pair is set or not set at the cache level.

packages/core/types/zapier.custom.d.ts
```diff
@@ -213,7 +213,7 @@ export interface ZObject {
 
   cache: {
     get: (key: string) => Promise<any>;
-    set: (key: string, value: any, ttl?: number) => Promise<boolean>;
+    set: (key: string, value: any, ttl?: number, scope?: string[], nx?: boolean) => Promise<boolean|null>;
```
kola-er, line 216: The return value of null is for throttled requests, while the boolean true/false is for when the key/value pair is set or not set at the cache level.

PR #922, opened 2024-11-21
chore(ts): Update bundle.meta typings [IQQ-2105]

tkcranny:
This updates the `bundle.meta` types with clearer docs, deprecation markings, and support for the new `timezone` and `withSearch` fields.

I've followed the [Core Docs](https://docs.zapier.com/platform/build-cli/core#bundle-meta) on this subject, but if anything is not strictly correct with the platform's behaviour I'll be happy to adjust.

- Deprecated `meta.zap`, and all of the inner `id`, `user`, and `user.timezone` properties
- Deprecated `isTestingAuth`, because the docs say "Legacy property".
- Added `withSearch: string | undefined`. Docs unclear if this is always set and will be `string | undefined`, or _if_ the key is present it will be a valid string, meaning `withSearch?: string` would be better.
- Added `timezone: string | null`. Same as above.
- `isBulkRead`: Unchanged, but curious if it should be marked deprecated or internal only too.

### DIFF
--- a/packages/core/types/zapier.custom.d.ts
+++ b/packages/core/types/zapier.custom.d.ts
@@ -37,13 +37,71 @@ export interface Bundle<InputData = { [x: string]: any }> {
   inputDataRaw: { [x: string]: string };
   meta: {
     isBulkRead: boolean;
+
+    /**
+     * If true, this poll is being used to populate a dynamic dropdown.
+     * You only need to return the fields you specified (such as id and
+     * name), though returning everything is fine too.
+     */
     isFillingDynamicDropdown: boolean;
+
+    /**
+     * If true, this run was initiated manually via the Zap Editor.
+     */
     isLoadingSample: boolean;
+
+    /**
+     * If true, the results of this poll will be used to initialize the
+     * deduplication list rather than trigger a zap. You should grab as
+     * many items as possible.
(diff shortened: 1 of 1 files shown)


packages/core/types/zapier.custom.d.ts
```diff
@@ -37,13 +37,71 @@ export interface Bundle<InputData = { [x: string]: any }> {
...
+     * be returned.
+     */
     page: number;
-    zap?: { id: string };
+
+    /**
+     * When a create is called as part of a search-or-create step,
+     * this will be the key of the search.
+     */
+    withSearch: string | undefined;
```
kola-er, line 86:
```suggestion
    withSearch?: string;
```

The backend code shows that it is not always present.

kreddlear (approved):
@tkcranny looks good! If you're curious, looks like for `timezone`, it always exists but can be `None`, [based on this code](https://gitlab.com/zapier/zapier/-/blob/main/web/backend/zapier/app_platform/developer_cli/api_base.py#L808), so your existing code is correct 👍 ([as compared with `withSearch`](https://gitlab.com/zapier/zapier/-/blob/main/web/backend/zapier/app_platform/developer_cli/api_base.py#L2674))

PR #925, opened 2024-11-25
Release v15.19.0

kola-er:


CHANGELOG.md
```diff
@@ -1,3 +1,37 @@
...
+- :tada: Add `meta` field support for the `FieldSchema` ([#883](https://github.com/zapier/zapier-platform/pull/883))
+
+### schema-to-ts
+
+None!
+
+### misc
+
+- :hammer: Bump elliptic from 6.5.7 to 6.6.0 ([#911](https://github.com/zapier/zapier-platform/pull/911))
+- :tada: Add schema-to-ts as part of yarn workspace ([#913](https://github.com/zapier/zapier-platform/pull/913))
```
rnegron, line 32: @kola-er I believe this MR (#913) is not part of 15.19.0; it was merged into the `zapier-platform-major-release-16.0.0-dev` branch (not `main`)

CHANGELOG.md
```diff
@@ -1,3 +1,37 @@
+## 15.19.0
+
+_released `2024-11-25`_
+
+For optimal performance, an optional argument has been added to [z.dehydrate](https://docs.zapier.com/platform/build-cli/overview#dehydration) to allow devs to control the caching duration on hydrators.
```
rnegron, line 5: would it be worth it to add a bit more context on the default?

```suggestion
For optimal performance, an optional argument has been added to [z.dehydrate](https://docs.zapier.com/platform/build-cli/overview#dehydration) to allow devs to control the caching duration on hydrators. The default cache duration for hydrators is unchanged (5 minutes).
```

CHANGELOG.md
```diff
@@ -1,3 +1,37 @@
...
+
+Also, with the new `meta` field object added to the [FieldSchema](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#fieldschema) object, devs have more flexibility storing field-specific context which may be obtained when constructing dynamic needs, and then making that context available during actual action execution in the perform method.
```
rnegron, line 7: Adding "optional" working here as well? also changing the working to make it clear this is the expected behavior now, if needed

```suggestion
Also, with the new optional `meta` field object added to the [FieldSchema](https://github.com/zapier/zapier-platform/blob/main/packages/schema/docs/build/schema.md#fieldschema) object, devs now have a supported mechanism for storing field-specific context which may be obtained when constructing dynamic needs, and then making that context available during actual action execution in the perform method.
```

CHANGELOG.md
kola-er: Thanks, removed it.

CHANGELOG.md
kola-er: Updated

PR #926, opened 2024-11-25
BREAKING CHANGE(all): [v16] Bump dependency package versions

rnegron:
# root (monorepo `package.json`) package updates
## Major version changes
* `lerna`: 6.4.1 -> 8.0.2 (drops Node.js 16 support, [changelog](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes))
* `husky`: 7.0.4 -> 9.1.7 (no real breaking changes, just configuration and caller changes, [changelog](https://github.com/typicode/husky/releases/tag/v9.0.1)) 
* `mocha`: 10.8.2 -> 11.0.1 (no real breaking changes, [release plan](https://github.com/mochajs/mocha/issues/5249))

# zapier-platform-cli package updates
## Major version changes
* `archiver` 5.3.1 -> 7.0.1: (requires Node.js 12, [changelog](https://github.com/archiverjs/node-archiver/releases/tag/7.0.0))
* `gulp-prettier`: 4.0.0 -> 5.0.0 (requires Node.js 16, [changelog](https://github.com/thomasvantuycom/gulp-prettier/releases/tag/v5.0.0))
* `marked` 4.2.12 -> 14.1.4 (requires Node.js 18, [changelog](https://github.com/markedjs/marked/releases/tag/v5.0.0))
    * Don't think we can upgrade to v15 until `marked-terminal` supports it officially?
* `marked-terminal`: 5.2.0 -> 7.2.1 (required Node.js 16, [changelog](https://github.com/mikaelbr/marked-terminal/releases/tag/v6.0.0))
* `read`: 2.1.0 -> 4.0.0 (requires Node.js 18, [changelog](https://github.com/npm/read/releases/tag/v4.0.0))

## Minor version changes
* `fs-extra`: 11.1.1 -> 11.2.0
* `yeoman-environment`: 3.3.0 -> 3.19.3
* `semver`: 7.5.2 -> 7.6.3
* `node-fetch`: 2.6.7 -> 2.7.0

## Patch version changes
* `@oclif/*` dependencies: patch version bumps after the major version bumps performed in: https://github.com/zapier/zapier-platform/pull/890
* `adm-zip`: 0.5.10 -> 0.5.16 
* `browserify`: 17.0.0 -> 17.0.1
* `cli-table3`: 0.6.3 -> 0.6.5
* `tmp`: 0.2.1 -> 0.2.3
* `traverse`: 0.6.7 -> 0.6.10

# zapier-platform-core package updates
## Major version changes
[...]

PR #927, opened 2024-12-03
chore(all): [v16] Bump eslint and related packages, bump prettier and run on codebase

rnegron:
**Note:** This MR is based on top of https://github.com/zapier/zapier-platform/pull/926. It should be re-based and merged into the v16 branch after that one is merged.

# Packages Updates

* `eslint` 8.32.0 -> 9.16.0
* `eslint-config-prettier`  8.6.0 -> 9.16.0
* `eslint-plugin-import`: 2.27.5 -> 2.31.0
* `eslint-plugin-mocha`: 9.0.0 -> 10.5.0
* `eslint-plugin-n`:  15.6.1 -> 17.14.0
*  `eslint-plugin-promise`: 6.1.1 -> 7.2.1
* `prettier`: 2.8.3 -> 3.4.1

# Package Removals

* `eslint-plugin-standard`: see https://github.com/standard/standard/issues/1316

# Other Changes
* Pin `engine` to `>=18.18` for the `cli` package, to mirror the upgraded `eslint` dependency...

* The vast majority of the file changes in this PR are due to the `trailingCommas` [option in prettier](https://prettier.io/docs/en/options.html#trailing-commas), which had its default value changed from "es5" to "all" in v3.0.0
  * Even when using "es5", running `prettier` on the codebase generates a lot of file changes. So I figured we should just keep the default value then.

* `.eslintrc` is replaced by `eslint.config.cjs` (auto-generated via `npx @eslint/migrate-config .eslintrc --commonjs` according to the [migration guide](https://eslint.org/docs/latest/use/configure/migration-guide), which then required we install `globals`, `@eslint/js` and `@eslint/eslintrc`)
* `.eslintignore`: "ESLintIgnoreWarning: The ".eslintignore" file is no longer supported. Switch to using the "ignores" property in "eslint.config.js": https://eslint.org/docs/latest/use/configure/migration-guide#ignoring-files"


scripts/bump.js
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-
+/* eslint-disable array-callback-return */
```
rnegron, line 2: Calling out this specific disable I had to add to avoid changing some `.map()` calls below

packages/cli/src/utils/convert.js
```diff
@@ -40,8 +40,9 @@ const createFile = async (content, filename, dir) => {
   endSpinner();
 };
 
-const prettifyJs = (code) =>
+const prettifyJs = async (code) =>
```
rnegron, line 43: prettier.format returns a Promise since v3. This means `renderDefinitionSlice` below which uses `prettifyJs` needs to await it. Interestingly, the callers for `renderDefinitionSlice` itself already awaited the response 👍🏻

The other usage of `prettifyJs` is in `renderIndex` which already returns a Promise.

package.json
```diff
@@ -46,7 +48,7 @@
       "eslint --fix --quiet"
     ],
     "*.{js,json}": [
-      "prettier --write"
+      "prettier --write --ignore-path=.prettierignore"
```
rnegron, line 51: Not sure if this has much effect; keeping previous behavior to avoid surprises just in case.

packages/core/test/moduleuserapp/export-func.js
```diff
@@ -1,3 +1,3 @@
 module.exports = (z, bundle) => {
-  return [{ id: 1234 }] || bundle;
```
rnegron, line 2: Calling out this change I did to fix `no-constant-binary-expression` eslint error

eslint.config.cjs
```diff
@@ -0,0 +1,64 @@
...
+} = require("@eslint/eslintrc");
+
+const compat = new FlatCompat({
+    baseDirectory: __dirname,
+    recommendedConfig: js.configs.recommended,
+    allConfig: js.configs.all
+});
+
+module.exports = [{
+    ignores: [
```
rnegron, line 16: Moved `.eslintignore` into this `ignores` property as required by eslint 9

eslint.config.cjs
```diff
@@ -0,0 +1,64 @@
...
+        // this needs to combine all other eslintignore files so that lint-staged knows what to ignore
+        "packages/cli/scaffold/*.js",
+        "packages/cli/src/generators/templates/**/*.template.js",
+        "**/.yarn",
+        "example-apps/onedrive/test/*",
```
rnegron, line 21: Note: I added this new ignore rule for onedrive example app tests; there seems to be some issue with eslint trying to read the `.eslintrc.json` in that root directory (?!) which I didn't want to get into...

packages/cli/package.json
```diff
@@ -19,7 +19,7 @@
     "/oclif.manifest.json"
   ],
   "engines": {
-    "node": ">=18"
+    "node": ">=18.18"
```
rnegron, line 22: Note this stricter requirement; eslint 9 requires the [18.18 minor version](https://nodejs.org/en/blog/release/v18.18.0), so I figure it makes sense to enforce this for at least `zapier-platform-cli`

eslint.config.cjs
kreddlear: Yikes, yeah that seems reasonable!

kreddlear (approved):
Whew, that's a lot of commas 😅 Thanks for adding comments on the main changes for this, that made it much easier to review! Looks great 👍

PR #934, opened 2024-12-17
PDE-5506 fix(core): add output check dynamicFieldsHaveKeys

eliangcs:


eliangcs:
There's a failing smoke test, which has been fixed by https://github.com/zapier/zapier-platform/pull/935.

rnegron (approved):

PR #935, opened 2024-12-17
PDE-5506: fix(cli): move swapi.dev to swapi.info for dynamic-dropdown template

eliangcs:
https://swapi.dev is down. It has been down at least twice as far as I can remember. I found https://swapi.info seems to be a more reliable site. Let's move to that.


rnegron (approved):
Thanks for catching and fixing this one!

PR #942, opened 2025-01-03
[PDE-5660] feat(cli): Require `--force` flag when updating production integration version environment variables via `env:set` or `env:unset`

rnegron:
The idea here is to provide a bit of a warning for a potentially dangerous operation for live app versions. In fact, UI integrations can't update env vars for production versions at all, so it makes sense to add a buffer here for CLI integrations.

Since `is_production` is a backend check, let's react to a specific HTTP status and let the user know how to override via `--force`.  I'm not using `this.confirm` (similar to [user:add](https://github.com/zapier/zapier-platform/blob/4f55292cdb2b5a7d890fbb40c877c1a7c3922587/packages/cli/src/oclif/commands/users/add.js#L11)) since  having users re-write the command with the added flag is even stricter.

See related backend MR: https://gitlab.com/zapier/zapier/-/merge_requests/60723

### DIFF
--- a/packages/cli/src/oclif/commands/env/set.js
+++ b/packages/cli/src/oclif/commands/env/set.js
@@ -1,4 +1,4 @@
-const { Args } = require('@oclif/core');
+const { Args, Flags } = require('@oclif/core');
 const { cyan } = require('colors/safe');
 const { omit } = require('lodash');
 
@@ -45,21 +45,30 @@ class SetEnvCommand extends BaseCommand {
     }
 
     const url = `/apps/${app.id}/versions/${version}/multi-environment`;
+    const requestOptions = {
+      body: payload,
+      method: 'POST',
+    };
+
+    if (this.flags.force) {
+      requestOptions.extraHeaders = {
+        'X-Force-Env-Var-Update': true,
+      };
+    }
 
     try {
-      // currently, this returns nothing
-      await callAPI(
-        url,
-        {
-          body: payload,
-          method: 'POST',
-        },
-        true,
-      );
+      await callAPI(url, requestOptions, true);
 
       this.log(successMessage(version));
       this.logJSON(payload);
     } catch (e) {
+      if (e.statusCode === 409) {
+        this.error(
+          `App version ${version} is the production version. Are you sure you want to set potentially live environment variables?` +
+            ` If so, run this command again with the --force flag.`,
+        );
+      }
+
       // comes back as json: { errors: [ 'The following keys failed to update: 3QER, 4WER' ] },
       const failedKeys = e.json.errors[0].split('update: ')[1].split(', ');
       const successfulResult = omit(payload, failedKeys);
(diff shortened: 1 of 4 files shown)


packages/cli/src/utils/api.js
```diff
@@ -68,6 +68,7 @@ const callAPI = async (
     method: options.method || 'GET',
     body: options.body ? JSON.stringify(options.body) : null,
     headers: {
+      ...options.extraHeaders, // extra headers first so they don't override defaults
```
rnegron, line 71: Looking for thoughts/feedback on this approach! I was surprised that I didn't find a way to override request headers. It's easier to use headers in this case due to how the API uses the entire request body.

packages/cli/src/oclif/commands/env/set.js
```diff
@@ -45,21 +45,30 @@ class SetEnvCommand extends BaseCommand {
...
 
     const url = `/apps/${app.id}/versions/${version}/multi-environment`;
+    const requestOptions = {
+      body: payload,
+      method: 'POST',
+    };
+
+    if (this.flags.force) {
+      requestOptions.extraHeaders = {
+        'X-Force-Env-Var-Update': 'true',
```
Natay, line 55: hmmm I don't see this header referenced in the `/multi-environment/` [endpoint](https://gitlab.com/zapier/zapier/-/blob/main/web/backend/zapier/app_platform/developer_cli/api/resources.py#L2462) 🤔 or the Mixin classes it's a child of, are the headers directly passed to the lambda pools or am I looking at the wrong place?

packages/cli/src/utils/api.js
Natay: This makes sense to me!

packages/cli/src/oclif/commands/env/set.js
```diff
@@ -85,7 +94,16 @@ SetEnvCommand.args = {
...
   }),
 };
-SetEnvCommand.flags = buildFlags();
+SetEnvCommand.flags = buildFlags({
+  commandFlags: {
+    force: Flags.boolean({
+      char: 'f',
+      description:
+        'Force the update of environment variables regardless if the app version is production or not. Use with caution.',
+      hidden: true,
```
eliangcs, line 103: I think it'd be more clear if `--force` is visible in the help text. Any reason to hide it?

packages/cli/src/oclif/commands/env/unset.js
```diff
@@ -42,13 +42,27 @@ class UnsetEnvCommand extends BaseCommand {
...
+
+    try {
+      await callAPI(url, requestOptions);
+    } catch (e) {
+      if (e.statusCode === 409) {
+        this.error(
+          `App version ${version} is the production version. Are you sure you want to unset potentially live environment variables?` +
+            ` If so, run this command again with the --force flag.`,
+        );
+      }
```
eliangcs, line 64: This will silently swallow any non-409 errors. Should we rethrow it?

```suggestion
      } else {
        throw e;
      }
```

packages/cli/src/utils/api.js
```diff
@@ -96,6 +97,7 @@ const callAPI = async (
   }
 
   debug(`>> ${requestOptions.method} ${requestOptions.url || res.url}`);
+  debug(`>> Request Headers: ${JSON.stringify(requestOptions.headers)}`);
```
eliangcs, line 100: Would it print sensitive data like the deploy key to the console? If so, we should avoid it.

packages/cli/src/oclif/commands/env/set.js
eliangcs: I believe [this MR](https://gitlab.com/zapier/zapier/-/merge_requests/60723) is the corresponding backend change.

eliangcs (commented):
I think the developer experience change is good. But I have some suggestions on the code. Let me know what you think. Thanks!

packages/cli/src/oclif/commands/env/set.js
rnegron: You're right; I think in my head I was hoping to discourage folks from modifying live variables by "hiding" the option. But ultimately, they will see how to override via the error shown on `409`. So let's not complicate their lives and just show the option 😆

PR #944, opened 2025-01-07
PDE-5641: feat(core) capture CLI arguments in analytics safely

Natay:
Capture the argument + flags keys ( no values ) that are being used

Related MR: https://gitlab.com/zapier/zapier/-/merge_requests/60772

Screenshot 2025-01-08 at 11 58 46 AM

Screenshot 2025-01-08 at 11 59 06 AM

### DIFF
--- a/packages/cli/src/utils/analytics.js
+++ b/packages/cli/src/utils/analytics.js
@@ -21,6 +21,24 @@ const shouldSkipAnalytics = (mode) =>
   process.env.DISABLE_ZAPIER_ANALYTICS ||
   mode === ANALYTICS_MODES.disabled;
 
+const cleanArgs = (args) => {
+  // Do not record "arguments" for following commands
+  const BLACKLISTED_ARGS = [
+    'key-value pairs...',
+    'keys...',
+    'path',
+    'email',
+    'message',
+  ];
+  const cleanedArgs = Object.keys(args)
+    .filter((key) => !BLACKLISTED_ARGS.includes(key))
+    .reduce((obj, key) => {
+      obj[key] = args[key];
+      return obj;
+    }, {});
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/analytics.js
```diff
@@ -21,13 +21,36 @@ const shouldSkipAnalytics = (mode) =>
...
+    'key-value pairs...',
+    'keys...',
+    'path',
+    'email',
+    'message',
+    'user',
+    'account',
+    'redirect-uri',
+    'inputData',
+  ];
```
eliangcs, line 36: It'd be easy to forget to update this list when developing a new CLI command or adding a command argument.

Or, do we need to capture the _values_ of arguments and flags? I think we can learn enough by capture only the _keys_ of arguments and flags.

Natay: I was thinking the extra detail enables more precise segmentation of user behavior but we don't need to be that granular right now! I'll update this!

If we want this in the future, do you think having the `doNotRecord` list as an exported asset in `ZapierBaseCommand.js` close to where new CLI arguments/commands are added would be a sufficient safeguard?

eliangcs: Thanks for updating!

Regarding:

> do you think having the `doNotRecord` list as an exported asset in `ZapierBaseCommand.js` close to where new CLI arguments/commands are added would be a sufficient safeguard?

I still don't think that would be sufficient. I think an allowlist (like `doRecord`) instead of a denylist would be a better safeguard. So, even if we forget to update the allowlist, we won't accidentally send unwanted data to the analytics endpoint.

eliangcs (approved):

PR #946, opened 2025-01-10
cli(chore): Remove browserify, replace with esbuild [PDE-5243]

kreddlear:
[Spike PR for reference](https://github.com/zapier/zapier-platform/pull/833)

This PR removes browserify, which is outdated and has caused some issues, and replaces it with esbuild.


kreddlear:
@szchenghuang asked a great question about testing [here](https://zapier.slack.com/archives/C5Z9BP4U9/p1737664626597369?thread_ts=1736870660.985229&cid=C5Z9BP4U9) - I've tested this using logic that compared the output of all the example apps for both `esbuild` and `browserify`, and tested it as well with some of my personal test apps. All resulted in the exact same results, which indicates that this change should result in minimal changes for users.

PR #950, opened 2025-01-16
fix(schema): Ensure that the default value aligns with inputField type

kola-er:
Check the description of this [ticket](https://zapierorg.atlassian.net/browse/PDE-5702) for context.


kola-er:
Closing this for a backend fix for a non-breaking approach.

PR #951, opened 2025-01-16
[PDE-5602] feat(cli): Add support for running pre-checks via API before promotion and migration

rnegron:
This PR adds logic to run soft-checks before promote/migration via a new backend endpoint. Upon receiving an `HTTP 409` from the backend, the CLI will confirm the user's intent before proceeding.

## Migration
migrate

## Promote
promote

### DIFF
--- a/packages/cli/src/oclif/commands/migrate.js
+++ b/packages/cli/src/oclif/commands/migrate.js
@@ -1,3 +1,4 @@
+const _ = require('lodash');
 const { Args, Flags } = require('@oclif/core');
 
 const BaseCommand = require('../ZapierBaseCommand');
@@ -6,6 +7,41 @@ const { callAPI } = require('../../utils/api');
 const { buildFlags } = require('../buildFlags');
 
 class MigrateCommand extends BaseCommand {
+  async run_migration_pre_checks(app, requestBody) {
+    const assumeYes = 'yes' in this.flags;
+    const url = `/apps/${app.id}/migrations/pre-checks`;
+
+    try {
+      await callAPI(
+        url,
+        {
+          method: 'POST',
+          body: requestBody,
+        },
+        true,
+      );
+    } catch (response) {
+      // 409 from the backend specifically signals pre-checks failed
+      if (response.status === 409) {
+        const softCheckErrors = _.get(response, 'json.errors');
+
+        this.log('Non-blocking checks prior to migration returned warnings:');
+        this.log(softCheckErrors);
+        this.log();
+
+        const shouldContinuePreChecks =
+          assumeYes ||
+          (await this.confirm(
+            'Would you like to continue with the migration regardless?',
+          ));
+
+        if (!shouldContinuePreChecks) {
+          this.error('Cancelled migration.');
+        }
+      }
+    }
+  }
+
   async perform() {
     const percent = this.args.percent;
     if (isNaN(percent) || percent < 1 || percent > 100) {
@@ -14,6 +50,7 @@ class MigrateCommand extends BaseCommand {
 
     const account = this.flags.account;
     const user = this.flags.user;
+
     const fromVersion = this.args.fromVersion;
     const toVersion = this.args.toVersion;
     let flagType;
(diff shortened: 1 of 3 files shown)


packages/cli/src/oclif/commands/migrate.js
```diff
@@ -6,6 +7,49 @@ const { callAPI } = require('../../utils/api');
...
+        },
+        true,
+      );
+    } catch (response) {
+      this.stopSpinner();
+
+      // 409 from the backend specifically signals pre-checks failed
+      if (response.status === 409) {
+        const softCheckErrors = _.get(response, 'json.errors');
+        const formattedErrors = softCheckErrors.map((e) => `* ${e}`).join('\n');
```
Natay, line 31: hmmm if `response.json` or `response.json.errors` is undefined, `.map()` will throw an error right? would it be viable to do something like this so we don't get runtime exceptions? 

```
const softCheckErrors = _.get(response, 'json.errors', []);
```

packages/cli/src/oclif/commands/promote.js
```diff
@@ -43,6 +43,50 @@ const hasAppChangeType = (metadata, changeType) => {
 };
 
 class PromoteCommand extends BaseCommand {
+  async run_require_confirmation_pre_checks(app, requestBody) {
```
Natay, line 46: hmmm this seems duplicative of [`run_require_confirmation_pre_checks`](https://github.com/zapier/zapier-platform/pull/951/files#diff-896ab13a96dca05c7bf8e2db52daaab026388f54621cf0d9efb24d714fb09a0dR10) in `MigrateCommand`, would it make sense to pull it into a shared resource?

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -6,6 +7,49 @@ const { callAPI } = require('../../utils/api');
...
+      await callAPI(
+        url,
+        {
+          method: 'POST',
+          body: requestBody,
+        },
+        true,
+      );
+    } catch (response) {
+      this.stopSpinner();
```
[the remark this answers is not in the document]
rnegron: Good question; we need to stop the spinner before any `this.logs()` or the terminal starts to look strange 😅 Which is why it's stopped in `catch` in which we log. We also need to stop the spinner if the `callAPI` requests succeeds to mark it as a success. So we also want that `stopSpinner` in `finally`. It could go after `callAPI` instead of inside a `finally` block, but calling `stopSpinner()` on an already-stopped spinner is a no-op ([source](https://github.com/zapier/zapier-platform/blob/019cca50f6d67e9ae9266f48c5c9f9b0df310e87/packages/cli/src/utils/display.js#L220)) so it's safe to do and looks a bit cleaner I think!

packages/cli/src/oclif/commands/migrate.js
rnegron: Good catch! Will add

packages/cli/src/oclif/commands/promote.js
rnegron: I thought about it, but then we'd also have to pass in the `jobType` for example in order to separate logs for `promote` and `migrate`, and refactor the logic further in order to accommodate the shared resource (for example, we could create a subclass for `ZapierBaseCommand` since these methods are a bit too specific to migrations for that base class.)

Personally, I believe that two similar methods is OK; three is where a refactor should be considered 😅

packages/cli/src/oclif/commands/migrate.js
```diff
@@ -6,6 +7,49 @@ const { callAPI } = require('../../utils/api');
...
+
+      // 409 from the backend specifically signals pre-checks failed
+      if (response.status === 409) {
```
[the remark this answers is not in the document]
rnegron: Adding some `debug` logs for us, just in case!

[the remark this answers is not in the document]
Natay: That makes sense, thanks for adding the debug log!

packages/cli/src/oclif/commands/promote.js
Natay: I am okay with leaving it as is 👍 !

Natay (approved):

kola-er (approved):

PR #957, opened 2025-01-28
feat(schema): PDE-5719 split FieldSchema into InputFieldSchema, OutputFieldSchema, and AuthFieldSchema

Natay:
https://zapierorg.atlassian.net/browse/PDE-5719

This pull request addresses the need to split the existing FieldSchema into three separate schemas: InputFieldSchema, OutputFieldSchema, and AuthFieldSchema. This change enhances the modularity and clarity of the schema definitions, making it easier to manage and extend them in the future for other use cases ( Agents, Canvas, unFurl )


packages/schema/lib/schemas/BaseFieldSchema.js
```diff
@@ -0,0 +1,210 @@
+'use strict';
+
+const makeSchema = require('../utils/makeSchema');
```
Natay, line 3: This is a copy of `FieldSchema.js` the spit fields inherit from, I did this to decouple the `FieldSchema` incase we need to make changes to the constraints without breaking existing behavior.

packages/schema/lib/schemas/AppSchema.js
```diff
@@ -106,6 +106,21 @@ module.exports = makeSchema(
         description: 'Top-level app options',
         $ref: AppFlagsSchema.id,
       },
+      // inputFields: {
```
Natay, line 109: @kola-er I was not sure what where the split sub-schemas should be used at the top level, this worked to generated the docs and exported-schema but didn't feel right 🤔

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const makeSchema = require('../utils/makeSchema');
+const BaseFieldSchema = require('./BaseFieldSchema');
+const schema = makeSchema(
+  {
+    id: '/AuthFieldSchema',
+    description: 'Field schema specialized for auth fields (e.g., OAuth).',
```
kola-er, line 8:
```suggestion
    description: 'Field schema specialized for auth input fields (e.g., OAuth).',
```

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,47 @@
...
+    type: 'object',
+    allOf: [
+      { $ref: BaseFieldSchema.id },
+      {
+        type: 'object',
+        properties: {
+          // New property
+          isSafe: {
+            description: 'Indicates if this auth field is safe (not secret).',
```
kola-er, line 17:
```suggestion
            description: 'Indicates if this auth input field is safe (not secret).',
```

packages/schema/lib/schemas/BaseFieldSchema.js
kola-er: Do you mean functional constraints in lib/functional-constraints? I think, as long as there are no breaking changes to the FieldSchema, any changes/additions to the constraints shouldn't block us from inheriting directly from it. Please, let's start with inheriting directly from it for the new schemas; AuthFieldSchema, InputFieldSchema, and OutputFieldSchema.

packages/schema/lib/schemas/InputFieldsSchema.js
```diff
@@ -0,0 +1,17 @@
...
+const makeSchema = require('../utils/makeSchema');
+
+const InputFieldSchema = require('./InputFieldSchema');
+
+module.exports = makeSchema(
+  {
+    id: '/InputFieldsSchema',
+    description: 'An array or collection of input fields.',
+    type: 'array',
+    items: { $ref: InputFieldSchema.id },
```
kola-er, line 12:
```suggestion
    items: {
      oneOf: [{ $ref: InputFieldSchema.id }, { $ref: FunctionSchema.id }],
    },
```

Currently, we can define input field with a function.

packages/schema/lib/schemas/BaseFieldSchema.js
Natay: Will do!

packages/schema/lib/schemas/InputFieldsSchema.js
Natay: Ahh thanks for the heads up!

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,47 @@
...
+                'api_key',
+                'password',
+                'secret',
+              ],
+            },
+            isSafe: { const: true },
+          },
+          required: ['key', 'isSafe'],
+        },
+      },
```
kola-er, line 40: This is out of scope of this ticket. Meant to be handled [separately](https://zapierorg.atlassian.net/browse/PDE-5720), but it is fine if you prefer to handle it together because it is small.

packages/schema/lib/schemas/OutputFieldSchema.js
```diff
@@ -0,0 +1,29 @@
...
+        type: 'object',
+        properties: {
+          isCanonicalUrl: { type: 'boolean' },
+          isCanonicalFile: { type: 'boolean' },
+          isTitle: { type: 'boolean' },
+          fileType: {
+            type: 'string',
+            description: "A string indicating a file's MIME type, if relevant.",
+          },
+        },
```
kola-er, line 23: This is out of scope of this ticket. Meant to be handled [separately](https://zapierorg.atlassian.net/browse/PDE-5722).

packages/schema/lib/schemas/OutputFieldsSchema.js
```diff
@@ -0,0 +1,17 @@
...
+const makeSchema = require('../utils/makeSchema');
+
+const OutputFieldSchema = require('./OutputFieldSchema');
+
+module.exports = makeSchema(
+  {
+    id: '/OutputFieldsSchema',
+    description: 'An array or collection of output fields.',
+    type: 'array',
+    items: { $ref: OutputFieldSchema.id },
```
kola-er, line 12:
```suggestion
    items: {
      oneOf: [{ $ref: OutputFieldSchema.id }, { $ref: FunctionSchema.id }],
    },
```

Ditto https://github.com/zapier/zapier-platform/pull/957/files#r1932314176

packages/schema/lib/schemas/AuthFieldSchema.js
Natay: I was thinking of handling the "smaller" field constraints in it makes perfect sense to separate them out 👍

packages/schema/lib/schemas/AppSchema.js
kola-er: They should only show up where they are referenced/used. Currently, we don't use FieldSchema at the app level. It is easier to locate those places by searching for `inputFields` and `outputFields`. You would reference these new schemas there instead of the `/DynamicFieldsSchema` currently being used. And reference the new AuthfieldsSchema in `AuthenticationSchema` for the fields property.

packages/schema/lib/schemas/AuthFieldSchema.js
kola-er: Sounds good

kola-er: Feel free to collapse both tickets into 1.

packages/schema/lib/schemas/InputFieldsSchema.js
Natay: > instead of having to create these 2 different schemas for both inputField and outputField, renaming this module to DynamicInputFieldsSchema with that code change suggestion would do.

hmmm just to clarify, this means using `DynamicInputFieldsSchema` to be something like this? so both inputField and outputField can be represented? 
```
 items: {
    oneOf: [{ $ref: InputFieldSchema.id }, { $ref: OutputFieldSchema.id }, { $ref: FunctionSchema.id }],
  },
```

kola-er: No, that of outputField would be different but similar. Doing the above, that has both represented, would mean an inputField can specify a property unique to only outputField.

I am suggesting this for `DynamicInputFieldsSchema`

```
    items: {
      oneOf: [{ $ref: InputFieldSchema.id }, { $ref: FunctionSchema.id }],
    },
```

And this for `DynamicOutputFieldsSchema`

```
    items: {
      oneOf: [{ $ref: OutputFieldSchema.id }, { $ref: FunctionSchema.id }],
    },
```

Natay: Ah that makes perfect sense, thank you!!

packages/schema/lib/schemas/DynamicAuthInputFieldsSchema.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const makeSchema = require('../utils/makeSchema');
+
+const AuthInputFieldSchema = require('./AuthInputFieldSchema');
+
+module.exports = makeSchema(
+  {
+    id: '/DynamicAuthInputFieldsSchema',
```
Natay, line 9: Tests are providing an array of field objects for `AuthenticationSchema` (e.g. [{ key: 'access_token' }, { key: 'refresh_token' }]) so we need to do this to avoid “is not of a type(s) object” errors.

Natay:
@kola-er I addressed the changes and this is ready for a re-review, I am going to add some unit tests before taking it out of draft!

packages/schema/lib/schemas/DynamicAuthInputFieldsSchema.js
kola-er:
```suggestion
    id: '/AuthFieldsSchema',
```

I suggest we don't use `Input` in the naming because there is no indication of input/output where it is used in the /AuthenticationSchema. Also, unlike for inputFields and outputFields, the `Dynamic` part is not necessary here because a function can't be used to define the auth fields.

kola-er: In the same vein, let's rename the `AuthInputFieldSchema` to `AuthFieldSchema`.

packages/schema/lib/schemas/AuthInputFieldSchema.js
```diff
@@ -0,0 +1,39 @@
...
+            'password',
+            'secret',
+          ],
+        },
+        isSafe: { const: true },
+      },
+      required: ['key', 'isSafe'],
+    },
+  },
+
```
kola-er, line 37: Let's add some examples and anti-examples specifically for authentication with the new `isSafe` property. We could add only for basic auth (email and password) and custom auth (api_key). OAuth auths don't always require these fields data from users.

packages/schema/lib/schemas/AuthInputFieldSchema.js
```diff
@@ -0,0 +1,39 @@
...
+const makeSchema = require('../utils/makeSchema');
+const FieldSchema = require('./FieldSchema');
+
+module.exports = makeSchema(
+  {
+    ...FieldSchema.schema,
+    id: '/AuthInputFieldSchema',
+    description:
+      'Field schema specialized for auth input fields (e.g., OAuth).',
```
kola-er, line 11:
```suggestion
      'Field schema specialized for authentication fields.',
```

The examples/anti-examples would give the hint on the auth types.

packages/schema/lib/schemas/DynamicAuthInputFieldsSchema.js
```diff
@@ -0,0 +1,19 @@
...
+    description: 'An array or collection of auth input fields.',
+    type: 'array',
+    items: {
+      oneOf: [{ $ref: AuthInputFieldSchema.id }],
+    },
+    examples: [[{ key: 'abc' }]],
```
kola-er, line 15:
```suggestion
    examples: [{ key: 'username', isSafe: True}, { key: 'password'}],
```

Now that the FieldSchema is broken, I think we should use specific examples for the purpose it is meant for.

packages/schema/lib/utils/makeValidator.js
```diff
@@ -64,6 +64,12 @@ const processBaseError = (err, path) => {
  */
 const cleanError = (validationError, path, validator, definition) => {
   if (ambiguousTypes.includes(validationError.name)) {
+    // If argument is not an array, we can't do .every(...)
+    if (!Array.isArray(validationError.argument)) {
+      // fallback: treat it like a standard error
+      return processBaseError(validationError, path);
+    }
+
```
kola-er, line 72: Please, I would love to see the error you are getting without this change. Just for my learning.

packages/schema/lib/schemas/DynamicAuthInputFieldsSchema.js
kola-er: That of the DynamicInputFieldsSchema and DynamicOutputFieldsSchema can stay though.

kola-er:
@Natay Looking close, thank you. I left some comments. Also, with both `FieldsSchema` and `DynamicFieldsSchema` not being used anymore, I think they can be deleted.

packages/schema/lib/utils/makeValidator.js
Natay: I was getting errors in `cleanErrors()` that were hiding the actual schema errors. It was because `cleanError` assumed the argument is always an array and this small patch is just ensure a sensible error is returned when it's not, examples of the errors I had:

```
 1) app
       validation
         should be a valid app:
     TypeError: validationError.argument.every is not a function
      at cleanError (lib/utils/makeValidator.js:70:34)
      at /Users/natay/src/github.com/zapier-platform/packages/schema/lib/utils/makeValidator.js:139:30
      at Array.map (<anonymous>)
      at Object.validate [as validateAppDefinition] (lib/utils/makeValidator.js:139:19)
      at Context.<anonymous> (test/index.js:17:30)
      at process.processImmediate (node:internal/timers:476:21)
```

packages/schema/lib/schemas/DynamicAuthInputFieldsSchema.js
Natay: Makes sense, I'll update!

packages/schema/lib/utils/makeValidator.js
kola-er: Thanks. Pulled your branch to debug further and get the exact value of the `validationError.argument` causing the error, but couldn't get it because the tests are all passing for me with this change commented out.

Please, do you mind running the test again with this change commented out?

Natay: Yeah, the tests pass with it commented, I'll remove it! 
The tests failed initially because I had an invalid schema, but I was not able to tell what was actually wrong since all the errors were `validationError.argument.every is not a function`. Once I added this small patch I was able to tell the reason behind the invalid schema and I thought the patch was harmless enough to leave in but I can leave it out for sure 👍

packages/cli/src/tests/utils/build.js
```diff
@@ -57,7 +57,7 @@ describe('build (runs slowly)', function () {
     smartPaths.filter((p) => p.endsWith('.ts')).length.should.equal(0);
     smartPaths.should.not.containEql('tsconfig.json');
 
-    smartPaths.length.should.be.within(200, 301);
+    smartPaths.length.should.be.within(200, 303);
```
Natay, line 60: [This kept failing ](https://github.com/zapier/zapier-platform/actions/runs/13035582314/job/36365075151?pr=957) consistently after the new schema additions, is this expected?

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,90 @@
...
+      // Basic Auth - email & password
+      {
+        key: 'email',
+        type: 'string',
+        isSafe: true,
+        required: true,
+      },
+      {
+        key: 'password',
+        type: 'string',
```
kola-er, line 47:
```suggestion
        type: 'password',
```

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,90 @@
...
+        isSafe: false,
+        required: true,
+      },
+    ],
+
+    antiExamples: [
+      {
+        example: {
+          key: 'password',
+          type: 'string',
```
kola-er, line 65:
```suggestion
          type: 'password',
```

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,90 @@
...
+        },
+        reason: 'A "password" field cannot have isSafe = true.',
+      },
+      {
+        example: {
+          key: 'api_key',
+          isSafe: true,
+        },
+        reason:
+          'Missing "type". Also "api_key" is a sensitive field and cannot have isSafe=true.',
```
kola-er, line 77: The `key` is the only required property of a field https://github.com/zapier/zapier-platform/blob/main/packages/schema/lib/schemas/FieldSchema.js#L27

kola-er, line 77:
```suggestion
          '"api_key" is a sensitive field and cannot have isSafe set as true.',
```

The `key` is the only required property of a field https://github.com/zapier/zapier-platform/blob/main/packages/schema/lib/schemas/FieldSchema.js#L27

packages/schema/lib/schemas/AuthFieldsSchema.js
```diff
@@ -0,0 +1,61 @@
...
+    type: 'array',
+    items: {
+      oneOf: [{ $ref: AuthFieldSchema.id }],
+    },
+    // Some complete examples
+    examples: [
+      // 1) Basic Auth: username (safe) + password (not safe)
+      [
+        { key: 'username', type: 'string', isSafe: true, required: true },
+        { key: 'password', type: 'string', isSafe: false, required: true },
```
kola-er, line 20:
```suggestion
        { key: 'password', type: 'password', isSafe: false, required: true },
```

packages/schema/lib/schemas/AuthFieldsSchema.js
```diff
@@ -0,0 +1,61 @@
...
+      ],
+
+      // 2) Just a single auth field for custom usage (e.g., api_key)
+      [{ key: 'api_key', type: 'string', isSafe: false, required: true }],
+
+      // 3) Mix of fields for extended usage
+      [
+        { key: 'email', type: 'string', isSafe: true },
+        { key: 'password', type: 'string', required: true },
```
kola-er, line 29:
```suggestion
        { key: 'password', type: 'password', required: true },
```

packages/schema/lib/schemas/AuthFieldsSchema.js
```diff
@@ -0,0 +1,61 @@
...
+        reason:
+          'api_key is considered sensitive and cannot have isSafe = true.',
+      },
+      {
+        example: [{ isSafe: false }],
+        reason: 'Missing required "key" property.',
+      },
+      {
+        example: [{ key: 'username', type: 'string', isSafe: true }, 12345],
+        reason: 'An array item is not an object (must match AuthFieldSchema).',
```
kola-er, line 56:
```suggestion
        reason: 'Array item 12345 is not an object (must match AuthFieldSchema).',
```

packages/schema/lib/schemas/AuthFieldsSchema.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+      {
+        example: [{ key: 'password', isSafe: true }],
+        reason:
+          'A password field cannot be marked as isSafe: true (violates AuthFieldSchema).',
+      },
+      {
+        example: [{ key: 'api_key', isSafe: true }],
+        reason:
+          'api_key is considered sensitive and cannot have isSafe = true.',
```
kola-er, line 48:
```suggestion
          '"api_key" is a sensitive field and cannot have isSafe set as true.',
```

packages/schema/lib/schemas/AuthFieldsSchema.js
```diff
@@ -0,0 +1,61 @@
...
+    // Anti-examples showing invalid arrays or invalid field objects
+    antiExamples: [
+      {
+        example: {},
+        reason: 'Must be an array (currently an object).',
+      },
+      {
+        example: [{ key: 'password', isSafe: true }],
+        reason:
+          'A password field cannot be marked as isSafe: true (violates AuthFieldSchema).',
```
kola-er, line 43:
```suggestion
          '"password" is a sensitive field and cannot have isSafe set as true.',
```

packages/cli/src/tests/utils/build.js
kola-er: Not sure why, but I think it is fine. It was bumped recently in this [commit](https://github.com/zapier/zapier-platform/commit/37b2b614d43bd88d0d01ba52fd86a5b07d54e92b#diff-0d2de618f30d0923bb5f97731eccdaae00205f82f016b607f5fe9ca46022069c) too.

packages/schema/test/index.js
```diff
@@ -225,6 +225,70 @@ describe('app', () => {
...
+
+      const results = schema.validateAppDefinition(appCopy);
+
+      // Expect at least one error because "password" can't have isSafe = true
+      results.errors.should.have.length(1);
+      should(results.valid).eql(false);
+
+      const [error] = results.errors;
+      // Check that the error specifically complains about a "not" rule or something similar
+      error.name.should.equal('not');
```
kola-er, line 290: Checked the error object to look for a clearer error message that would be helpful to builders when they face this invalidation, but couldn't find any.

```
ValidationError {
    property: 'instance.authentication.fields[0].',
    message: 'is of prohibited type [object Object]',
    schema: '/AuthFieldSchema',
    instance: { key: 'password', type: 'string', isSafe: true, required: true },
    name: 'not',
    argument: { properties: [Object], required: [Array] },
    stack: 'instance is of prohibited type [object Object]',
    codeLinks: [
      'https://github.com/zapier/zapier-platform/blob/zapier-platform-schema@16.2.0/packages/schema/lib/schemas/AuthFieldSchema.js'
    ],
    docLinks: [
      'https://platform.zapier.com/cli_docs/schema#authfieldschema@16.2.0'
    ]
  }
```

kola-er: Wondering if we should replace the "not" in AuthFieldSchema with a functional-contsraint in https://github.com/zapier/zapier-platform/tree/main/packages/schema/lib/functional-constraints. That way we have control over the error message. What's your thought?

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,90 @@
...
+    // Do not allow certain isSafe = True if sensitive tokens exists
+    not: {
+      properties: {
+        key: {
+          enum: [
+            'access_token',
+            'refresh_token',
+            'api_key',
+            'password',
+            'secret',
```
kola-er, line 29: @FokkeZB In case you have more auth fields to add to the list of fields that cannot be marked as isSafe.

packages/schema/test/index.js
Natay: Thats a good idea! I went with the declarative "not" by default, will change!

Natay: Just calling out it does make it harder to validate the anti-examples defined in the schema. Options seem to be 
- Remove anti-examples with `isSafe` so auto tests pass
- use `mainSchema.id` ( [example](https://github.com/zapier/zapier-platform/blob/1227d816d5542d3a3110e85f771a502b7a59d741/packages/schema/lib/functional-constraints/requiredSamples.js#L28) ) in functional-constraint so the sub-schema's also pass through to these validations 

I went with option 2 here so we can use the anti-examples as they are and have it pass validation rules!

kola-er: Just learning this, thanks. I agree, option 2 is better.

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,64 @@
...
+  'secret',
+];
+
+const checkAuthField = (field) => {
+  const errors = [];
+
+  if (FORBIDDEN_KEYS.includes(field.key) && field.isSafe === true) {
+    errors.push(
+      new jsonschema.ValidationError(
+        `Cannot set isSafe=true for the sensitive key "${field.key}".`,
```
kola-er, line 22:
```suggestion
        `cannot set isSafe as true for the sensitive field with key "${field.key}".`,
```

Starting with a lowercase because the `err.stack` is formed by prepending the `err.property` to this message to form a sentence.

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,64 @@
...
+        field,
+        '/AuthFieldSchema',
+        'instance.key',
```
kola-er, line 25:
```suggestion
        'instance.field',
```

Because we are dealing with the entire auth field

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,64 @@
...
+        'sensitive',
+        'key',
```
kola-er, line 27:
```suggestion
        'field',
```

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  // If there's no authentication or no fields, we have nothing to check
+  if (!definition.authentication || !definition.authentication.fields) {
+    return errors;
+  }
+
+  definition.authentication.fields.forEach((field, index) => {
+    checkAuthField(field).forEach((err) => {
+      err.property = `instance[${index}]`;
```
kola-er, line 57:
```suggestion
      err.property = `instance.authentication.fields[${index}]`;
```

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,64 @@
...
+      err.stack = err.stack.replace('instance.key', `instance[${index}].key`);
```
kola-er, line 58:
```suggestion
      err.stack = err.stack.replace('instance.field', err.property);
```

packages/schema/lib/utils/makeValidator.js
```diff
@@ -64,6 +64,7 @@ const processBaseError = (err, path) => {
  */
 const cleanError = (validationError, path, validator, definition) => {
   if (ambiguousTypes.includes(validationError.name)) {
+    console.log(validationError, 'hereeeee');
```
kola-er, line 67:
```suggestion
```

Forgotten?

packages/schema/test/index.js
```diff
@@ -225,6 +225,74 @@ describe('app', () => {
...
+            isSafe: true, // invalid: password cannot be safe
+            required: true,
+          },
+        ],
+      };
+
+      const results = schema.validateAppDefinition(appCopy);
+
+      // Expect at least one error because "password" can't have isSafe = true
+      console.log(results);
```
kola-er, line 285:
```suggestion
```

Forgotten?

packages/schema/test/index.js
```diff
@@ -225,6 +225,74 @@ describe('app', () => {
...
+      results.errors.should.have.length(1);
+      should(results.valid).eql(false);
+
+      const [error] = results.errors;
+      // Check that the error specifically complains about a "not" rule or something similar
```
kola-er, line 290:
```suggestion
```

Not needed anymore

packages/schema/test/index.js
```diff
@@ -225,6 +225,74 @@ describe('app', () => {
...
+      error.name.should.equal('sensitive');
+      error.message.should.equal(
+        'Cannot set isSafe=true for the sensitive key "password".',
+      );
```
kola-er, line 294:
```suggestion
      error.stack.should.eql(
        'instance.authentication.fields[0] cannot set isSafe as true for the sensitive key "password".',
      );
```

packages/schema/lib/utils/makeValidator.js
Natay: Yup 🤦

packages/schema/lib/schemas/AuthFieldSchema.js
FokkeZB: @kola-er you may want to copy https://gitlab.com/zapier/team-developer-platform/secret-scrubber-js/-/blob/main/src/convenience.ts?ref_type=heads#L6-L19

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+* [**Source Code**](https://github.com/zapier/zapier-platform/blob/zapier-platform-schema@16.2.0/packages/schema/lib/schemas/AuthFieldSchema.js)
+
+#### Properties
+
+Key | Required | Type | Description
+--- | -------- | ---- | -----------
+`key` | **yes** | `string` | A unique machine readable key for this value (IE: "fname").
+`label` | no | `string` | A human readable label for this value (IE: "First Name").
+`helpText` | no | `string` | A human readable description of this value (IE: "The first part of a full name."). You can use Markdown.
+`type` | no | `string` in (`'string'`, `'text'`, `'integer'`, `'number'`, `'boolean'`, `'datetime'`, `'file'`, `'password'`, `'copy'`, `'code'`) | The type of this value. Use `string` for basic text input, `text` for a large, `<textarea>` style box, and `code` for a `<textarea>` with a fixed-width font. Field type of `file` will accept either a file object or a string. If a URL is provided in the string, Zapier will automatically make a GET for that file. Otherwise, a .txt file will be generated.
```
FokkeZB, line 167: This should be limited to what actually works.

Just from [rendering](https://cdn.zappy.app/07033c49fc406de8261e71fbf8108335.gif) it seems `code` doesn't, but I also doubt `file` would actually process well.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`required` | no | `boolean` | If this value is required or not.
```
FokkeZB, line 168: For auth fields this defaults to `true`, which should be documented.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`placeholder` | no | `string` | An example value that is not saved.
+`default` | no | `string` | A default value that is saved the first time a Zap is created.
+`primary` | no | `boolean` | Use this field as part of the primary key for deduplication. You can set multiple fields as "primary", provided they are unique together. If no fields are set, Zapier will default to using the `id` field. `primary` only makes sense for `outputFields`; it will be ignored if set in `inputFields`. It only works in static `outputFields`; will not work in custom/dynamic `outputFields`. For more information, see [How deduplication works in Zapier](https://platform.zapier.com/build/deduplication).
```
FokkeZB, line 171: Not relevant to auth fields.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`dynamic` | no | [/RefResourceSchema](#refresourceschema) | A reference to a trigger that will power a dynamic dropdown.
```
FokkeZB, line 172: This renders, but doesn't show choices, so should be dropped from this schema:

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`search` | no | [/RefResourceSchema](#refresourceschema) | A reference to a search that will guide the user to add a search step to populate this field when creating a Zap.
```
FokkeZB, line 173: Not relevant to auth fields.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`choices` | no | [/FieldChoicesSchema](#fieldchoicesschema) | An object of machine keys and human values to populate a static dropdown.
+`list` | no | `boolean` | Acts differently when used in inputFields vs. when used in outputFields. In inputFields: Can a user provide multiples of this field? In outputFields: Does this field return an array of items of type `type`?
+`children` | no | `array`[[/FieldSchema](#fieldschema)] | An array of child fields that define the structure of a sub-object for this field. Usually used for line items.
+`dict` | no | `boolean` | Is this field a key/value input?
+`computed` | no | `boolean` | Is this field automatically populated (and hidden from the user)? Note: Only OAuth and Session Auth support fields with this key.
+`altersDynamicFields` | no | `boolean` | Does the value of this field affect the definitions of other fields in the set?
```
FokkeZB, line 179: Doesn't work in the auth dialog, so should be dropped until it does.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`steadyState` | no | `boolean` | Prevents triggering on new output until all values for fields with this property remain unchanged for 2 polls. It can be used to, e.g., not trigger on a new contact until the contact has completed typing their name. NOTE that this only applies to the `outputFields` of polling triggers.
```
FokkeZB, line 180: Not relevant to auth fields.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`inputFormat` | no | `string` | Useful when you expect the input to be part of a longer string. Put "{{input}}" in place of the user's input (IE: "https://{{input}}.yourdomain.com").
+`meta` | no | [/FieldMetaSchema](#fieldmetaschema) | Allows for additional metadata to be stored on the field. Supports simple key-values only (no sub-objects or arrays).
```
FokkeZB, line 182: Not relevant to auth fields.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +148,91 @@ Key | Required | Type | Description
...
+`isSafe` | no | `boolean` | Indicates if this authentication field is safe (not secret).
```
FokkeZB, line 183: I'm not sure about the name of this new property. Safe for what? Perhaps `isNoSecret` is more explicit? I know it violates some style guides because it is a negation, but I find it more clear.

packages/schema/docs/build/schema.md
```diff
@@ -1211,6 +1269,68 @@ Key | Required | Type | Description
...
+`list` | no | `boolean` | Acts differently when used in inputFields vs. when used in outputFields. In inputFields: Can a user provide multiples of this field? In outputFields: Does this field return an array of items of type `type`?
+`children` | no | `array`[[/FieldSchema](#fieldschema)] | An array of child fields that define the structure of a sub-object for this field. Usually used for line items.
+`dict` | no | `boolean` | Is this field a key/value input?
+`computed` | no | `boolean` | Is this field automatically populated (and hidden from the user)? Note: Only OAuth and Session Auth support fields with this key.
+`altersDynamicFields` | no | `boolean` | Does the value of this field affect the definitions of other fields in the set?
+`steadyState` | no | `boolean` | Prevents triggering on new output until all values for fields with this property remain unchanged for 2 polls. It can be used to, e.g., not trigger on a new contact until the contact has completed typing their name. NOTE that this only applies to the `outputFields` of polling triggers.
```
FokkeZB, line 1301: Not relevant for input fields and should does be dropped

packages/schema/docs/build/schema.md
```diff
@@ -1211,6 +1269,68 @@ Key | Required | Type | Description
...
+`dynamic` | no | [/RefResourceSchema](#refresourceschema) | A reference to a trigger that will power a dynamic dropdown.
+`search` | no | [/RefResourceSchema](#refresourceschema) | A reference to a search that will guide the user to add a search step to populate this field when creating a Zap.
+`choices` | no | [/FieldChoicesSchema](#fieldchoicesschema) | An object of machine keys and human values to populate a static dropdown.
+`list` | no | `boolean` | Acts differently when used in inputFields vs. when used in outputFields. In inputFields: Can a user provide multiples of this field? In outputFields: Does this field return an array of items of type `type`?
+`children` | no | `array`[[/FieldSchema](#fieldschema)] | An array of child fields that define the structure of a sub-object for this field. Usually used for line items.
+`dict` | no | `boolean` | Is this field a key/value input?
+`computed` | no | `boolean` | Is this field automatically populated (and hidden from the user)? Note: Only OAuth and Session Auth support fields with this key.
```
FokkeZB, line 1299: Not relevant for input fields and should does be dropped

packages/schema/docs/build/schema.md
```diff
@@ -1211,6 +1269,68 @@ Key | Required | Type | Description
...
+`required` | no | `boolean` | If this value is required or not.
+`placeholder` | no | `string` | An example value that is not saved.
+`default` | no | `string` | A default value that is saved the first time a Zap is created.
+`primary` | no | `boolean` | Use this field as part of the primary key for deduplication. You can set multiple fields as "primary", provided they are unique together. If no fields are set, Zapier will default to using the `id` field. `primary` only makes sense for `outputFields`; it will be ignored if set in `inputFields`. It only works in static `outputFields`; will not work in custom/dynamic `outputFields`. For more information, see [How deduplication works in Zapier](https://platform.zapier.com/build/deduplication).
```
FokkeZB, line 1292: Not relevant for input fields and should does be dropped

packages/schema/docs/build/schema.md
```diff
@@ -1288,6 +1408,68 @@ List of before or after middlewares. Can be an array of functions or a single fu
...
+`children` | no | `array`[[/FieldSchema](#fieldschema)] | An array of child fields that define the structure of a sub-object for this field. Usually used for line items.
+`dict` | no | `boolean` | Is this field a key/value input?
+`computed` | no | `boolean` | Is this field automatically populated (and hidden from the user)? Note: Only OAuth and Session Auth support fields with this key.
+`altersDynamicFields` | no | `boolean` | Does the value of this field affect the definitions of other fields in the set?
+`steadyState` | no | `boolean` | Prevents triggering on new output until all values for fields with this property remain unchanged for 2 polls. It can be used to, e.g., not trigger on a new contact until the contact has completed typing their name. NOTE that this only applies to the `outputFields` of polling triggers.
+`inputFormat` | no | `string` | Useful when you expect the input to be part of a longer string. Put "{{input}}" in place of the user's input (IE: "https://{{input}}.yourdomain.com").
+`meta` | no | [/FieldMetaSchema](#fieldmetaschema) | Allows for additional metadata to be stored on the field. Supports simple key-values only (no sub-objects or arrays).
```
FokkeZB, line 1442: Not relevant for output fields and should does be dropped.

packages/schema/docs/build/schema.md
```diff
@@ -1288,6 +1408,68 @@ List of before or after middlewares. Can be an array of functions or a single fu
...
+`dynamic` | no | [/RefResourceSchema](#refresourceschema) | A reference to a trigger that will power a dynamic dropdown.
+`search` | no | [/RefResourceSchema](#refresourceschema) | A reference to a search that will guide the user to add a search step to populate this field when creating a Zap.
+`choices` | no | [/FieldChoicesSchema](#fieldchoicesschema) | An object of machine keys and human values to populate a static dropdown.
+`list` | no | `boolean` | Acts differently when used in inputFields vs. when used in outputFields. In inputFields: Can a user provide multiples of this field? In outputFields: Does this field return an array of items of type `type`?
+`children` | no | `array`[[/FieldSchema](#fieldschema)] | An array of child fields that define the structure of a sub-object for this field. Usually used for line items.
+`dict` | no | `boolean` | Is this field a key/value input?
+`computed` | no | `boolean` | Is this field automatically populated (and hidden from the user)? Note: Only OAuth and Session Auth support fields with this key.
+`altersDynamicFields` | no | `boolean` | Does the value of this field affect the definitions of other fields in the set?
```
FokkeZB, line 1439: Not relevant for output fields and should does be dropped.

packages/schema/docs/build/schema.md
```diff
@@ -1288,6 +1408,68 @@ List of before or after middlewares. Can be an array of functions or a single fu
...
+`required` | no | `boolean` | If this value is required or not.
+`placeholder` | no | `string` | An example value that is not saved.
+`default` | no | `string` | A default value that is saved the first time a Zap is created.
+`primary` | no | `boolean` | Use this field as part of the primary key for deduplication. You can set multiple fields as "primary", provided they are unique together. If no fields are set, Zapier will default to using the `id` field. `primary` only makes sense for `outputFields`; it will be ignored if set in `inputFields`. It only works in static `outputFields`; will not work in custom/dynamic `outputFields`. For more information, see [How deduplication works in Zapier](https://platform.zapier.com/build/deduplication).
+`dynamic` | no | [/RefResourceSchema](#refresourceschema) | A reference to a trigger that will power a dynamic dropdown.
+`search` | no | [/RefResourceSchema](#refresourceschema) | A reference to a search that will guide the user to add a search step to populate this field when creating a Zap.
+`choices` | no | [/FieldChoicesSchema](#fieldchoicesschema) | An object of machine keys and human values to populate a static dropdown.
```
FokkeZB, line 1434: Not relevant for output fields and should does be dropped.

packages/schema/docs/build/schema.md
```diff
@@ -1288,6 +1408,68 @@ List of before or after middlewares. Can be an array of functions or a single fu
...
+
+Key | Required | Type | Description
+--- | -------- | ---- | -----------
+`key` | **yes** | `string` | A unique machine readable key for this value (IE: "fname").
+`label` | no | `string` | A human readable label for this value (IE: "First Name").
+`helpText` | no | `string` | A human readable description of this value (IE: "The first part of a full name."). You can use Markdown.
+`type` | no | `string` in (`'string'`, `'text'`, `'integer'`, `'number'`, `'boolean'`, `'datetime'`, `'file'`, `'password'`, `'copy'`, `'code'`) | The type of this value. Use `string` for basic text input, `text` for a large, `<textarea>` style box, and `code` for a `<textarea>` with a fixed-width font. Field type of `file` will accept either a file object or a string. If a URL is provided in the string, Zapier will automatically make a GET for that file. Otherwise, a .txt file will be generated.
+`required` | no | `boolean` | If this value is required or not.
+`placeholder` | no | `string` | An example value that is not saved.
+`default` | no | `string` | A default value that is saved the first time a Zap is created.
```
FokkeZB, line 1430: Not relevant for input fields and should does be dropped.

packages/schema/docs/build/schema.md
```diff
@@ -1288,6 +1408,68 @@ List of before or after middlewares. Can be an array of functions or a single fu
...
+* **Type** - `object`
+* [**Source Code**](https://github.com/zapier/zapier-platform/blob/zapier-platform-schema@16.2.0/packages/schema/lib/schemas/OutputFieldSchema.js)
+
+#### Properties
+
+Key | Required | Type | Description
+--- | -------- | ---- | -----------
+`key` | **yes** | `string` | A unique machine readable key for this value (IE: "fname").
+`label` | no | `string` | A human readable label for this value (IE: "First Name").
+`helpText` | no | `string` | A human readable description of this value (IE: "The first part of a full name."). You can use Markdown.
```
FokkeZB, line 1426: Not relevant for output fields and should does be dropped.

packages/schema/docs/build/schema.md
FokkeZB, line 1430: Not relevant for output fields and should does be dropped.

packages/schema/docs/build/schema.md
```diff
@@ -1288,6 +1408,68 @@ List of before or after middlewares. Can be an array of functions or a single fu
...
+`type` | no | `string` in (`'string'`, `'text'`, `'integer'`, `'number'`, `'boolean'`, `'datetime'`, `'file'`, `'password'`, `'copy'`, `'code'`) | The type of this value. Use `string` for basic text input, `text` for a large, `<textarea>` style box, and `code` for a `<textarea>` with a fixed-width font. Field type of `file` will accept either a file object or a string. If a URL is provided in the string, Zapier will automatically make a GET for that file. Otherwise, a .txt file will be generated.
```
FokkeZB, line 1427: We should drop the types that make no sense for output fields, which at least are: `text` (as for output it's just about the value type, not the type of input field), `integer` (same), `copy` and `code`.

Not sue about `datetime` and `password`, but little harm in letting them be. And we'll need `file` for unfurl.

FokkeZB (commented):
Perhaps this was still TBD, but the point of splitting the schema is that they can be tailored. I pointed out quite a few places where e.g. the input field schema has properties that only make sense in auth fields.

Natay:
> Perhaps this was still TBD, but the point of splitting the schema is that they can be tailored. I pointed out quite a few places where e.g. the input field schema has properties that only make sense in auth fields.

Thanks for calling it out! This was initially intended to not be a breaking change so it can be released before v17 ( sometime in March ) so we inherited from FieldSchema, but with unfurl moving around to later we can make the split schemas tailored to their functionality like you suggested and have it as a breaking change in v17.

packages/schema/docs/build/schema.md
Natay: Good point, maybe `isConfidential` or `isSecret` convey the intent better without need for the negation?

FokkeZB: @Natay I don't believe that is an option, as we currently consider all auth fields secret, so the new flag needs to opt fields out of that.

Natay: Ahh that makes sense, I'll have it as `isNotSecret` and add a note then, thanks for the heads up!

packages/schema/docs/build/schema.md
Natay: That's really neat! How did you get it to render? I tried running the monolith and editor locally then create an app using CLI locally, but I wasn't able to see it via the editor, any tips would be awesome!

FokkeZB: Yes, that's what I did as well.

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +147,82 @@ Key | Required | Type | Description
...
+
+* **Type** - `object`
+* [**Source Code**](https://github.com/zapier/zapier-platform/blob/zapier-platform-schema@16.3.0/packages/schema/lib/schemas/AuthFieldSchema.js)
+
+#### Properties
+
+Key | Required | Type | Description
+--- | -------- | ---- | -----------
+`key` | **yes** | `string` | A unique machine readable key for this value (IE: "fname").
+`isNotSecret` | no | `boolean` | Indicates if this authentication field is safe (not secret).
```
FokkeZB, line 164: Nitpick, but perhaps drop `t` and give example of what it is safe for:

```suggestion
`isNoSecret` | no | `boolean` | Indicates if this authentication field is safe to e.g. be stored without encryption or displayed (not a secret).
```

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +147,82 @@ Key | Required | Type | Description
...
+`label` | no | `string` | A human readable label for this value (IE: "First Name").
+`helpText` | no | `string` | A human readable description of this value (IE: "The first part of a full name."). You can use Markdown.
+`type` | no | `string` in (`'string'`, `'number'`, `'boolean'`, `'datetime'`, `'password'`, `'copy'`) | The type of this value. Field type of `file` will accept either a file object or a string. If a URL is provided in the string, Zapier will automatically make a GET for that file. Otherwise, a .txt file will be generated.
```
FokkeZB, line 167: The description refers to `file` even though that is no listed as a valid value:

```suggestion
`type` | no | `string` in (`'string'`, `'number'`, `'boolean'`, `'datetime'`, `'password'`, `'copy'`) | The type of this value.
```

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +147,82 @@ Key | Required | Type | Description
...
+`required` | no | `boolean` | If this value is required or not. This defaults to True.
```
FokkeZB, line 168:
```suggestion
`required` | no | `boolean` | If this value is required or not. This defaults to `true`.
```

packages/schema/docs/build/schema.md
```diff
@@ -145,6 +147,82 @@ Key | Required | Type | Description
...
+`placeholder` | no | `string` | An example value that is not saved.
```
FokkeZB, line 169: I do believe we also support `default`, which is no longer listed.

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,78 @@
...
+  FORBIDDEN_KEYS.some((forbidden) =>
+    key.toLowerCase().includes(forbidden.toLowerCase()),
+  );
+
+const checkAuthField = (field) => {
+  const errors = [];
+
+  // if the field key contains any forbidden substring (case-insensitive),
+  // AND 'isSecret' is true, throw a validation error
+  if (isSensitiveKey(field.key) && field.isNotSecret === true) {
```
FokkeZB, line 35: I would also prevent fields with type `password` to be marked as no secret:

```suggestion
  if ((field.key === 'password' || isSensitiveKey(field.key)) && field.isNotSecret === true) {
```

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,78 @@
...
+const isSensitiveKey = (key = '') =>
+  FORBIDDEN_KEYS.some((forbidden) =>
+    key.toLowerCase().includes(forbidden.toLowerCase()),
+  );
+
+const checkAuthField = (field) => {
+  const errors = [];
+
+  // if the field key contains any forbidden substring (case-insensitive),
+  // AND 'isSecret' is true, throw a validation error
```
FokkeZB, line 34:
```suggestion
  // AND 'isNotSecret' is true, throw a validation error
```

FokkeZB (commented):
Left a few last things

packages/schema/docs/build/schema.md
kola-er: Argh, Natay just called my attention to this suggestion being for AuthFieldSchema and not necessarily for InputFieldSchema. I agree then. Thanks @FokkeZB

Natay: Thanks both for the guidance! I was able to [render](https://cdn.zappy.app/dab1dd0f286c5e998a7dd650aa6ccfad.gif) and keep only what works.

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,128 @@
...
+      type: 'string',
+      minLength: 1,
+    },
+    default: {
+      description: 'A default value for authentication field.',
+      type: 'string',
+      minLength: 1,
+    },
+    list: {
+      description: 'Can a user provide multiples of this field?',
```
kola-er, line 54: `choices` and `computed` are missing.

packages/schema/lib/schemas/InputFieldSchema.js
```diff
@@ -0,0 +1,176 @@
...
+      default: {
+        description:
+          'A default value that is saved the first time a Zap is created.',
+        type: 'string',
+        minLength: 1,
+      },
+      dict: {
+        description: 'Is this field a key/value input?',
+        type: 'boolean',
+      },
```
kola-er, line 95: `placeholder` is missing

packages/schema/lib/schemas/OutputFieldSchema.js
```diff
@@ -0,0 +1,113 @@
...
+        type: 'array',
+        items: { $ref: '/OutputFieldSchema' },
+        description:
+          'An array of child fields that define the structure of a sub-object for this field. Usually used for line items.',
+        minItems: 1,
+      },
+      dict: {
+        description: 'Is this field a key/value output?',
+        type: 'boolean',
+      },
```
kola-er, line 63: `steadyState` is missing

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,109 @@
...
+        type: 'string',
+        minLength: 1,
+      },
+      inputFormat: {
+        description:
+          'Useful when you expect the input to be part of a longer string. Put "{{input}}" in place of the user\'s input (IE: "https://{{input}}.yourdomain.com").',
+        type: 'string',
+        // TODO: Check if it contains one and ONLY ONE '{{input}}'
+        pattern: '^.*{{input}}.*$',
+      },
```
kola-er, line 53: Seems we need to overwrite the `default` property too to indicate its value is `true` by default for auth.

Trivial but can we rearrange as close as possible to the order in the old FieldSchema. So I would expect the order to be:
- helpText
- type
- placeholder
- default
- choices
- computed
- inputFormat
- isNotSecret

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,109 @@
...
+
+    antiExamples: [
+      {
+        example: {
+          key: 'password',
+          type: 'password',
+          isNoSecret: true,
+          required: true,
+        },
+        reason: 'A "password" field cannot have isSafe = true.',
```
kola-er, line 89:
```suggestion
        reason: 'A "password" field cannot have isNotSecret = true.',
```

kola-er: Also, the message format needs to be consistent with others like the example on api_key.

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,109 @@
...
+      },
+      {
+        example: {
+          key: 'api_key',
+          isNoSecret: true,
+        },
+        reason:
+          '"api_key" is a sensitive field and cannot have isSafe set as true.',
```
kola-er, line 97:
```suggestion
          '"api_key" is a sensitive field and cannot have isNotSecret set as true.',
```

packages/schema/lib/schemas/FieldSchema.js
```diff
@@ -19,192 +13,132 @@ const formatBullet = (f) => `* ${f.map(wrapInBackticks).join(' & ')}`;
...
-      {
-        example: { key: 'abc', children: ['$func$2$f$'] },
-        reason:
-          'Invalid value for key: children (must be array of FieldSchema)',
-      },
-    ],
-    additionalProperties: false,
+module.exports = makeSchema({
+  id: '/FieldSchema',
+  description: `Defines a field an app either needs as input, or gives as output. In addition to the requirements below, the following keys are mutually exclusive:\n\n${incompatibleFieldsList}`,
```
kola-er, line 18:
```suggestion
  description: `In addition to the requirements below, the following keys are mutually exclusive:\n\n${incompatibleFieldsList}`,
```

Intended to be appended to the description of each of the schemas inheriting from FieldSchema.

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,109 @@
+'use strict';
+
+const makeSchema = require('../utils/makeSchema');
+const FieldChoicesSchema = require('./FieldChoicesSchema');
+const FieldSchema = require('./FieldSchema');
+
+module.exports = makeSchema(
+  {
+    id: '/AuthFieldSchema',
+    description: 'Field schema specialized for authentication fields.',
```
kola-er, line 10:
```suggestion
    description: 'Field schema specialized for authentication fields. ${FieldSchema.schema.description}',
```

packages/schema/lib/schemas/InputFieldSchema.js
```diff
@@ -0,0 +1,124 @@
...
+
+const makeSchema = require('../utils/makeSchema');
+const RefResourceSchema = require('./RefResourceSchema');
+const FieldChoicesSchema = require('./FieldChoicesSchema');
+const FieldSchema = require('./FieldSchema');
+const FieldMetaSchema = require('./FieldMetaSchema');
+
+module.exports = makeSchema(
+  {
+    description: 'Field schema specialized for input fields.',
```
kola-er, line 11:
```suggestion
    description: 'Field schema specialized for input fields. ${FieldSchema.schema.description}',
```

Based on this suggestion https://github.com/zapier/zapier-platform/pull/957/files#r1940592149

packages/schema/lib/schemas/OutputFieldSchema.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const makeSchema = require('../utils/makeSchema');
+const FieldSchema = require('./FieldSchema');
+
+module.exports = makeSchema(
+  {
+    description: 'Field schema specialized for output fields.',
```
kola-er, line 8:
```suggestion
    description: 'Field schema specialized for output fields. ${FieldSchema.schema.description}',
```

Based on this suggestion https://github.com/zapier/zapier-platform/pull/957/files#r1940592149

packages/schema/lib/schemas/FieldSchema.js
```diff
@@ -19,192 +13,132 @@ const formatBullet = (f) => `* ${f.map(wrapInBackticks).join(' & ')}`;
...
+    {
+      // default must be at least 1 character long
+      example: { key: 'abc', default: '' },
+      reason: 'Invalid value for key: default (cannot be empty string)',
+    },
+    {
+      // key must be at least 1 character long
+      example: { key: '' },
+      reason: 'Invalid value for key: key (cannot be empty string)',
+    },
```
kola-er, line 124: Since the FieldSchema is not actually inherited in each of the new 3 schemas, these examples and antiExamples wouldn't be used in any auto-validation for fields' values. I think it is safe to remove both.

packages/schema/test/readability.js
```diff
@@ -86,13 +86,13 @@ describe('readability', () => {
...
         sample: { id: 1 },
-        inputFields: [{ key: 'field', type: 'string', default: '' }],
+        inputFields: [{ key: 'field', type: 'string', children: [] }],
       },
     });
     results.errors.should.have.length(1);
     should(results.errors[0].property.endsWith('instance')).be.false();
     results.errors[0].property.should.eql(
-      'instance.operation.inputFields[0].default',
+      'instance.operation.inputFields[0].children',
```
kola-er, line 95: Why the change? Was it failing?

packages/schema/test/index.js
```diff
@@ -225,6 +225,73 @@ describe('app', () => {
...
+      };
+
+      const results = schema.validateAppDefinition(appCopy);
+
+      // Expect at least one error because "password" can't have isNoSecret = true
+      results.errors.should.have.length(1);
+      should(results.valid).eql(false);
+
+      const [error] = results.errors;
+      console.log(error);
```
kola-er, line 289:
```suggestion
```

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,81 @@
...
+  'refresh_token',
+  'secret',
+  'set-cookie',
+  'signature',
+  'token',
+];
+
+const isSensitiveKey = (key = '') =>
+  FORBIDDEN_KEYS.some((forbidden) =>
+    key.toLowerCase().includes(forbidden.toLowerCase()),
```
kola-er, line 27:
```suggestion
    key.toLowerCase().includes(forbidden),
```

Already in lower case

packages/schema/lib/functional-constraints/AuthFieldisSafe.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const jsonschema = require('jsonschema');
+
+const AUTH_FIELD_ID = '/AuthFieldSchema';
+const AUTH_FIELDS_ID = '/AuthFieldsSchema';
+
+const FORBIDDEN_KEYS = [
+  'access_token',
```
kola-er, line 9: Just like the different versions of api_key, perhaps, we should do the same for access_token, refresh_token, and set-cookie as follows:

- access_token
- access-token
- accesstoken
- refresh_token
- refresh-token
- refreshtoken
- set-cookie
- set_cookie
- setcookie

packages/schema/docs/build/schema.md
kola-er: @FokkeZB Fine to go with `isNoSecret`, but just wanted to point out that the description could do the work of explaining what `isSafe` means like what we currently have and still use it:

`Indicates if this authentication field is safe to e.g. be stored without encryption or displayed (not a secret).`

kola-er:
@Natay  Please git squash all the commits into 1 commit of "Split FieldSchema into AuthFieldSchema, InputFieldSchema, OutputFieldSchema"

packages/schema/test/readability.js
Natay: It was failing earlier because default was removed and I changed it, passes now with `default` but since they effectively test the same thing I left it as is. I'll change it back so the # of files changes decreases at least!

packages/schema/lib/schemas/FieldSchema.js
Natay: hmmm the [auto-validation for FieldSchema](https://cdn.zappy.app/f582cc546189a5cced52e07541486da7.gif) does seem to run the examples/anti-examples though, when removed it shows as "2 tests pending" for the schema

Natay:
@kola-er I plan on squashing before merging if that's okay! 
Screenshot 2025-02-04 at 8 26 08 AM

packages/schema/lib/schemas/FieldSchema.js
kola-er: Sounds good

kola-er: It seems you forgot to return them; they are missing.

packages/schema/lib/schemas/AuthFieldSchema.js
```diff
@@ -0,0 +1,116 @@
...
+        description: 'The type of this value used to be.',
+        type: 'string',
+        enum: ['string', 'number', 'boolean', 'datetime', 'password'],
+      },
+      required: {
+        description:
+          'If this value is required or not. This defaults to `true`.',
+        type: 'boolean',
+      },
+
```
kola-er, line 32:
```suggestion
```

packages/schema/lib/schemas/FieldSchema.js
Natay: Yup added them back in now!

kola-er (approved):
Thanks for your effort. Looks good to go.

Natay:
Thank you for all the guidance here @kola-er @FokkeZB !

PR #960, opened 2025-01-29
[PDE-5716] feat(cli): Require `--overwrite-partner-changes` flag when a `zapier push` from staff could impact partner changes

rnegron:
This PR is similar to https://github.com/zapier/zapier-platform/pull/942, where we added a `--force` option to `env:set` and `env:unset`. 

In this MR, we add a `--overwrite-partner-changes` flag to force a `zapier push` to skip a new check when the user is a Zapier staff and the latest integration version contains partner-pushed code ("clobbering" protection).

Screen Shot 2025-01-29 at 4 09 09 PM

### DIFF
--- a/packages/cli/src/oclif/commands/push.js
+++ b/packages/cli/src/oclif/commands/push.js
@@ -1,10 +1,10 @@
 const ZapierBaseCommand = require('../ZapierBaseCommand');
+const { BUILD_PATH, SOURCE_PATH } = require('../../constants');
 const { Flags } = require('@oclif/core');
 
 const BuildCommand = require('./build');
 
 const { buildAndOrUpload } = require('../../utils/build');
-const { BUILD_PATH, SOURCE_PATH } = require('../../constants');
 
 class PushCommand extends ZapierBaseCommand {
   async perform() {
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/api.js
```diff
@@ -449,17 +452,42 @@ const upload = async (app, { skipValidation = false } = {}) => {
...
+      {
+        method: 'PUT',
+        body: {
+          zip_file: buffer,
+          source_zip_file: sourceBuffer,
+          skip_validation: skipValidation,
+        },
+        extraHeaders: headers,
+      },
+      true,
```
rnegron, line 473: This `true` adds `rawError = True` so we can catch a response as `err` and can do `err.status`.

packages/cli/src/oclif/commands/push.js
```diff
@@ -21,7 +23,14 @@ class PushCommand extends ZapierBaseCommand {
   }
 }
 
-PushCommand.flags = BuildCommand.flags;
+PushCommand.flags = {
+  ...BuildCommand.flags,
+  'overwrite-partner-changes': Flags.boolean({
+    description:
+      '(Internal Use Only) Allows Zapier Staff to push changes to integrations in certain situations.',
+    hidden: true,
```
kola-er, line 31: 👍

kola-er (approved):

PR #965, opened 2025-02-03
feat(cli, core): Add ESM support to CLI and platform-core [PDE-5244]

kreddlear:
This patch adds ECMAScript Modules (ESM) support to the Zapier CLI and platform core, alongside the existing CommonJS module system. Here are the key changes:

### 1. Adding ESM Support

* Creates a new `zapierwrapper.mjs` file for ESM integrations
* Adds utilities to copy the appropriate `zapierwrapper` (`.js` or `.mjs`) based on the integration's module type
* Handles both named exports and default exports in entry points
* Adds comprehensive tests for both CommonJS and ESM formats

### 2. Build Process Improvements

* Updates the build system to properly handle ESM integrations
* Modifies how dependencies are detected and bundled
* Enhances how entry points are identified and processed
* Improves the zip file creation process to work with both module formats

### 3. App Loading Refactoring

* Rewrites `loadApp` to handle both string paths and direct app objects
* Updates the Lambda handler creation process to support ESM
* Modernizes code to use async/await patterns where appropriate

### 4. Testing

* Adds extensive test cases for the ESM functionality
* Ensures backward compatibility with existing CommonJS integrations
* Tests various package.json configurations for proper module resolution

This represents a significant enhancement allowing Zapier integrations to use modern JavaScript module syntax while maintaining compatibility with existing code.


packages/core/include/zapierwrapper.mjs
```diff
@@ -0,0 +1,22 @@
+// not intended to be loaded via require() - copied during build step
+import { createRequire } from 'module';
+import path from 'node:path';
+import zapier from 'zapier-platform-core';
+
+const require = createRequire(import.meta.url);
```
kreddlear, line 6: Necessary because `import` is `undefined` for CommonJS apps, which we want to continue supporting.

packages/core/include/zapierwrapper.mjs
eliangcs, line 1: I'm still experimenting if there's a better implementation for `zapierwrapper.js`. And it's not easy! Sorry if I'm holding you up for too long.

Things I'd like to achieve:

* Keep the `.js` extension after copied into `build.zip` so the backend doesn't have to rename `zapierwrapper.js` to `zapierwrapper.mjs`
* Avoid resolving module path twice
* Use `import` rather than `require` because it'd be better to select the ESM entry point in the case of conditional exports like `{"exports": {"require": "./one.js", "import": "./two.js"}}`

While I’m still figuring it out, let me know if you have any ideas. Thanks!

packages/core/include/zapierwrapper.mjs
```diff
@@ -0,0 +1,22 @@
...
+let appPath;
+
+try {
+    appPath = require.resolve(process.cwd());
+} catch (e) {
+    // if it failed, try to resolve the app path with self-referencing
+    // in case index.js is not in the root directory
+    try {
+        appPath = require.resolve(path.basename(process.cwd()));
```
eliangcs, line 15: The package name should be taken from `package.json` rather than from the directory name.

packages/core/include/zapierwrapper.mjs
kreddlear: Thanks so much for digging into this ChangHung! I've been trying different options too, nothing solid so far. Thanks for listing out those preferred options, that's helpful to know.

The only new idea I had was to add another new file to `include`, something like `getEntryPoint.mjs`, and have `zapierwrapper.js` import from it - so we don't have to rename `zapierwrapper.js` but also are able to take advantage of ESM a bit more. I haven't gotten that to a point where it works well either, though.

eliangcs: @kreddlear take a look at #978, which I'm quite satisfied with. The idea is to split `zapierwrapper.js` into two versions: one for CommonJS, one for ESM.
- The content of `zapierwrapper.js` remains the same. It is and will continue to be used for integrations that use CommonJS.
- The new `zapierwrapper.mjs` file is specifically for integrations that use ESM.

If it looks good to you, can you merge it into your branch? We can keep iterating from there.

See below for implementation details.

-------

### `zapierwrapper.mjs`

The code for `zapierwrapper.mjs` is surprisingly simple:

```js
import zapier from 'zapier-platform-core';

// Import the integration's entry point by self-referencing its package name
// https://nodejs.org/docs/latest-v22.x/api/all.html#all_packages_self-referencing-a-package-using-its-name
const appRaw = (await import('<package_name_in_packageJson>')).default;
export const handler = zapier.createAppHandler(appRaw);
```

No need to resolve any paths here. Besides a string path, `zapier.createAppHandler()` already [accepts an "appRaw" object](https://github.com/zapier/zapier-platform/blob/dcb6fd87d350cb6f21a4cfa78734ac958c69b7a2/packages/core/src/tools/create-lambda-handler.js#L168). So we can take advantage of that, letting `zapierwrapper.mjs` do the import and pass it to `zapier.createAppHandler()`.

It's also tricky to import the app object within `zapier.createAppHandler()`. `zapier.createAppHandler()` in [defined](https://github.com/zapier/zapier-platform/blob/dcb6fd87d350cb6f21a4cfa78734ac958c69b7a2/packages/core/src/tools/create-lambda-handler.js#L188) in zapier-platform-core. During the build phase, zapier-platform-core is placed in `<integrationRoot>/node_modules/zapier-platform-core`. It's tricky for anything in `node_modules` to use dynamic `import()` to import anything in `<integrationRoot>`. It was possible for `require()`, but not easy for dynamic `import()`, since dynamic `import()` doesn't support "[folders as modules](https://nodejs.org/docs/latest-v22.x/api/modules.html#folders-as-modules)."

### On the naming of `zapierwrapper.js`

I tried renaming `zapierwrapper.js` to `zapierwrapper.cjs` to make it clearer that it's for CommonJS. But then I found out, by keep the `.js` extension, we can make it backward compatible -- zapier-platform-cli v17 can continue to work with previous versions of zapier-platform-core, and vice versa.

kreddlear: @eliangcs that looks great, very nice solution! This resolves the main tricky point of backwards compatibility with CommonJS apps while also supporting ESM _and_ index.js not being required in the root. And also means the actual content of `zapierwrapper.js` doesn't have to be very complex since it only has to support either CJS or ESM at a time 👏 

I'll merge your PR into this one - we might actually be good to go from there.

eliangcs:
@kreddlear I found there's an issue with `zapierwrapper.mjs`, which can lead to broken `build.zip`. See #982 for details. That PR also has some other improvements. Let me know what you think. And if it looks good to me, can you merge it to your branch again? Thanks!

eliangcs (approved):
I tested the `zapier build` command using several Zapier-owned CLI integrations. The generated `build.zip` files are consistent with those produced by the current CLI v16—everything looks great!

I also did a quick test on building ESM integrations, and it worked. So I think this is ready to merge. After merging, we probably want to ask @tkcranny to run internal tests.

Suggestions:
- Since you're already adding two ESM example apps in #976, maybe we can remove example-app/esmodule here?
- I added an AI-generated PR description and title. I think they're quite accurate. Let me know if they look good to you too.

kreddlear:
@eliangcs thanks so much for the help with getting this PR past the finish line! I think the AI description is accurate and helpful 👍 Removing the `esmodule` example app makes sense to me since I'll be adding the other ones in the `init` PR. I've removed that and the associated test, and will merge this into the release branch.

PR #968, opened 2025-02-10
PDE-5772 feat(cli): add `--authentication-id` flag to `invoke` command

eliangcs:
Adds a new command flag `--authentication-id` (`-a` for short) to the `zapier invoke` command.

When `--authentication-id` is specified, `zapier invoke` still runs the code locally but delegates any outgoing requests to the Zapier relay endpoint (`POST /api/platform/cli/apps/{app_id}/relay`), which sends authenticated requests to the partner API using production auth data.

### How it works

Node.js's [`http.request`](https://nodejs.org/docs/latest-v22.x/api/http.html#httprequestoptions-callback) and [`https.request`](https://nodejs.org/docs/latest-v22.x/api/https.html#httpsrequestoptions-callback) are patched so that whenever they are called, the request is sent to the Zapier relay endpoint instead of the original target server. The relay endpoint then injects the required auth data and forwards the request to the intended URL. Throughout the process, the integration code has no knowledge of the auth credentials, but can still send authenticated requests to the partner API via the Zapier relay.

Besides `http(s).request`, this PR also patches the native `fetch()` API, since some integrations use the native `fetch()` rather than `z.request()` (which relies on `http(s).request` under the hood).

### Handling `{{subdomain}}` in URLs

Before invoking the `operation.perform` method, `zapier invoke` populates `bundle.authData` so that each auth field `k` is set as `bundle.authData[k] = '{{k}}'`. This renders expressions like `{{bundle.authData.subdomain}}.example.com` as `{{subdomain}}.example.com` before the request is sent.

However, node-fetch [validates](https://github.com/node-fetch/node-fetch/blob/9b9d45881e5ca68757077726b3c0ecf8fdca1f29/src/request.js#L36) URLs. And `{` and `}` are not valid characters used in a hostname. To bypass the "[Only absolute URLs are supported](https://github.com/node-fetch/node-fetch/blob/9b9d45881e5ca68757077726b3c0ecf8fdca1f29/src/request.js#L216)" error, this PR injects a `beforeRequest` middleware to the app definition that replaces `{` and `}` with `lcurly-` and `-rcurly`, respectively. This allows node-fetch to parse `lcurly-subdomain-rcurly.example.com` instead of failing on `{{subdomain}}.example.com`.

In the patched `http(s).request`, `lcury-` and `-rcurly` is replaced back with `{` and `}` before it hands over the request to the Zapier relay endpoint.

To keep this encoding/decoding invisible to the integration code, an `afterResponse` middleware is also injected to the app definition to reverse the substitution in the response object.


packages/cli/docs/cli.md
```diff
@@ -379,6 +387,7 @@ The following is a non-exhaustive list of current limitations and may be support
...
 * `--local-port` | Only used by `auth start` subcommand. The local port that will be used to start the local HTTP server to listen for the OAuth2 callback. This port can be different from the one in the redirect URI if you have port forwarding set up.  Defaults to `9000`.
+* `-a, --authentication-id` | EXPERIMENTAL: Instead of using the local .env file, use the production authentication data with the given authentication ID (aka the "app connection" on Zapier). Find them at https://zapier.com/app/connections or specify '-' to interactively select one from your available authentications. When specified, the code will be still run locally, but all outgoing requests will be proxied through Zapier with the production auth data.
```
kola-er, line 390:
```suggestion
* `-a, --authentication-id` | EXPERIMENTAL: Instead of using the local .env file, use the production authentication data with the given authentication ID (aka the "app connection" on Zapier). Find them at https://zapier.com/app/connections or specify '-' to interactively select one from your available authentications. When specified, the code will still run locally, but all outgoing requests will be proxied through Zapier with the production auth data.
```

packages/cli/src/utils/local.js
```diff
@@ -2,8 +2,231 @@ const _ = require('lodash');
...
+    if (isAlreadyRelay) {
+      // Just call the original fetch function; do *not* re-relay
+      return fetchFunc(originalUrl, originalOptions);
+    }
+
+    // Combine the relay's pathname with the "host + path" from the original
+    // For example: relayUrl = http://my-relay.test
+    //   => parsedRelayUrl.pathname might be '/'
+    //   => combinedPath = '/example.com:8080/some/path'
+    const combinedPath = `${parsedRelayUrl.pathname}/${host}${parsedOriginalUrl.pathname}`;
```
kola-er, line 200:
```suggestion
    const combinedPath = `${parsedRelayUrl.pathname}/${host}${parsedOriginalUrl.pathname}`.replaceAll('lcurly-', '{{').replaceAll('-rcurly', '}}');
```

Wondering if this is necessary just like it is done here https://github.com/zapier/zapier-platform/pull/968/files#diff-7640b561b92d724ea2ef660a2e675d5f917e95b5ebde5490430cfc2fdddc46e9R104-R106 for `wrapHttpRequestFuncWithRelay`

kola-er: Argh, I can see you have explained why it is not necessary here as `node-fetch` throws error on `{` and `}`.

kola-er (commented):
This should speed up the testing/debug process for builders. Thanks for this big feature. Left a comment on typo. Happy to approve after.

packages/cli/docs/cli.md
eliangcs: Fixed in [63c688a](https://github.com/zapier/zapier-platform/pull/968/commits/63c688ac7e0001bea94b46c7d59aed34832a2649).

eliangcs:
@kola-er thanks for taking a look! I fixed the typo and also added a few command usage examples. Can you re-review? Thanks!

kola-er (approved):
Would be great to have some unit tests for it if not complicated to test. Forgot to point it out in my initial review. However, it is fine to go ahead with this.

eliangcs:
@kola-er thanks! Since this is an experimental feature and we have smoke tests to verify existing functionalities, I'll skip adding tests for this PR. Merging.

PR #969, opened 2025-02-13
fix(cli) - Fix invoke auth label 

standielpls:
For string `connectionLabel`, running `zapier invoke auth test` prints the output of the `Authentication.test` results. 
For `zapier invoke auth label`, this returns an error: `Error: bundle is not defined`. While the two are similar, in that calls `testAuth`, the `label` option has additional template rendering where it looks for `bundle` and the `testResult` value does not. 
To fix, wrap the data in a bundle object.

### DIFF
--- a/packages/cli/src/oclif/commands/invoke.js
+++ b/packages/cli/src/oclif/commands/invoke.js
@@ -259,7 +259,7 @@ const getAuthLabel = async (labelTemplate, authData, meta, zcacheTestObj) => {
   const testResult = await testAuth(authData, meta, zcacheTestObj);
   labelTemplate = labelTemplate.replace('__', '.');
   const tpl = _.template(labelTemplate, { interpolate: /{{([\s\S]+?)}}/g });
-  return tpl({ bundle: { authData, inputData: testResult } });
+  return tpl({ ...testResult, bundle: { authData, inputData: testResult } });
 };
 
 const getLabelForDynamicDropdown = (obj, preferredKey, fallbackKey) => {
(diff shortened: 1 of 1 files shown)


packages/cli/src/oclif/commands/invoke.js
```diff
@@ -259,7 +259,7 @@ const getAuthLabel = async (labelTemplate, authData, meta, zcacheTestObj) => {
   const testResult = await testAuth(authData, meta, zcacheTestObj);
   labelTemplate = labelTemplate.replace('__', '.');
   const tpl = _.template(labelTemplate, { interpolate: /{{([\s\S]+?)}}/g });
-  return tpl(testResult);
+  return tpl({ bundle: { authData, inputData: testResult } });
```
Natay, line 262: out of curiosity, how come the testResult is added into inputData as oppose to outputData?

eliangcs: I'm about 80% sure. It's because 1) the test result is input for `connectionLabel` and 2) `outputData` was only introduced later for other method calls.

eliangcs, line 262: For `connectionLabel`, these both templates actually work and are equivalent:

- `{{username}}`
- `{{bundle.inputData.username}}`

This PR fixes the second case, but will break the first case. @standielpls can you fix it so that it works for both cases? Here's the corresponding [backend code](https://gitlab.com/zapier/zapier/-/blob/3c9b81f79824f8ca2869cf2957e5331e304ec63b/web/backend/zapier/app_platform/engine/models/authentication.py#L456-457) we should match.

eliangcs (approved):

PR #972, opened 2025-02-17
BREAKING CHANGE(schema): split authentication.fields into inputFieds and outputFields (PDE-5589 )

Natay:
Fixes the issue: authentication.fields are a combination of input fields (e.g. subdomain or API key - which the user needs to enter), and output fields (computed: true - which we expect authentication procedure to return). [Coda](https://coda.io/d/Edge-Team_dCaLL2dK8Zu/Auth-Fields_su9fjyvI#_luF05mor), [Slack](https://zapier.slack.com/archives/C5Z9BP4U9/p1733307242418659?thread_ts=1733255865.570059&cid=C5Z9BP4U9)


example-apps/basic-auth/authentication.js
```diff
@@ -32,7 +32,8 @@ module.exports = {
 
     // Define any input app's auth requires here. The user will be prompted to enter
     // this info when they connect their account.
-    fields: [],
+    inputFields: [],
```
Natay, line 35: Bit of a weird scenario,
-  Leaving `fields` as is  results [in this error](https://github.com/zapier/zapier-platform/actions/runs/13375926420/job/37355062393)
```smoke tests - setup will take some time
    ✔ package size should not change much (70764 -> 70508 bytes) (248ms)
    basic-auth
Validating project locally
┌────────────────────┬─────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────┐
│ Property           │ Message                                                         │ Links                                                                   │
├────────────────────┼─────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────┤
│ App.authentication │ additionalProperty "fields" exists in instance when not allowed │ https://platform.zapier.com/cli_docs/schema#authenticationschema@[16](https://github.com/zapier/zapier-platform/actions/runs/13375926420/job/37355062393#step:5:17).3.0 │
```

- But updating to `inputFields` causes [this error](https://github.com/zapier/zapier-platform/actions/runs/13376674259/job/37357302905)
```
 ›   Error: "https://zapier.com/api/platform/cli/check" returned "400" saying 
 ›   "App must validate against schema before running app checks"
```
Going to reach out to team on best approach here, I think I may need to leave the "fields" as is and update it's description to "DEPRECATED"?

Natay: I'll dig up some older PRs to see how fields are deprecated and do the same here

Natay: Okay, 
- Marked "fields" as deprecated following this [PR](https://github.com/zapier/zapier-platform/pull/644/files#diff-6f1e7532bff63046d0592fdbcf21951588b35ee94db04d662c1feec33c8c64d2)
- [Created a ticket](https://zapierorg.atlassian.net/browse/PDE-5769) to update `example-apps` after this MR, from testing I am still getting we `Error: "https://zapier.com/api/platform/cli/check" returned "400"` if they are updated in this PR. I think this is because the major version needs to be published first but that's a guess.

packages/schema/docs/build/schema.md
```diff
@@ -414,7 +414,9 @@ Key | Required | Type | Description
 --- | -------- | ---- | -----------
 `type` | **yes** | `string` in (`'basic'`, `'custom'`, `'digest'`, `'oauth1'`, `'oauth2'`, `'session'`) | Choose which scheme you want to use.
 `test` | **yes** | oneOf([/RequestSchema](#requestschema), [/FunctionSchema](#functionschema)) | A function or request that confirms the authentication is working.
-`fields` | no | [/AuthFieldsSchema](#authfieldsschema) | Fields you can request from the user before they connect your app to Zapier.
+`fields` | no | [/AuthFieldsSchema](#authfieldsschema) | **DEPRECATED** fields requested from the user before they connect the app.
```
rnegron, line 417:
```suggestion
`fields` | no | [/AuthFieldsSchema](#authfieldsschema) | **DEPRECATED: use inputFields/outputFields instead** fields requested from the user before they connect the app.
```

kola-er:
@Natay Good start. Reading the requirement again, it seems the auth `inputFields` and `outputFields` have some non-common properties as below:

- For the inputFields
   - helpText
   - placeholder
   - choices
   - inputFormat
- For the outputFields
   - isNoSecret
   - computed (however this would be redundant and can be removed)
   
Apologies for leading you on to use same for both, but we would need to break the `/AuthFieldSchema` into `/AuthInputFieldSchema` and `/AuthOutputFieldSchema`.

Also before this is merged, we need to update the backend to support both auth `inputFields` and `outputFields` in addition to the auth `fields`.

Natay:
@kola-er thank you!

Would it make sense to include `isNoSecret` for `inputFields` as well? Or do we expect things like "api_key" to be in the output only?

Natay:
[Backend MR](https://gitlab.com/zapier/zapier/-/merge_requests/61553) to support inputFields and outputFields

packages/cli/src/oclif/commands/invoke.js
```diff
@@ -519,7 +516,7 @@ class InvokeCommand extends BaseCommand {
       );
     }
     const authData = await this.promptForAuthFields(
-      appDefinition.authentication.fields,
+      appDefinition.authentication.inputFields,
```
eliangcs, line 519: `zapier invoke` is still gaining adoption. It would hurt if it starts breaking. So I'd like the `invoke` command to continue working on previous versions of platform-core for at least some time. Maybe we could do something like:

```suggestion
      appDefinition.authentication.fields || appDefinition.authentication.inputFields,
```

And add this back at line 336-338:

```js
      if (field.computed) {
        continue;
      }
```

PR #973, opened 2025-02-18
fix(schema): Reference the proper FieldSchema as necessary

kola-er:
With the /FieldSchema now hidden, its references are displayed `undefined` in https://github.com/zapier/zapier-platform/blob/f59896291048bfe858e967689eabe86c58133c35/packages/schema/docs/build/schema.md#inputfieldschema


Natay (approved):
🙏

PR #974, opened 2025-02-18
PDE-5774 fix(core): Catch HTTP 500 along with status codes >500 in RPC client

rnegron:
We see `Got id undefined but expected ID_NUMBER when calling RPC` errors when it comes to RPC endpoints returning a 500 response. Inspecting the code, I believe we should likely catch the error before we check `res.content.id`.

After this change is released, we should update our log stream alarms to ALSO consider the `Unable to reach the RPC server` text instead of just the `Got id ${res.content.id} but expected ${id} when calling RPC` text. Both texts should be viable since we could have apps still running on previous versions.

### DIFF
--- a/packages/core/src/tools/create-rpc-client.js
+++ b/packages/core/src/tools/create-rpc-client.js
@@ -108,7 +108,7 @@ const createRpcClient = (event) => {
       try {
         res = await request(req);
 
-        if (res.status > 500) {
+        if (res.status >= 500) {
           throw new Error('Unable to reach the RPC server');
         }
         if (res.content) {
--- a/packages/core/test/tools/mocky.js
+++ b/packages/core/test/tools/mocky.js
@@ -51,10 +51,10 @@ const mockRpcGetPresignedPostCall = (key) => {
     });
 };
 
-const mockRpcFail = (error) => {
+const mockRpcFail = (error, status = 400) => {
   nock(FAKE_ZAPIER_URL)
     .post(RPC_URL_PATH)
-    .reply(400, (uri, requestBody) => {
+    .reply(status, (uri, requestBody) => {
       const id = JSON.parse(requestBody).id;
       return { error, id };
     });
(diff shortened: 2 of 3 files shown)


rnegron:
Following [this PR](https://github.com/zapier/zapier-platform/pull/875), we should keep the retry logic for 5xx errors.

packages/core/src/tools/create-rpc-client.js
```diff
@@ -108,7 +108,7 @@ const createRpcClient = (event) => {
       try {
         res = await request(req);
 
-        if (res.status > 500) {
+        if (res.status >= 500) {
```
rnegron, line 111: HTTP 500 is now a non-standard response code (note this PR is targeting a major release branch as this could be a breaking change)

packages/core/test/tools/mocky.js
```diff
@@ -51,10 +51,10 @@ const mockRpcGetPresignedPostCall = (key) => {
     });
 };
 
-const mockRpcFail = (error) => {
+const mockRpcFail = (error, status = 400) => {
```
rnegron, line 54: HTTP 400 is the default error code for manual failures in the RPC API now, so I've set it as the default for this test

packages/core/test/tools/rpc-client.js
```diff
@@ -16,17 +16,17 @@ describe('rpc client', () => {
 
   it('should handle an explosion', () => {
     // mock 3 explosions due to retry
-    mocky.mockRpcFail('this is an expected explosion');
-    mocky.mockRpcFail('this is an expected explosion');
-    mocky.mockRpcFail('this is an expected explosion');
+    mocky.mockRpcFail('this is an expected explosion', 500);
```
rnegron, line 19: The `explode` RPC method actually throws an HTTP 500 via an exception being raised 😅

Natay (approved):
🙏

PR #976, opened 2025-02-26
feat(cli): Add ESM support to `zapier init` command for Minimal and Typescript templates

kreddlear:


packages/cli/src/generators/templates/typescript-esm/index.js
```diff
@@ -0,0 +1,2 @@
+// TODO likely should remove this once we support non-root index.js
```
eliangcs, line 1: We _do_ support non-root `index.js`, and this root `index.js` can be deleted, right?

eliangcs (commented):
Left some suggestions on the templates. Don't forget to regenerate the example-apps. Thanks!

CHANGELOG.md
```diff
@@ -135,7 +135,6 @@ Here are the detailed release notes for this release (**note that ❗ denotes a
     - Remove `@oclif/command`
     - Remove `@oclif/config`
     - Update `archiver` 5.3.1 to 7.0.1
-    - Update `gulp-prettier` from 4.0.0 to 5.0.0
```
kreddlear, line 138: This didn't actually make it into [the v16 PR](https://github.com/zapier/zapier-platform/pull/926), removing it to prevent future confusion

eslint.config.cjs
```diff
@@ -58,6 +58,20 @@ module.exports = [{
...
+}, {
+    files: ["example-apps/minimal-esm/*"],
+
+    languageOptions: {
+        ecmaVersion: 2025,
+        globals: {
+            ...globals.mocha,
+        },
+        sourceType: "module",
+    },
```
kreddlear, line 73: I had to add this to keep the precommit hook from failing when it runs eslint, since eslint support for `import` attributes (`with { type: "json" }`) was added recently in https://github.com/eslint/eslint/pull/19076. The `typescript-esm` example app did not trigger these errors for some reason.

packages/cli/src/generators/templates/index-esm.template.js
```diff
@@ -0,0 +1,36 @@
...
+<% } %>
+
+import { version as packageVersion } from './package.json';
+import { version as platformVersion } from '<%= corePackageName %>';
+
+export default {
+    // This is just shorthand to reference the installed dependencies you have.
+    // Zapier will need to know these before we can upload.
+    version: packageVersion,
+    platformVersion: platformVersion,
```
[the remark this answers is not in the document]
kreddlear: Thank you for catching this, apologies for missing it! I've made sure to run `zapier validate` on the apps generated by `zapier init` this time before pushing.

packages/cli/src/generators/templates/typescript-esm/index.js
kreddlear: Yes, thanks for calling this out! I've removed it from the template and the example app ✅

kreddlear:
Thank you for the detailed feedback, @eliangcs ! I think I've implemented all your feedback and it should be ready for another review if you have time today. Thanks in advance!

eliangcs:
@kreddlear thanks! I didn't have enough time to do a full review/test today. I'll keep reviewing my tomorrow. See the following suggestions I have so far.

1. The leftover `zapierwrapper.js` should be cleaned up. To reproduce, run this in sequence:

```
zapier-dev init mini-esm -t minimal -m esm
cd mini-esm
yarn install
yarn link zapier-platform-core
zapier-dev validate
```

Then `zapierwrapper.js` file will be left behind in the app directory. It should be cleaned up.

2. It seems we have a lot of duplicate files with the `typescript` template. With TypeScript, `-m commonjs` and `-m esm` should be able to share most of the `.ts` files. The only difference is that when transpiled, they are converted to CommonJS and ESM, respectively.

kreddlear:
Thanks @eliangcs , no rush! I consolidated the Typescript templates so they just use one template directory now. I had to make some small changes (mostly adding file extensions) to the `src` files to make sure they work for both CJS and ESM, but nothing major.

Re: the `zapierwrapper.js` file not getting cleaned up, I noticed `shouldDeleteWrapper` defaulted to `false` in local.js per your original PR adding those zapierwrapper utils - is there any reason it shouldn't default to `true` instead? That would achieve the cleanup results. I tested it with the commands that use `localAppCommand` and they all seemed to work properly, so hopefully that simple change will do the trick.

eliangcs (approved):
I tested the templates and they all work nicely. Thanks for all the work!

See my final suggestions in #986. If it looks good to you, merge it into your branch. Then feel free to ask someone else to approve this PR and merge it.

kreddlear:
Thank you for the followup PR, @eliangcs ! It looks great, I've merged it in. I'd been trying to figure out why those `typescript` and other smoke tests were hanging/not finishing when run locally, it is great to know it wasn't something wrong with my own local setup.

I'll go ahead and merge this into the v17 branch, it looks like I actually don't need another approval for that 👍

PR #979, opened 2025-03-05
feat( schema ): update dynamic type to include functions and requests

Natay:
This PR updates the schema so we can use functions/requests to power dynamic dropdowns 

## Before / Current Behavior

Current behavior has "dynamic" fields being a reference to another trigger.

```
module.exports = {
  key: 'people',
  noun: 'person',

  operation: {
    inputFields: [
      {
        key: 'species_id',
        dynamic: 'refrence.to.trigger',
      },
    ],
    perform,
  },
};
```

## After / New Behavior

With this change ( plus schema change PR ), apps can now define a "dynamic" field to be functions , in addition to the using previous `dynamic: 'refrence.to.trigger'`

```
const dynamic = async (z, bundle) => {
  const request = {
    url: 'https://swapi.info/api/species/',
    params: {},
  };
  const response = await z.request(request);
  return response
};


module.exports = {
  key: 'people',
  noun: 'person',
  operation: {
    inputFields: [
      {
        key: 'species_id',
        dynamic,
      },
    ],
    perform,
  },
};
```


eliangcs:
@Natay quick question: would `dynamic: 'refrence.to.trigger'` still work? If so, this isn't a breaking change then?

Natay:
@eliangcs Yeah it would still work! I added the "breaking change" out of paranoia/safety because it is a schema change. We would still need to merge this into v17 branch because it builds off of a previous breaking schema change

packages/core/types/zapier.generated.d.ts
```diff
@@ -715,13 +715,14 @@ export interface InputField {
...
+   * dynamic dropdown.
    */
-  dynamic?: RefResource;
+  dynamic?: RefResource | Request | Function;
 
   /**
-   * A reference to a search that will guide the user to add a search
-   * step to populate this field when creating a Zap.
+   * Reference a resource by key and the data it returns. In the
+   * format of: `{resource_key}.{foreign_key}(.{human_label_key})`.
```
kola-er, line 725: 🤔 how come this was generated when it wasn't changed in the InputFieldSchema.js?

packages/schema/lib/schemas/InputFieldSchema.js
```diff
@@ -30,8 +32,12 @@ module.exports = makeSchema(
...
       dynamic: {
         description:
-          'A reference to a trigger that will power a dynamic dropdown.',
-        $ref: RefResourceSchema.id,
+          'A reference to a trigger, request, or function that will power a dynamic dropdown.',
+        oneOf: [
+          { $ref: RefResourceSchema.id },
+          { $ref: RequestSchema.id },
+          { $ref: FunctionSchema.id },
+        ],
```
kola-er, line 40: Initially wanted to suggest we add usage examples for the 3 options, but it seems a more detailed examples in https://docs.zapier.com/platform/home would do a better job.

packages/core/types/zapier.generated.d.ts
Natay: Good question 🤔 , not sure tbh, I'll regenerate and keep an eye!

Natay: Ah just noticed it's the description for [`RefResourceSchema` ](https://github.com/zapier/zapier-platform/blob/63bf8bbb4841048e6a32ebf825169f1a0fed3741/packages/schema/lib/schemas/RefResourceSchema.js#L8), I think it got inserted because of the Oneof introduced  ( not sure why ) but it does seem correct 👍

kola-er: Sorry, can we manually revert just this change and run the commands once more.

Natay: Good catch, did not see that! Tried regenerating and had the same effect, removed `RefResourceSchema` from the `oneof` in dynamic and it stops overwriting the description of the `search`

Natay: Confirmed that is the case, https://cdn.zappy.app/effc64a6336c0c498d017d6ab9f6605f.mp4

Having a `oneOf` goes around the root cause, but it seems like a bug in the .ts generator 
```
      search: {
        description: "..."
        oneOf: [{$ref: RefResourceSchema.id}]
      },
```

Natay: it happens on fresh branches too, blocked some time to dig into why

Natay: Okay, took me a bit to understand how the typescript generation works 😅, here is what I think might be going on:

1. The issue seems to stem from how the schema compilation and dereferencing works in the `precompile.ts` file. When it processes schemas with `$ref` references, [it uses the dereference function from json-schema-to-typescript](https://github.com/zapier/zapier-platform/blob/abf5b415636ca3d14b1c16fd463dc62fdc03ead2/schema-to-ts/src/precompile.ts#L108).
2. Looking at `InputFieldSchema.js`, we have two fields that reference `RefResourceSchema` twice ( apparently the first time from what I can tell ):
```
dynamic: {
  description: '...',
  oneOf: [
    { $ref: RefResourceSchema.id },  // First reference
    { $ref: RequestSchema.id },
    { $ref: FunctionSchema.id },
  ],
},
search: {
  description: '...',
  $ref: RefResourceSchema.id,  // Second reference
}
```
3. When `RefResourceSchema` is referenced in both `dynamic` and `search`, the dereferencer is likely merging the schema properties incorrectly, causing the description from `RefResourceSchema` to override the local description in the `search` field.
4. This behavior is particularly odd because it only happens when both references exist - when we remove the `RefResourceSchema` reference from the `dynamic` field, the local description for `search` is preserved correctly.

5. Similarly I was able to get it working by creating a separate schema for the `search` field that extends `RefResourceSchema`: 

```
'use strict';

const makeSchema = require('../utils/makeSchema');
const RefResourceSchema = require('./RefResourceSchema');

// Copy everything exactly from RefResourceSchema, only changing id and description
module.exports = makeSchema({
  id: '/SearchRefSchema',
  description:
    'A reference to a search that will guide the user to add a search step to populate this field when creating a Zap.',
  type: RefResourceSchema.schema.type,
  pattern: RefResourceSchema.schema.pattern,
  examples: RefResourceSchema.schema.examples,
  antiExamples: RefResourceSchema.schema.antiExamples
});
```

6. Weirdly enough, I get the same issue when creating a new schema that directly inherits from `RefResourceSchema`.
```
const SearchRefSchema = makeSchema({
  id: '/SearchRefSchema',
  ...RefResourceSchema.schema,
  description: 'A reference to a search that will guide the user to add a search step to populate this field when creating a Zap.'
});
```

I think the above solution would suffice for this PR but the bug appears to be in how the [dereferencing process](https://github.com/zapier/zapier-platform/blob/abf5b415636ca3d14b1c16fd463dc62fdc03ead2/schema-to-ts/src/precompile.ts#L108) in `compileToAST` handles multiple references to the same schema. 
I am going to create a ticket outside this PR to modify the schema compilation process in to preserve local descriptions when dereferencing. We would need to modify the `compileToAST` function to handle description merging differently ( I am not 100% sure how to do that to be honest )

kola-er: @Natay Odd, I just tried moving the `search` above the `dynamic` and got it the zapier.generated.d.ts file not touching the the search's description.



Please, give the reordering a trial.

Natay: Oh wow 🤦  lol I had serious tunnel vision and didn't even think of that, thank you so much!

kola-er (approved):

PR #980, opened 2025-03-07
BREAKING CHANGE(core): Remove Bluebird, replace with native promises

kreddlear:
### Summary

This PR removes the dependency on the Bluebird package and modernizes async handling across the codebase by switching to native Promises and async/await.

### Changes

- Removed usage of the Bluebird package in favor of native Promises - the Bluebird library is no longer actively maintained nor recommended, and its presence is causing issues for the Integrations teams, especially with Typescript
    - This also means that `Promise` is no longer globally patched with Bluebird Promises - instead `Promise` refers to a native Promise as would normally be expected.
- Replaced Bluebird error contextifying with a `whatHappened` array of strings stored in the `_zapier` object, to ensure error context is still bubbled up as expected
- Added `enrichErrorMessages` method to middleware to enable above error context
- Updated Lambda handler to use async/await instead of callbacks
- Refactored middleware functions to use async/await for improved readability and maintainability

### Motivation

Simplifies the async flow, reduces external dependencies, and aligns the codebase with modern JavaScript practices.


packages/core/src/tools/create-lambda-handler.js
```diff
@@ -187,22 +177,7 @@ const loadApp = async (event, rpc, appRawOrPath) => {
...
-  const handler = (event, context, callback) => {
-    // Wait for all async events to complete before callback returns.
-    // This is not strictly necessary since this is the default now when
-    // using the callback; just putting it here to be explicit.
-    // In some cases, the code hangs and never exits because the event loop is not
-    // empty, so we can override the default behavior and exit after the app is done.
-    context.callbackWaitsForEmptyEventLoop = true;
-    if (event.skipWaitForAsync === true) {
-      context.callbackWaitsForEmptyEventLoop = false;
-    }
```
kreddlear, line 199: Removing this since we're not using callbacks anymore, so it likely will not apply. Need to remove the corresponding z/z code and Gargoyle switch as well - [context here](https://gitlab.com/zapier/zapier/-/merge_requests/59754), [AWS docs here](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html#nodejs-handler-async)

packages/core/test/tools/promise.js
kreddlear, line 1: Since we are deleting the `core/src/tools/promise.js` file above, which mostly powers this test file, do we want to delete it as well?

The only remaining failing tests are these 3 `contextual promise` tests. I have not added the ability to "contextify" via a function in this PR, rather context is just handled by pushing up strings to the `whatHappened` array in `_zapier`, as originally suggested in [the spike PR here](). I wasn't able to find a strong reason to have this functionality but am open to adding it if it's necessary.

kreddlear:
@eliangcs 👋 Hello, thanks again for all your help so far! This MR is almost ready for review, I'll spend time tomorrow canarying it using a test platform version to make sure it is fully ready. In the meantime though, I'm wondering if you could take a look at [this question](https://github.com/zapier/zapier-platform/pull/980/files/fa91781a544550889a9269cb5ce820673edef6b0#diff-b04576a318a41fec7dc3242b29c30b5ca500b392d1f8b73048d4ca0c566f08b2) and the failing tests [here](https://github.com/zapier/zapier-platform/actions/runs/14232828367/job/39886707180?pr=980)?

My basic question is if you think it's worth refactoring to keep the logic to pass a function in to "contextify" results, or if we should just remove the test file for `ZapierPromise` along with the `promise.js` logic itself. I'd appreciate any thoughts, thank you in advance! And of course if there are any other red flags with the PR that stand out right away, please feel free to bring them to my attention as well, it should be mostly done.

packages/core/test/tools/promise.js
eliangcs: I don't think there's a reason to keep the "contextify" functionality either. So yeah, let's remove the test file.

packages/core/src/middleware.js
```diff
@@ -55,47 +65,46 @@ const applyMiddleware = (befores, afters, app, options) => {
...
+    const beforeMiddleware = async (beforeInput) => {
+      let newInput = beforeInput;
+      for (const func of befores) {
+        const args = [newInput].concat(options.extraArgs);
+        const maybePromise = func.apply(undefined, args);
+        // legacy scripting runner returns a Promise for beforeRequest
+        if (typeof maybePromise !== 'object') {
+          throw new Error('Middleware should return an object.');
+        }
+        newInput = await Promise.resolve(maybePromise);
```
kreddlear, line 77: I added this promise logic here since the legacy scripting `beforeRequest` logic can now return a promise.

packages/core/integration-test/integration-test.js
```diff
@@ -52,16 +52,14 @@ runLambda.testName = 'runLambda';
 const runLocally = (event) => {
   return new Promise((resolve, reject) => {
     const handler = createLambdaHandler(
-      path.resolve(__dirname, '../test/userapp/index.js'),
```
kreddlear, line 55: This should have been removed in the previous ESM PR.

eliangcs (commented):
Left a few comments. I don't see big issues.

I'd like to test it further using `zapier invoke`, but it didn't work because the previous changes with ESM didn't work well with the `zapier invoke` implementation in the `main` branch. I'll solve merge conflict between this and `main` (in a separate PR) and continue my test.

eliangcs:
I found some issues while working on #997.

1. `bluebird` is still in `packages/core/package.json`. We should remove it.

2. It's possible for a developer to use zapier-platform-cli v17 on zapier-platform-core v16. I tried to run `zapier validate` and `zapier describe` on an example integration that uses core v16, and saw an error saying `Cannot set properties of undefined (setting 'callbackWaitsForEmptyEventLoop')`:

```
$ zapier-dev validate --debug
  oclif:zapier:validate argv is [] +0ms
  oclif:zapier:validate args are {} +1ms
  oclif:zapier:validate flags are { debug: true, format: 'table' } +0ms
  oclif:zapier:validate ------------ +0ms
Validating project locally
  oclif:zapier:validate TypeError: Cannot set properties of undefined (setting 'callbackWaitsForEmptyEventLoop')
  oclif:zapier:validate     at handler (/Users/eliang/Projects/zapps/todoist/node_modules/zapier-platform-core/src/tools/create-lambda-handler.js:195:44)
  oclif:zapier:validate     at /Users/eliang/Projects/zapier-platform/packages/cli/src/utils/local.js:343:18
  oclif:zapier:validate     at localAppCommand (/Users/eliang/Projects/zapier-platform/packages/cli/src/utils/local.js:358:26)
  oclif:zapier:validate     at async ValidateCommand.perform (/Users/eliang/Projects/zapier-platform/packages/cli/src/oclif/commands/validate.js:14:20)
  oclif:zapier:validate     at async Promise.all (index 1)
  oclif:zapier:validate     at async ValidateCommand._run (/Users/eliang/Projects/zapier-platform/node_modules/@oclif/core/lib/command.js:312:22)
  oclif:zapier:validate     at async Config.runCommand (/Users/eliang/Projects/zapier-platform/node_modules/@oclif/core/lib/config/config.js:435:25)
  oclif:zapier:validate     at async run (/Users/eliang/Projects/zapier-platform/node_modules/@oclif/core/lib/main.js:95:16)
  oclif:zapier:validate     at async /Users/eliang/Projects/zapier-platform/packages/cli/src/bin/run:5:3 +67ms
 ›   Error: Cannot set properties of undefined (setting 'callbackWaitsForEmptyEventLoop')
```

It's because the way `localAppCommand` calls the Lambda handler is now different. On v16 we call it [with a callback](https://github.com/zapier/zapier-platform/blob/43f0abbf20d08270716cd56904390e551cc162c4/packages/cli/src/utils/local.js#L309), whereas on v17 we call it [with await](https://github.com/zapier/zapier-platform/blob/f1a09b34e85c5656ced1aa3d8e550ee58dfcf712/packages/cli/src/utils/local.js#L42).

For a good DX, zapier-platform-cli v17 should work with zapier-platform-core v16. But if that's not possible, we should at least print a clear error message telling the developer to upgrade zapier-platform-core v17.

eliangcs:
Update: 2️⃣ mentioned above has been fixed in https://github.com/zapier/zapier-platform/pull/997/commits/1855605c06a593615ee66ace5604185ad7c24fd8, so don't worry about it in this PR.

packages/core/src/tools/create-lambda-handler.js
```diff
@@ -238,24 +203,34 @@ const createLambdaHandler = (appRawOrPath) => {
...
-      const logData = { err, log_type: 'error' };
-      logErrorAndCallbackOnce(logMsg, logData, err);
-    });
+    // TODO ideally we should not use a promise here
+    // also ideally we could remove domain since it's deprecated...
+    return new Promise((resolve, reject) => {
+      handlerDomain.on('error', async (err) => {
+        // This error handler is only called when someone (we or devs) missed
+        // catching an error in a callback. Errors thrown by promises should be
+        // already caught by the try-catch below.
```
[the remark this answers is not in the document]
eliangcs: Update: I verified it locally by writing a trigger `perform` method like this:

```js
  setTimeout(() => {
    throw new Error('Random error!');
  }, 1);
    url: 'https://auth-json-server.zapier-staging.com/recipes',
  });
  return response.data;
};
```

When running the trigger with `zapier invoke trigger recipe --debug`, I can see the "Uncaught error: ..." error message.

```
$ zapier-dev invoke trigger recipe --debug
... (omitted)
  zapier:invoke Uncaught error: Error: Random error!
  zapier:invoke Error: Random error!
  zapier:invoke     at Timeout._onTimeout (/Users/eliang/Projects/zapps/oauth/triggers/recipe.js:3:11)
  zapier:invoke     at listOnTimeout (node:internal/timers:581:17)
  zapier:invoke     at process.processTimers (node:internal/timers:519:7) +18ms
  zapier:invoke {
  zapier:invoke   err: Error: Random error!
  zapier:invoke       at Timeout._onTimeout (/Users/eliang/Projects/zapps/oauth/triggers/recipe.js:3:11)
  zapier:invoke       at listOnTimeout (node:internal/timers:581:17)
  zapier:invoke       at process.processTimers (node:internal/timers:519:7) {
  zapier:invoke     domainThrown: true
  zapier:invoke   },
  zapier:invoke   log_type: 'error',
  zapier:invoke   request_headers: undefined,
  zapier:invoke   response_headers: undefined
  zapier:invoke } +0ms
✖ Invoking triggers.recipe.operation.perform
  oclif:zapier:invoke Error: Random error!
  oclif:zapier:invoke     at Timeout._onTimeout (/Users/eliang/Projects/zapps/oauth/triggers/recipe.js:3:11)
  oclif:zapier:invoke     at listOnTimeout (node:internal/timers:581:17)
  oclif:zapier:invoke     at process.processTimers (node:internal/timers:519:7) +102ms
 ›   Error: Random error!
 ›
 ›   Console logs:
 ›
```

eliangcs:
Update: bluebird removed in [bcd28a6](https://github.com/zapier/zapier-platform/pull/997/commits/bcd28a6a021b066d8bf1b192632056d623e62b2b).

packages/core/src/tools/create-lambda-handler.js
kreddlear: > So it's as if context.callbackWaitsForEmptyEventLoop is now always false when we refactor code from callbacks to async/await. I assume that's fine because we don't rely on callbacks to finish anything in background. The only way to make sure this behavioral difference doesn't break existing integrations is canary testing, I think.

Yes, this is my impression as well. I _think_ it will be fine, but it's hard to know exactly how integrations will behave.

Are there any particular integrations/integration types you'd recommend to canary to make sure that this doesn't break certain functionality? I'll be canary testing integrations relying on the legacy scripting runner, and if you have ideas for other integration functionality to focus on, I'd be happy to test those cases as well.

kreddlear:
Thanks so much @eliangcs ! I resolved the conflict and added some test cases for the `Uncaught error` vs `Unhandled error` logging per your comment [here](https://github.com/zapier/zapier-platform/pull/980/#discussion_r2039018222) - I had also confirmed the behavior locally, but it seems good to have that confirmed in tests as well. I'll merge this into v17 and continue canary testing 👍

eliangcs (approved):
Thanks for adding the test!

PR #982, opened 2025-03-12
PDE-5244 fix(cli,core): make zapierwrapper.mjs statically analyzable, more test cases

eliangcs:
Summary:

* Make `zapierwrapper.mjs` statically analyzable
* Add unit tests for `zapierwrapper.js` and `zapierwrapper.mjs`
* Remove the hard-coded `index.js` from `getLocalAppHandler()`

### Make `zapierwrapper.mjs` statically analyzable

I found that `zapier build` produced bad `build.zip`. It could miss the entry point specified in the `exports` field in the integration's `package.json`.

Why? `zapierwrapper.mjs` previously passed a variable to `import()` that couldn't be [statically analyzed](https://esbuild.github.io/api/#non-analyzable-imports) by esbuild. This PR updates it to use a placeholder string (`{REPLACE_ME_PACKAGE_NAME}`) that gets replaced with the integration package name during the build process. This allows for proper static analysis by esbuild.

### Add unit tests for `zapierwrapper.js` and `zapierwrapper.mjs`

This PR adds test coverage for both `zapierwrapper.js` and `zapierwrapper.mjs`. The tests verify that the wrappers correctly handle different `package.json` configurations, including various export patterns (conditional exports, default exports, and named exports).

### Remove the hard-coded `index.js` from `getLocalAppHandler()`

The local app handler powers some CLI commands such as `zapier describe` and `zapier validate`. Previously, it assumed that `index.js` was always the entry point. This PR removes this assumption, respecting the `exports` field specified in `package.json`.

How? Before running the CLI command, we copy the corresponding `zapierwrapper` file to the integration directory. Then `getLocalAppHandler()` can just `import('./zapierwrapper.js')`, letting `zapierwrapper.js` find out the entry point of the integration. When the CLI command finishes, we delete `zapierwrapper.js` from the integration directory.

### DIFF
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -41,6 +41,7 @@ const checkMissingAppInfo = require('./check-missing-app-info');
 
 const { runCommand, isWindows, findCorePackageDir } = require('./misc');
 const { respectGitIgnore } = require('./ignore');
+const { localAppCommand } = require('./local');
 
 const debug = require('debug')('zapier:build');
 
@@ -157,7 +158,10 @@ const writeZipFromPaths = (dir, zipPath, paths) => {
 };
 
 const makeZip = async (dir, zipPath, disableDependencyDetection) => {
-  const entryPoints = [path.resolve(dir, 'zapierwrapper.js')];
+  const entryPoints = [
+    path.resolve(dir, 'zapierwrapper.js'),
+    path.resolve(dir, 'index.js'), // For CommonJS integration
+  ];
 
   let paths;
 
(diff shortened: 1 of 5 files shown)


kreddlear (approved):
Thank you for identifying and solving this issue, ChangHung, and also for the detailed PR description which made it much easier to understand these changes. I like the solution of splitting out the zapierwrapper-related code into separate `utils` - I was struggling to figure out how `local.js` could respect the entry point without the hardcoded `index.js`, so I think reusing that `copyZapierWrapper()` is a clean way of handling it (or, as clean as this code can get 😵 ).

I'll merge this into my PR shortly!

PR #987, opened 2025-04-02
feat(cli) - Add new zapier legacy command

standielpls:
`legacy` is a new lifecycle state for app versions we want to support.
Accompanying documentation can be found [here](https://gitlab.com/zapier/public-api/public-api-docs/-/merge_requests/56)

### DIFF
--- a/packages/cli/src/oclif/commands/legacy.js
+++ b/packages/cli/src/oclif/commands/legacy.js
@@ -1,0 +1,67 @@
+const BaseCommand = require('../ZapierBaseCommand');
+const { Args, Flags } = require('@oclif/core');
+const { buildFlags } = require('../buildFlags');
+const colors = require('colors/safe');
+
+const { callAPI } = require('../../utils/api');
+
+class LegacyCommand extends BaseCommand {
+  async perform() {
+    const app = await this.getWritableApp();
+    const { version } = this.args;
+
+    this.log(
+      `${colors.yellow('Warning: .')}\n` +
+        `${colors.yellow('If all your changes are non-breaking, use `zapier migrate` instead to move users over to a newer version.')}\n`,
+    );
+
+    if (
+      !this.flags.force &&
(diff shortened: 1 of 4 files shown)


packages/cli/src/oclif/commands/versions.js
```diff
@@ -18,6 +18,7 @@ class VersionCommand extends BaseCommand {
         ['Platform', 'platform_version'],
         ['Users', 'user_count'],
         ['Deployment', 'deployment'],
+        ['Legacy Date', 'legacy_date'],
```
rnegron, line 21: 👍🏻

packages/cli/src/oclif/commands/legacy.js
```diff
@@ -0,0 +1,67 @@
...
+
+const { callAPI } = require('../../utils/api');
+
+class LegacyCommand extends BaseCommand {
+  async perform() {
+    const app = await this.getWritableApp();
+    const { version } = this.args;
+
+    this.log(
+      `${colors.yellow('Warning: .')}\n` +
```
rnegron, line 14: Is there an extra dot or is that on purpose for formatting?
```suggestion
      `${colors.yellow('Warning:')}\n` +
```

rnegron (dismissed):

packages/cli/src/oclif/commands/legacy.js
standielpls: Whoops - it was a typo. But also, let's omit the warning here (warning mentioned using zapier migrate!)

standielpls: But it led me to add another bit about migrating to legacy (and deprecated versions) [here](https://gitlab.com/zapier/public-api/public-api-docs/-/merge_requests/56/diffs?commit_id=d3fa78f394e426102c343b1bc64f369e5e7f0d2a)

rnegron (approved):

PR #989, opened 2025-04-04
PDE-5908 fix(cli): don't send authorization header if `zapier invoke` relay mode

eliangcs:
In `zapier invoke` relay mode, the `Authorization` header is automatically sent by Zapier relay. The CLI should remove it before sending the request to the relay endpoint.

This PR also improves the help doc.


packages/cli/docs/cli.md
```diff
@@ -339,28 +345,43 @@ authData_refresh_token='abcdefg'
 authData_account_name='zapier'
 ```
 
+
+### The `--authentication-id` flag (EXPERIMENTAL)
+
+Setting up local auth data can be troublesome. You'd have to configure your app server to allow localhost redirect URIs or use a port forwarding tool. This is sometimes not easy to get right.
+
+The `--authentication-id` flag (`-a` for short) gives you an alternative (and perhaps easier) way to supply your auth data. You can use `-a` to specify an existing production authentication/connection. The available authentication IDs can be found at https://zapier.com/app/connections.
```
rnegron, line 353: Hmm, I am actually not sure how to find authentication IDs  (like `123456`) from the connections page 😅 Should this say `available authentications` instead?

rnegron (dismissed):

packages/cli/docs/cli.md
eliangcs: Yeah, it's pretty obscure. You have to click the "N Zaps" and the ID will be in the URL as `connection={authentication_id}`:

https://github.com/user-attachments/assets/d2430ecd-df25-4395-9a77-986d129577c9

But I agree "available authentications" rather than "available authentication IDs" may be better.

docs/cli.md
eliangcs, line 1: This seems duplicate with `packages/cli/docs/cli.md`. I don't think think is used by anyone anymore, so I deleted it.

packages/cli/docs/cli.md
szchenghuang: Can we be specific about getting the IDs from the page?

szchenghuang (dismissed):

packages/cli/docs/cli.md
eliangcs: Created a [wiki page](https://github.com/zapier/zapier-platform/wiki/How-to-Find-Authentication-ID-on-zapier.com) with the instructions and added a link to it.

szchenghuang (approved):

PR #997, opened 2025-04-11
PDE-5261 chore: native promise with updates from main branch

eliangcs:
This PR merges the latest changes from the `main` branch into the `pde-5261-replace-bluebird` branch (#980). The `v17` branch has diverged from `main` for quite some time, resulting in many conflicts that need resolution in this PR. Notably, the recently introduced "relay mode" in the `zapier invoke` command has significantly altered `packages/cli/utils/local.js`, leading to considerable conflicts with the changes in the `v17` branch.

Additionally, this PR addresses an issue mentioned in [this comment](https://github.com/zapier/zapier-platform/pull/980#issuecomment-2801216473), where CLI v17 breaks on Core v16. The `localAppCommand` must adjust its call to the Lambda handler based on the function signature. If the handler accepts three arguments, `(event, context, callback)`, it indicates the integration is on Core v16 or older. Otherwise, it should be treated as an async function accepting a single argument, `(event, context = {})`.

Finally, I also removed `bluebird` from the dependencies.


eliangcs:
@kreddlear this is ready for review. If it looks good to you, can you help merge it into your branch? Thanks!

I'd like to remove the usage of `domain`, but this PR is already huge. Let's save it for another PR.

kreddlear:
Thanks for resolving those conflicts and merging the branch in, as well as fixing it up to work with v16 as well, @eliangcs ! This looks good to me, I'll merge it into my PR.

I feel similarly about `domain`, ideally we could just remove it soon in another PR!

PR #999, opened 2025-04-16
feat: V17 Typing Improvements – Bundle Inference, Define Helpers

tkcranny:
# Typing Improvements

This introduces a lot of changes and improvements to typing:

- Schema-to-ts has been rewritten to be smaller, faster, fewer dependencies, and easier to extend.
- Types for the different Perform functions have been added to Zapier Platform Core.
- Bundle `inputData` types will now be inferred from input fields.
- New helper functions have been added to help with typing:
  - `defineApp`, `defineTrigger`, `defineCreate`, `defineSearch`, `defineInputFields`
  - These are now recommended over the equivalent `satisfies Xyz` statements now.
  - `satisfies` is still used for perform functions, with `InferInputData`, and for other features without `define` helpers. For example `export default {…} satisfies Authentication` is still expected.
- Update templates and example apps for new define style


kreddlear (approved):
Thanks for tackling these changes and slowly dragging us into the future, Thomas! This looks great 👍

PR #1000, opened 2025-04-18
fix(core): PDE-6038 trim newline and whitespaces from request header

Natay:
Trim newline and whitespaces from the key/values of the request headers using a beforeMiddleware. This runs after the app specific before middleware, so headers are always ensured to be sanatized


packages/core/src/tools/create-request-client.js
```diff
@@ -24,6 +24,7 @@ const createRequestClient = (befores, afters, options) => {
   const httpAfters = [];
 
   if (!options.skipDefaultMiddle) {
+    httpBefores.push(sanitizeHeaders);
```
kola-er, line 27: I don't think it is necessary here any longer. You are already pushing it in the `befores` arg in both places the function `createRequestClient` is called. Moreover, the flow will never reach here for both callers because `skipDefaultMiddle` is set to `true` for both.

packages/core/src/tools/create-request-client.js
```diff
@@ -9,7 +9,6 @@ const applyMiddleware = require('../middleware');
 
 // before middles
 const prepareRequest = require('../http-middlewares/before/prepare-request');
-
```
kola-er, line 12: Let's add this line back

kola-er (approved):

PR #1001, opened 2025-04-21
PDE-5981 BREAKING CHANGE(core): stop replacing `{{curlies}}` unless it's a shorthand request

eliangcs:
This PR makes two changes:

* ❗ (BREAKING CHANGE) `z.request()` no longer supports `{{curlies}}` replacement unless it's used in a shorthand request ([PDE-5981](https://zapierorg.atlassian.net/browse/PDE-5981))
  * As a result, an exception needs to be made for legacy-scripting-runner, as legacy Web Builder scripts [may](https://github.com/zapier/zapier-platform/blob/43f0abbf20d08270716cd56904390e551cc162c4/packages/legacy-scripting-runner/test/example-app/index.js#L316) use `{{curlies}}` in the request options.
* 🐛 Fix a bug where not every `{{curlies}}` in the request object should be recursively replaced ([PDE-5652](https://zapierorg.atlassian.net/browse/PDE-5652))

## BREAKING CHANGE: `z.request()` no longer supports `{{curlies}}`

After this breaking change, calling `z.request()` with `{{bundle.*}}` or `{{process.env.*}}` will get an error. This code:

```js
    url: 'https://{{bundle.authData.subdomain}}.example.com',
    method: 'GET',
  });
  return response.data;
};
```

currently works in v16 but will get the following error in v17:

```
z.request() no longer supports {{bundle.*}} or {{process.*}} as of v17
unless it's used in a shorthand request.
Use JavaScript template literals instead.
Value in violation: "https://{{bundle.authData.subdomain}}.example.com" in attribute "url".
```

However, using `{{curlies}}` in a [shorthand request](https://docs.zapier.com/platform/build-cli/overview#shorthand-http-requests) will continue to work as is in v17:

```js
{
  operation: {
    perform: {
      url: 'https://{{bundle.authData.subdomain}}.example.com',
      method: 'GET',
    }
  }
}
```

We're introducing this breaking change because using JavaScript's [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) is generally recommended over the `{{curlies}}` syntax. For example, in v17, the `perform` method above should be rewritten as follows:

```js
    url: `https://${bundle.authData.subdomain}.example.com`,
    method: 'GET',
  });
  return response.data;
};
```

## Fix: Not every `{{curlies}}` in request object should be replaced

Besides request options like `url`, `params`, and `body`, the request object also holds a reference to [the `input` object](https://github.com/zapier/zapier-platform/blob/43f0abbf20d08270716cd56904390e551cc162c4/packages/core/src/http-middlewares/before/inject-input.js#L12). The `input` object is big—it holds the entire app definition under `input._zapier.app`. Currently, when `z.request()` replaces `{{curlies}}`, it does so recursively for _all_ the fields in the request object, including the `input` object. This isn't necessary. After the bug fix, `{{curlies}}` will be only replaced in the `url`, `params`, `body`, and `headers` fields—the only fields that are considered "replaceable".


kreddlear (approved):
Thanks for tackling these changes! This looks good to me 👍 

The only concern I can think of is that this error will only be thrown when the request is actually attempted (it won't be called out during validation), which seems like it could potentially cause issues for users if a dev updates their integration platform version but doesn't realize they need to replace curlies.

I suppose devs would read the docs to make sure they've handled breaking changes before updating to a major version though, and also the error will be thrown by `zapier test` if an integration has tests written. Just thought I'd call it out.

eliangcs:
> this error will only be thrown when the request is actually attempted (it won't be called out during validation)

That's true, and it's what this PR intents to do. To check it during validation, we'd rely on static analysis of the code. But I don’t think it's possible to detect the use of `{{curlies}}` through static analysis. Take this example:

```js
    url: bundle.inputData.url,
  });
  return response.data;
};
```

The `{{curlies}}` syntax doesn't appear in the code itself. But `bundle.inputData.url` could still contain `{{curlies}}`. This is one of those cases where you only can catch it by actually executing the code, not through static analysis.

PR #1002, opened 2025-04-22
fix(schema): PDE-6051 Add integer back to output fields schema

Natay:
Add `integer` type back to outputFields after Katie found some integrations that use this type.


kreddlear (approved):
Thanks for jumping on this so quickly!

PR #1005, opened 2025-05-01
chore(schema): PDE-6090 revert dynamic dropdown

Natay:
Remove function/request support in dynamic dropdowns in v17. We will release it as a minor version after ironing out some design decisions, more context in ticket


kreddlear (approved):
Thanks for reverting these changes, this looks good to me 👍

PR #1018, opened 2025-05-13
fix(core): Provide named esm exports of zpc objects

tkcranny:
This fixes issues where downstream apps can't access members of the zapier library via deconstructed import statements.

V17 can cause issues with the combination of trying to use destructured imports (i.e. `import {name} from 'zapier-platform-core`), from esm modules, while zapier-platform-core remains a CJS library. This adds a shim ES interface to the module, and updates package.json to use it when being consumed by ES modules.

```shell
$ zapier init ts-esm-test --template typescript -m esm
$ cd ts-esm-test && npm install
$ zapier build
```

### DIFF
--- a/packages/core/index.mjs
+++ b/packages/core/index.mjs
@@ -1,5 +1,9 @@
 import zapier from './src/index.js';
-
+import packageJson from './package.json' with { type: 'json' };
+import _tools from './src/tools/exported.js';
+zapier.version = packageJson.version;
+zapier.tools = _tools;
+// Allows `import { ... } from 'zapier-platform-core'`
 export const {
   createAppHandler,
   createAppTester,
@@ -10,4 +14,8 @@ export const {
   defineSearch,
   defineTrigger,
   integrationTestHandler,
+  tools,
+  version,
 } = zapier;
+// Allows `import zapier from 'zapier-platform-core'`
+export default zapier;
--- a/packages/core/index.mjs
+++ b/packages/core/index.mjs
@@ -1,0 +1,13 @@
+import zapier from './src/index.js';
+
+export const {
+  createAppHandler,
+  createAppTester,
+  defineApp,
+  defineCreate,
+  defineInputField,
+  defineInputFields,
+  defineSearch,
+  defineTrigger,
+  integrationTestHandler,
+} = zapier;
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,6 +15,7 @@
   "files": [
     "/include/",
     "/index.js",
+    "/index.mjs",
     "/src/",
     "/types/"
   ],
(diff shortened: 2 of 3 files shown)


packages/core/index.mjs
```diff
@@ -0,0 +1,13 @@
...
+  createAppHandler,
+  createAppTester,
+  defineApp,
+  defineCreate,
+  defineInputField,
+  defineInputFields,
+  defineSearch,
+  defineTrigger,
+  integrationTestHandler,
+} = zapier;
```
eliangcs, line 13: Looking at [`core/index.js`](https://github.com/zapier/zapier-platform/blob/e77f9be5bb99dd7044f31325273c48bd5ca2aefd/packages/core/index.js), the `zapier` object needs to contain `tools` and `version` as well.

Also, to allow default import like `import zapier from 'zapier-platform-core'`, we need to add `export default zapier;`.

Suggested change:

```suggestion
import zapier from './src/index.js';

import packageJson from './package.json' with { type: 'json' };
import _tools from './src/tools/exported.js';

zapier.version = packageJson.version;
zapier.tools = _tools;

// Allows `import { ... } from 'zapier-platform-core'`
export const {
  createAppHandler,
  createAppTester,
  defineApp,
  defineCreate,
  defineInputField,
  defineInputFields,
  defineSearch,
  defineTrigger,
  integrationTestHandler,
  tools,
  version,
} = zapier;

// Allows `import zapier from 'zapier-platform-core'`
export default zapier;
```

Since the code above uses import attributes `with`, I had to add an additional rule to `eslint.config.cjs` so eslint doesn't complain:

```
  {
    files: ['packages/core/**.mjs'],
    languageOptions: {
      ecmaVersion: 2025,
      sourceType: 'module',
    },
  }
```

packages/core/package.json
```diff
@@ -6,11 +6,16 @@
   "homepage": "https://platform.zapier.com/",
   "author": "Zapier Engineering <contact@zapier.com>",
   "license": "SEE LICENSE IN LICENSE",
-  "main": "index.js",
   "types": "types/index.d.ts",
+  "exports": {
+    "require": "./index.js",
+    "import": "./index.mjs",
+    "types": "./types/index.d.ts"
+  },
```
eliangcs, line 14: To fix the failing tests due to [the imports](https://github.com/zapier/zapier-platform/blob/9019b6f403587fc1e5ec97c9f9730a1f26c7ad8f/packages/legacy-scripting-runner/test/integration-test.js#L13-L16) in legacy-scripting-runner, maybe we need to do this:

```suggestion
  "exports": {
    ".": {
      "require": "./index.js",
      "import": "./index.mjs",
      "types": "./types/index.d.ts"
    },
    "./src/*": {
      "require": "./src/*.js"
    }
  },
```

eliangcs (approved):

Issue #1019, opened 2025-05-14
[Bug]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be  valid file:// URLs. Received protocol 'c:'


SergeyBonich:
### Bug Description

On Windows 11 zapier cli commands (test, build, validate) return following error:

`Error: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be  valid file:// URLs. Received protocol 'c:'`

It affects only version 17.0.0. Rolling back to 16.5.1 solves the issue

### Reproduction Steps

```
npm install -g zapier-platform-cli@17.0.0
zapier init test-integration
cd test-integration
npm install
zapier build
```

results in
`Error: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be  valid file:// URLs. Received protocol 'c:'`

### Zapier Platform version

17.0.0

### Node.js version

22.15.0

### Your Operating System

Windows 11

eliangcs:
@SergeyBonich Thanks for reporting! I don't have a Windows machine or VM to test. Can you post your output of `zapier build --debug` to help us troubleshoot?

SergeyBonich:
Here is the result of `zapier build --debug`

```
 oclif:zapier:build argv is [] +0ms
  oclif:zapier:build args are {} +1ms
  oclif:zapier:build flags are { debug: true } +0ms
  oclif:zapier:build ------------ +1ms
  zapier:analytics sending {
  command: 'build',
  isValidCommand: true,
  numArgs: 0,
  appId: undefined,
  argsKeys: [],
  flagKeys: [ 'debug' ],
  cliVersion: '17.0.0',
  os: 'win32',
  sendUserId: true
} +0ms
  zapier:build Using temp directory:  C:\Users\SD9D3~1.BON\AppData\Local\Temp\zapier-f05dfec8 +0ms
Copying project to temp directory
Installing project dependencies  zapier:misc
  zapier:misc  +0ms
  zapier:misc Running npm.cmd install --production command in C:\Users\SD9D3~1.BON\AppData\Local\Temp\zapier-f05dfec8:
  zapier:misc  +0ms
Installing project dependencies  zapier:misc npm warn config production Use `--omit=dev` instead.
  zapier:misc  +348ms
Installing project dependencies  zapier:api >> POST https://zapier.com/api/platform/cli/analytics +0ms
[...]

PR #1024, opened 2025-05-15
PDE-6136 - fix(cli) - Fix Path for windows

standielpls:
Fix for this issue: https://github.com/zapier/zapier-platform/issues/1019

### DIFF
--- a/packages/cli/src/utils/local.js
+++ b/packages/cli/src/utils/local.js
@@ -254,7 +254,8 @@ const loadAppRawUsingImport = async (
 };
 
 const loadAppRawUsingRequire = (appDir) => {
-  let appRaw = require(appDir);
+  const normalizedPath = path.resolve(appDir);
+  let appRaw = require(normalizedPath);
   if (appRaw && appRaw.default) {
     // Node.js 22+ supports using require() to import ESM.
     // For Node.js < 20.17.0, require() will throw an error on ESM.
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/local.js
```diff
@@ -228,7 +228,10 @@ const loadAppRawUsingImport = async (
   try {
     // zapierwrapper.mjs is only available since zapier-platform-core v17.
     // And only zapierwrapper.mjs exposes appRaw just for this use case.
-    appRaw = (await import(wrapperPath)).appRaw;
+    // Convert path to file:// URL for Windows compatibility with ESM imports
+    const pathForUrl = path.resolve(wrapperPath).replace(/\\/g, '/');
+    const wrapperUrl = new URL(`file:///${pathForUrl.replace(/^\//, '')}`);
```
Natay, line 233: I could be wrong but should `file:///` be `file://` ? ( just judging from the comment above )

kreddlear: Looks like it's part of the [File URI spec](https://en.wikipedia.org/wiki/File_URI_scheme), the format is `file://host/path` but if `host` is empty it means it's your local machine - good question, TIL!

Natay: Ahh thank you!

kreddlear (approved):

PR #1032, opened 2025-05-20
PDE-6149 fix(core): internalize flag that enables/disable `{{curlies}}` replacement, disable `{{curlies}}` for Raw Requests

eliangcs:
Changes:

- Internalize the flag that enables/disables `{{curlies}}` replacement in `z.request()`.
  - Use a [symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) instead of a property like `req.replace`. This makes it impossible to enable the replacement behavior with `z.request({ replace: true })`.
- Remove `{{curlies}}` support from Raw Requests ([`execute-request.js`](https://github.com/zapier/zapier-platform/blob/2cd98d1a404a7e9acbe1846c25244edb6769f67b/packages/core/src/execute-request.js))
- Disallow multi-pass `{{curlies}}` replacement.
  - See test case `should replace curlies in single pass` for what this change is about.
- legacy-scripting-runner's `z.request()` client still needs to keep `{{curlies}}` replacement for backward compatibility.

This is not a breaking change because:
- `req.replace` is already an internal flag, which this PR is removing.
- The v17.0.0 changelog already declared that `{{curlies}}` are no longer supported for manual requests (i.e. calling `z.request()` in a function).
- Raw Requests already don't support `{{curlies}}`. Currently, `{{curlies}}` are silently removed before the raw request is sent. After this change, it will throw an error.

### DIFF
--- a/packages/core/src/constants.js
+++ b/packages/core/src/constants.js
@@ -26,7 +26,6 @@ const RENDER_ONLY_METHODS = [
 const REPLACE_CURLIES = Symbol('replaceCurlies');
 
 const REQUEST_OBJECT_SHORTHAND_OPTIONS = {
-  isShorthand: true,
   [REPLACE_CURLIES]: true,
 };
 
--- a/packages/core/src/constants.js
+++ b/packages/core/src/constants.js
@@ -23,7 +23,12 @@ const RENDER_ONLY_METHODS = [
   'authentication.oauth1Config.authorizeUrl',
 ];
 
-const REQUEST_OBJECT_SHORTHAND_OPTIONS = { isShorthand: true, replace: true };
+const REPLACE_CURLIES = Symbol('replaceCurlies');
+
+const REQUEST_OBJECT_SHORTHAND_OPTIONS = {
+  isShorthand: true,
+  [REPLACE_CURLIES]: true,
+};
 
 const DEFAULT_LOGGING_HTTP_ENDPOINT = 'https://httplogger.zapier.com/input';
 const DEFAULT_LOGGING_HTTP_API_KEY = 'R24hzu86v3jntwtX2DtYECeWAB'; // It's ok, this isn't PROD
@@ -65,13 +70,14 @@ module.exports = {
   IS_TESTING,
   KILL_MAX_LIMIT,
   KILL_MIN_LIMIT,
+  NON_STREAM_UPLOAD_MAX_SIZE,
   PACKAGE_NAME,
   PACKAGE_VERSION,
   RENDER_ONLY_METHODS,
+  REPLACE_CURLIES,
   REQUEST_OBJECT_SHORTHAND_OPTIONS,
   RESPONSE_SIZE_LIMIT,
   SAFE_LOG_KEYS,
   STATUSES,
   UPLOAD_MAX_SIZE,
-  NON_STREAM_UPLOAD_MAX_SIZE,
 };
--- a/packages/core/src/execute.js
+++ b/packages/core/src/execute.js
@@ -10,18 +10,15 @@ const prepareRequest = require('./http-middlewares/before/prepare-request');
 const constants = require('./constants');
 
 const executeHttpRequest = (input, options) => {
-  options = _.extend(
-    {},
+  options = {
     // shorthand requests should always throw _unless_ the object specifically opts out
     // this covers godzilla devs who use shorthand requests (most of them) that rely on the throwing behavior
     // when we set the app-wide skip for everyone, we don't want their behavior to change
     // so, this line takes precedence over the global setting, but not the local one (`options`)
-    {
-      skipThrowForStatus: false,
-    },
-    options,
-    constants.REQUEST_OBJECT_SHORTHAND_OPTIONS,
-  );
+    skipThrowForStatus: false,
+    ...options,
+    ...constants.REQUEST_OBJECT_SHORTHAND_OPTIONS,
+  };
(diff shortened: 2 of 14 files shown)


packages/core/src/constants.js
```diff
@@ -23,7 +23,12 @@ const RENDER_ONLY_METHODS = [
   'authentication.oauth1Config.authorizeUrl',
 ];
 
-const REQUEST_OBJECT_SHORTHAND_OPTIONS = { isShorthand: true, replace: true };
+const REPLACE_CURLIES = Symbol('replaceCurlies');
+
+const REQUEST_OBJECT_SHORTHAND_OPTIONS = {
+  isShorthand: true,
```
FokkeZB, line 29: This isn't referenced anywhere, so we can nuke this.

But perhaps we better use something like `IS_INTERNAL_SHORTHAND = Symbol('isInternalShorthand')` to also for ourselves be more clear?

packages/core/src/tools/request-merge.js
```diff
@@ -13,6 +14,9 @@ const caseInsensitiveMerge = (requestOne, requestTwo, requestThree) => {
 
   // This is a very quick & efficient merge for all of request's properties
   const mergedRequest = _.merge(requestOne, requestTwo, requestThree);
+  if (requestThree[REPLACE_CURLIES]) {
+    mergedRequest[REPLACE_CURLIES] = requestThree[REPLACE_CURLIES];
+  }
```
FokkeZB, line 19: Is this because `_.merge` doesn't handle symbols well? If so, let's add a comment to explain.

packages/core/src/execute.js
```diff
@@ -10,18 +10,15 @@ const prepareRequest = require('./http-middlewares/before/prepare-request');
 const constants = require('./constants');
 
 const executeHttpRequest = (input, options) => {
-  options = _.extend(
-    {},
+  options = {
```
eliangcs, line 13: I found after `_.extend(a, b)`, the symbol keys in the object are gone, so I've replaced `_.extend()` with object spread syntax (`{...a, ...b}`).

FokkeZB (approved):
Minor naming suggestion, but otherwise 👍

packages/core/src/http-middlewares/before/prepare-request.js
```diff
@@ -108,8 +110,9 @@ const throwForCurlies = (value, path) => {
     if (/{{\s*(bundle|process)\.[^}]*}}/.test(value)) {
       throw new Error(
         'z.request() no longer supports {{bundle.*}} or {{process.*}} as of v17 ' +
-          "unless it's used in a shorthand request. " +
-          'Use JavaScript template literals instead. ' +
+          "unless it's used in a shorthand request defined by the integration. " +
+          'Zapier Customers: Remove "{{curly braces}}" from your request. ' +
```
eliangcs, line 114: Users who use Raw Requests might see this error message, so I added some info for Zapier customers.

packages/core/src/tools/request-merge.js
eliangcs, line 19: Symbol keys are gone after `_.merge()`, so I'm adding the symbol back here.

packages/core/types/schemas.generated.d.ts
```diff
@@ -4,7 +4,7 @@
  * files, and/or the schema-to-ts tool and run its CLI to regenerate
  * these typings.
  *
- * zapier-platform-schema version: 17.0.1
+ * zapier-platform-schema version: 17.0.2
```
eliangcs, line 7: This change should've been in the last released.

packages/core/src/constants.js
eliangcs: Deleted in [9735102](https://github.com/zapier/zapier-platform/pull/1032/commits/973510225b1f5dbcba908438101e1aae974047cc).

packages/core/src/tools/request-merge.js
eliangcs: Added in [c89bd2c](https://github.com/zapier/zapier-platform/pull/1032/commits/c89bd2c39f59000ab070c23cc9c2ab9b970c15e6).

Natay:
> After this change, it will throw an error.

🤔 Could this be considered a breaking change?

packages/core/src/tools/cleaner.js
```diff
@@ -43,66 +43,96 @@ const recurseCleanFuncs = (obj, path) => {
   return obj;
 };
 
+const findNextCurlies = (str) => {
+  const start = str.indexOf('{{');
```
Natay, line 47: Is it possible to do this using regext? something like: 

```
const findNextCurlies = (str) => {
  const match = str.match(/{{.*?}}/);
  if (!match) {
    return {
      start: -1,
      end: -1,
    };
  }

  return {
    start: match.index,
    end: match.index + match[0].length,
  };
};
```
Or are you avoiding that on purpose?

eliangcs: I assumed `indexOf` should be faster. The efficiency is probably insignificant. But I'm curious, so I did a little [benchmark](https://jsfiddle.net/s019axu2/). The result: The `indexOf` version is ~6x as fast as the regex version.

Natay: Ah that's really nice to know! Thanks for the extra due-diligence with the benchmark!

Natay (approved):

PR #1033, opened 2025-05-22
PDE-6151 fix(cli): "Cannot find base config file" tsconfig warning

eliangcs:
Fixes this warning message when running `zapier build`:


kreddlear (approved):
Nice, this is exactly the fix I came up with before the questions about building in place threw me off track. Thanks for confirming that `esbuild` does not transpile TS so it shouldn't cause any issues to set the config to empty here - that was my main worry that it would somehow break something. It should work great to resolve this warning 👍 

I'm not sure what is going on with the failing tests - they don't seem related to the change at all and they pass locally. I'll try re-running.

PR #1034, opened 2025-05-22
PDE-6145 fix(core): `{{curlies}}` in `requestTemplate` should be replaced

eliangcs:
This PR fixes a regression on v17 where `{{curlies}}` in `requestTemplate` (a property in [`AppSchema`](https://github.com/zapier/zapier-platform/blob/2cd98d1a404a7e9acbe1846c25244edb6769f67b/packages/schema/docs/build/schema.md#appschema)) are no longer rendered. For example, an integration may define:

```
{
  requestTemplate: {
    headers: {
      authorization: 'Bearer {{bundle.authData.access_token}}'
    }
  }
}
```

Given `bundle.authData.access_token = 'a_token'`:

Expected behavior: Every `z.request()` call automatically includes the header `authorization: Bearer a_token`.

Current behavior in v17 (>=17.0.0 <=17.0.2): `z.request()` throws an error:

```
z.request() no longer supports {{bundle.*}} or {{process.*}} as of v17
```

### DIFF
--- a/packages/core/src/http-middlewares/before/prepare-request.js
+++ b/packages/core/src/http-middlewares/before/prepare-request.js
@@ -148,33 +148,44 @@ const prepareRequest = function (req) {
 
   req = sugarBody(req);
 
-  // apply app requestTemplate to request
-  if (req.merge) {
-    const requestTemplate = (input._zapier?.app || {}).requestTemplate;
-    req = requestMerge(requestTemplate, req);
-  }
-
-  const replaceable = {
-    url: req.url,
-    headers: req.headers,
-    params: req.params,
-    body: req.body,
-  };
-
-  if (req[REPLACE_CURLIES]) {
-    // replace {{curlies}} in the request
+  if (req[REPLACE_CURLIES] || req.merge) {
     const bank = createBundleBank(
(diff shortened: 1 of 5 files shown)


packages/core/src/http-middlewares/before/prepare-request.js
```diff
@@ -148,33 +148,44 @@ const prepareRequest = function (req) {
...
-      input._zapier.app,
-      input._zapier.event,
+      input?._zapier?.event || {},
       req.serializeValueForCurlies,
     );
-    req = {
-      ...req,
-      ...recurseReplaceBank(replaceable, bank),
+
+    const replaceable = {
```
kola-er, line 157:
```suggestion
    let replaceable = {
```

eliangcs: Good callout! Instead of reusing the same variable name, I've changed to use two different variable names in [b8ae466](https://github.com/zapier/zapier-platform/pull/1034/commits/b8ae466fc60a736d485a4b03e62d2718c102464b).

kola-er (approved):

PR #1038, opened 2025-05-26
fix(cli): prevent .cmd on custom binaries

SamPoate:
Prevents .cmd from being added to anything like node, node.exe, or custom binaries.

// Copilots thoughts:
This pull request updates the `runCommand` utility function to improve compatibility and security when executing commands on Windows systems.

### Compatibility improvements:
* Updated the `runCommand` function in `packages/cli/src/utils/misc.js` to append `.cmd` only for specific commands (`npm`, `npx`, `yarn`) by introducing a conditional check. This ensures better compatibility and avoids unnecessary modifications for other commands.

### DIFF
--- a/packages/cli/src/utils/misc.js
+++ b/packages/cli/src/utils/misc.js
@@ -32,7 +32,10 @@ const runCommand = (command, args, options) => {
   options = options || {};
 
   if (isWindows()) {
-    command += '.cmd';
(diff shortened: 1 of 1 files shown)


eliangcs:
@SamPoate thanks for the patch! From a quick look, it seems the `runCommand` function is only used for `npm` and `yarn`. What’s your use case for using it with node or other custom binaries?

PR #1044, opened 2025-05-29
fix(core): Remove request body fields using undefined input data curlies for application/x-www-form-urlencoded content-type

kola-er:
Unlike the requests with the header `Content-Type: application/json`, fields of the request body dependent on an undefined input data curlies are not removed for requests with header `Content-Type: application/x-www-form-urlencoded`. This change fixes that for both cases to ensure the removal.

### DIFF
--- a/packages/core/src/http-middlewares/before/prepare-request.js
+++ b/packages/core/src/http-middlewares/before/prepare-request.js
@@ -52,6 +52,7 @@ const coerceBody = (req) => {
 
   // auto coerce form if header says so
   if (contentType === FORM_TYPE && req.body && !_.isString(req.body)) {
+    normalizeEmptyBodyFields(req);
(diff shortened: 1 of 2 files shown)


packages/core/src/http-middlewares/before/prepare-request.js
```diff
@@ -52,6 +52,7 @@ const coerceBody = (req) => {
 
   // auto coerce form if header says so
   if (contentType === FORM_TYPE && req.body && !_.isString(req.body)) {
+    normalizeEmptyBodyFields(req);
```
kola-er, line 55: Similar to what is done [here](https://github.com/zapier/zapier-platform/blob/2636637b61f74c8804d95e56f2b1a841519196b8/packages/core/src/http-middlewares/before/prepare-request.js#L66).

eliangcs (approved):

PR #1048, opened 2025-06-03
SPIKE: Accept JSON for zapier convert

FokkeZB:


packages/cli/src/tests/register.test.js
```diff
@@ -178,10 +178,12 @@ describe('RegisterCommand', () => {
...
-    fs.writeFileSync(
-      APP_RC_FILE,
-      `{"id":${privateApp.id},"key":"App${privateApp.id}"}`,
-    );
+    beforeEach(function () {
+      fs.writeFileSync(
+        APP_RC_FILE,
+        `{"id":${privateApp.id},"key":"App${privateApp.id}"}`,
+      );
+    });
```
FokkeZB, line 186: This wasn't wrapped in `beforeEach()` and thus always executed, even without the suite included and thus not cleaning up the file in its `after()`.

eliangcs (approved):
I didn't test but it looks good to me!

packages/cli/docs/cli.md
```diff
@@ -114,11 +114,14 @@ If you re-run this command on an existing directory it will leave existing files
...
 
 **Arguments**
-* (required) `integrationId` | To get the integration/app ID, go to "https://developer.zapier.com", click on an integration, and copy the number directly after "/app/" in the URL.
 * (required) `path` | Relative to your current path - IE: `.` for current directory.
 
 **Flags**
-* (required) `-v, --version` | Convert a specific version. Required when converting a Visual Builder integration.
+* `-i, --integrationId` | To get the integration/app ID, go to "https://developer.zapier.com", click on an integration, and copy the number directly after "/app/" in the URL.
+* `-v, --version` | Convert a specific version. Required when converting a Visual Builder integration.
+* `-j, --json` | The JSON definition to use, as alternative for reading from a Visual Builder integration. Must be a JSON-encoded object. The data can be passed from the command directly like '{"key": "value"}', read from a file like @file.json, or read from stdin like @-.
```
kreddlear, line 122: Could we add an example of what the definition JSON should look like below the flags, to make it easier for external users to take advantage of this? If it'll be quite long we could always put it in a [collapsed section](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections).

eliangcs: I'll merge it as is and cut a release. This is something that we can add later if @FokkeZB agrees.

FokkeZB: @kreddlear a full definition will indeed be lengthy, but we can link to https://github.com/zapier/zapier-platform/blob/main/packages/schema/examples/definition.json

PR #1050, opened 2025-06-05
feat(core): PDE-6179 add before middleware to fetch stashed bundle.

Natay:
### Summary
When input bundles exceed AWS Lambda size limits, they will be encrypted and stored in S3, then automatically decrypted when retrieved. This middleware detects encrypted bundles and decrypts them using a one-time secret, allowing integrations to handle large payloads without code changes.

### Key Changes
- New fetchStashedBundle middleware - that automatically detects and decrypts encrypted bundles. 
- fernet dependency - Included proper Fernet token handling library for secure encryption/decryption operations that match. backend.

### DIFF
--- a/packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
+++ b/packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
@@ -16,12 +16,11 @@ const fetchStashedBundle = async (input) => {
   const secret = process.env._ZAPIER_ONE_TIME_SECRET;
   if (stashedBundleKey && secret) {
     // Use the RPC to get a presigned URL for downloading the data
-    const s3UrlResponse = await rpc(
+    const rpcResponse = await rpc(
       'get_presigned_download_url',
       stashedBundleKey,
     );
-    const s3Url = s3UrlResponse.url;
-    const response = await withRetry(() => fetch(s3Url));
+    const response = await withRetry(() => fetch(rpcResponse.url));
     if (!response.ok) {
       const errorMessage = `Failed to read stashed bundle. Status: ${response.status} ${response.statusText}`;
       throw new StashedBundleError(errorMessage);


packages/legacy-scripting-runner/test/integration-test.js
```diff
@@ -29,10 +29,13 @@ const createAppWithCustomBefores = (appRaw, customBefores) => {
   const waitForPromises = require('zapier-platform-core/src/app-middlewares/after/wait-for-promises');
   const createCommandHandler = require('zapier-platform-core/src/create-command-handler');
   const applyMiddleware = require('zapier-platform-core/src/middleware');
+  const fetchStashedBundle = require('zapier-platform-core/src/app-middlewares/before/fetch-stashed-bundle');
```
standielpls: Does this test fail without this? 
I didn't know legacy-scripting-runner uses `zapier-platform-core` middleware 🤔

packages/core/src/create-app.js
```diff
@@ -29,6 +30,7 @@ const createApp = (appRaw) => {
   const befores = [
     addAppContext,
     injectZObject,
+    fetchStashedBundle,
```
standielpls: I think this needs to go first because `addAppContext` [uses the bundle](https://github.com/zapier/zapier-platform/blob/65c8ec25814a3ce8980a287b4ec1973949561725/packages/core/src/app-middlewares/before/add-app-context.js#L17)

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+const fetch = require('../../tools/fetch');
+const _ = require('lodash');
+const { StashedBundleError } = require('../../errors');
+const crypto = require('crypto');
+// Use the fernet npm package for proper Fernet token handling
+const fernet = require('fernet');
+
+const withRetry = async (fn, retries = 3, delay = 100, attempt = 0) => {
```
standielpls: I think this was copied from `create-response-stasher.js` right? Maybe it is worth extracting into a util function since we're using it again

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
```diff
@@ -0,0 +1,106 @@
...
+  try {
+    return await fn();
+  } catch (error) {
+    if (attempt >= retries) {
+      throw error;
```
standielpls: One thing I did want to optimize for in this function was to indicate it was retried and failed. This helps us debug if it was a request error or a server error

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
```diff
@@ -0,0 +1,106 @@
...
+ * This matches the backend implementation:
+ * 1. Hash the secret with SHA256 to get 32 bytes
+ * 2. Base64url encode those bytes to make Fernet-compatible key
+ * 3. Use Fernet library to decrypt (handles all token parsing internally)
+ *
+ * @param {string} fernetToken - The Fernet token to decrypt
+ * @param {string} secret - The secret key for decryption
+ * @returns {Object} The decrypted bundle object
+ */
+const decryptBundleWithSecret = (fernetToken, secret) => {
```
standielpls: Cool! Wonder if this is worth moving to a util function? But nbd, since we don't have any other encrypt/decrypt

packages/legacy-scripting-runner/test/integration-test.js
Natay: It passes without it, yeah I was surprised first seeing zapier-platform-core middlewares here 🤔 , honestly not sure if the app middlewares applied to legacy scripting as well? I added it here because I saw the other middlewares here and didn't think it was a big deal to include this for regression testing

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
Natay: I expanded the error we throw to indicate the number of failures and the last error message in https://github.com/zapier/zapier-platform/pull/1050/commits/025c35c4c6fbb269fe905ba880a5f80f4381b4e5

packages/core/src/create-app.js
Natay: Ah nice catch!

packages/legacy-scripting-runner/test/integration-test.js
eliangcs: Do we need to support stashed bundle for apps converted from the legacy Web Builder? If not, we probably don't need to add this to legacy-scripting-runner.

If we do, legacy-scripting-runner has to work compatibly with older versions of zapier-platform-core. That means this `require()` needs to be optional. Example:

```js
let fetchStashedBundle;
try {
  fetchStashedBundle = require('zapier-platform-core/src/app-middlewares/before/fetch-stashed-bundle');
} catch (err) {
  // Older versions of platform-core don't have fetch-stashed-bundle middleware
}

// ...

let befores = [addAppContext, injectZObject];
if (fetchStashedBundle) {
  befores.push(fetchStashedBundle);
}
befores = befores.concat(customBefores || []);
```

I also noticed there's the same problem with [the `sanitizeHeaders` middleware](https://github.com/zapier/zapier-platform/blob/0c5d8292d4435ce0a57f46264e50f6e4b1d85865/packages/legacy-scripting-runner/index.js#L34). @Natay can you help fix that as well? Thanks!

Natay: Thanks for the heads up! I don't think `fetchStashedBundle` will apply to apps converted from legacy web builder, the "LambdaPayloadTooLargeException" error is heavily concentrated to WebHookCLIAPI which from my understanding is not a legacy web builder app converted. I'll remove `fetchStashedBundle` from here and update sanitizeHeaders

Natay: Updated in https://github.com/zapier/zapier-platform/pull/1050/commits/bc548610e7fa0de931ea3b33d831bda7776f933b

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const fetch = require('../../tools/fetch');
+const _ = require('lodash');
+const { StashedBundleError } = require('../../errors');
+const { withRetry } = require('../../tools/retry-utils');
+const { decryptBundleWithSecret } = require('../../tools/bundle-encryption');
```
standielpls: import order

`lodash` goes first

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
```diff
@@ -0,0 +1,44 @@
...
+    const s3UrlResponse = await rpc(
+      'get_presigned_download_url',
+      stashedBundleKey,
+    );
+    const s3Url = s3UrlResponse.url;
+    const response = await withRetry(() => fetch(s3Url));
+    if (!response.ok) {
+      const errorMessage = `Failed to read stashed bundle from S3. Status: ${response.status} ${response.statusText}`;
+      throw new StashedBundleError(errorMessage);
+    }
```
standielpls: nitpick: we shouldn't need to have any notion of `s3`. Using S3 is a backend behaviour, that doesn't need to be exposed across the interface boundary - we just have to worry about how we fetch the bundle here.

packages/core/src/tools/bundle-encryption.js
```diff
@@ -0,0 +1,57 @@
...
+ * This matches the backend:
+ * 1. Hash the secret with SHA256 to get 32 bytes
+ * 2. Base64url encode those bytes to make Fernet-compatible key
+ * 3. Use Fernet library to decrypt (handles all token parsing internally)
+ *
+ * @param {string} fernetToken - The Fernet token to decrypt
+ * @param {string} secret - The secret key for decryption
+ * @returns {Object} The decrypted bundle object
+ */
+const decryptBundleWithSecret = (fernetToken, secret) => {
```
standielpls: i'd change the function signature naming a bit either to lean more general (my preference) or more specific. 
i.e.
`decryptWithSecret(payload, secret, keyLength? = 32)` 
or 
`decryptBundleWithSecret(bundle, secret)`

standielpls (dismissed):
I have some suggestions to make this a bit more clean but non-blocking

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
standielpls: I meant any reference to S3 in this file, not just the error msg. The interface is through `get_presigned_download_url` and you get back a `url` to fetch.

standielpls (dismissed):

packages/core/src/app-middlewares/before/fetch-stashed-bundle.js
Natay: ahh 🤦 that makes sense, I'll update it!

PR #1052, opened 2025-06-09
PDE-6175 feat(cli): faster `zapier build` and support for npm/yarn/pnpm workspaces

eliangcs:
Here are tables comparing behavior before and after:

Before:

Command | npm/yarn workspace | pnpm workspace
---|---|---
`zapier build` | ✅ | ✅ 
`zapier build --skip-npm-install` | ❗ (limited support) | ❌ (not supported)

After:

Command | npm/yarn workspace | pnpm workspace
---|---|---
`zapier build` | ✅ | ✅ 
`zapier build --skip-npm-install` | ✅ | ✅ 

For more details, see the following sections.

## Background & current behavior

Until now `zapier build` assumed a single-package repo, with limited support for workspaces only when [zapier-platform-core is hoisted](https://github.com/zapier/zapier-platform/blob/24f22e1aff363648967ed9f4fbae99af9b5b6a6d/packages/cli/src/utils/build.js#L325-L327) to the workspace root's node_modules.

```
my-zapier-app/                ← current working directory
├─ index.js
├─ triggers/
│  └─ new_message.js
├─ zapierwrapper.js
├─ definition.json
├─ node_modules/
│  └─ ...
└─ package.json
```

1. Copy the whole directory to a temporary folder.
2. `npm install` inside that temp folder (unless `--skip-npm-install`).
3. Run esbuild to follow `require()` / `import` graphs and collect additional files.
4. Add them to a zip.

`zapier build --skip-npm-install` skipped step 2 but still copied the directory. The zip always came from the app folder. That breaks in a workspace, where:

* dependencies are hoisted to the workspace root
* node_modules may be symlinks (pnpm)

### Detecting a workspace

`zapier build --skip-npm-install` will detect if the project is in a workspace by:

* npm/yarn: `package.json` contains a [`workspaces`](https://docs.npmjs.com/cli/v11/using-npm/workspaces) field.
* pnpm: a [`pnpm-workspace.yaml`](https://pnpm.io/settings) file exists.

If either test passes the workspace root becomes the "project root" for packaging.

### Example layouts

Note that npm/yarn and pnpm define a "workspace" differently. In npm/yarn, each package in a monorepo is a workspace, In pnpm, the monorepo is a workspace.

npm/yarn workspaces:

```
monorepo/
├─ package.json            ← { "workspaces": ["packages/*"] }
├─ packages/
│  ├─ app-1/               ← CWD when you run `zapier build`
│  │  ├─ index.js
│  │  ├─ zapierwrapper.js
│  │  ├─ definition.json
│  │  └─ package.json
│  └─ shared-lib/
└─ node_modules/           ← hoisted deps (zapier-platform-core, lodash…)
```

pnpm workspace:

```
monorepo/
├─ pnpm-workspace.yaml     ← defines packages/**
├── node_modules/
│   └── .pnpm/             ← actual store for deps (zapier-platform-core...)
│       ├── zapier-platform-core@17.0.0/
│       │   └── node_modules/
│       │       ├── zapier-platform-core/
│       │       │   ├── package.json
│       │       │   └── ...
│       │       ├── lodash/ -> ../../lodash@4.17.21/node_modules/lodash
│       │       └── deps of zapier-platform-core@17.0.0 -> ...
│       ├── lodash@4.17.21/
│       │   └── node_modules/
│       │       ├── lodash/
│       │       │   └── ...
│       │       └── deps of lodash@4.17.21 -> ...
│       └── ...
├── packages/
│   ├── app-1/
│   │   ├── node_modules/      ← symlinks into .pnpm store
│   │   │   ├── zapier-platform-core -> ../../../node_modules/.pnpm/zapier-platform-core@17.0.0/node_modules/zapier-platform-core
│   │   │   └── deps for app-1 -> ...
│   │   ├── package.json
│   │   ├── index.js
│   │   └── ...
│   └── ...
```

### How each mode now works

* `zapier build` - Legacy flow under the hood (copy app dir → `npm install` in temp → esbuild). Always works, even inside a workspace, because everything needed lives under the app folder after install.

* `zapier build --skip-npm-install` - Fast-path for monorepos.
  1. No copy, no `npm install`, much faster build.
  2. Sets workspace root as the zip root.
  3. Creates symlinks back to `packages/app-1/zapierwrapper.js` and `definition.json` so the Lambda runtime still finds them.
  4. Run esbuild to follow `require()` / `import` graphs and collect additional files.
  5. Add them to a zip.

Speed win: In a test on a large npm/yarn workspace monorepo, the build time (with `--skip-npm-install --skip-validation`) dropped from 7.4s → 2.4s.

https://github.com/user-attachments/assets/5231424e-7077-4036-9616-79ed42de98a2


packages/cli/src/constants.js
```diff
@@ -17,13 +18,6 @@ const BUILD_DIR = 'build';
 const DEFINITION_PATH = `${BUILD_DIR}/definition.json`;
 const BUILD_PATH = `${BUILD_DIR}/build.zip`;
 const SOURCE_PATH = `${BUILD_DIR}/source.zip`;
-const BLOCKLISTED_PATHS = [
-  // Will be excluded from build.zip and source.zip
-  '.git',
-  '.env',
-  '.environment',
-  'build',
-];
```
eliangcs, line 26: I moved this to `utils/ignore.js` as no one else is using this constant.

packages/cli/package.json
```diff
@@ -59,12 +59,10 @@
     "ignore": "5.2.4",
     "inquirer": "8.2.5",
     "jscodeshift": "^17.3.0",
-    "klaw": "4.1.0",
     "lodash": "4.17.21",
     "luxon": "3.6.1",
     "marked": "14.1.4",
     "marked-terminal": "7.2.1",
-    "minimatch": "9.0.3",
```
eliangcs, line 67: I was able to remove `klaw` and `minimatch` because we no longer need them for `zapier build`.

.github/workflows/ci.yaml
```diff
@@ -86,9 +86,13 @@ jobs:
         node-version: [18.x, 20.x]
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
```
eliangcs, line 91: I added a test suite that uses pnpm to set up a monorepo, so we need to install it in CI.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
     platform: 'node',
     metafile: true,
     logLevel: 'warning',
-    external: ['../test/userapp'],
+    external: [
+      '../test/userapp',
+      'zapier-platform-core/src/http-middlewares/before/sanatize-headers', // appears in zapier-platform-legacy-scripting-runner/index.js
```
eliangcs, line 69: Without this line, esbuild will error that `sanatize-headers` is missing because legacy-scripting-runner was [requiring sanatize-headers](https://github.com/zapier/zapier-platform/blob/zapier-platform-legacy-scripting-runner%403.8.17/packages/legacy-scripting-runner/index.js#L34), which wasn't available in all versions of platform-core. This issue happens on legacy-scripting-runner 3.8.15-3.8.17; it will be [fixed](https://github.com/zapier/zapier-platform/blob/4497179622c4d2413588cd813ecafc24fe809b20/packages/legacy-scripting-runner/index.js#L35-L40) in legacy-scripting-runner in the next release.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
+      './request-worker', // appears in zapier-platform-legacy-scripting-runner/zfactory.js
+      './xhr-sync-worker.js', // appears in jsdom/living/xmlhttprequest.js
```
eliangcs, line 71: I added these two because esbuild were warning about the `require.resolve()` usage on them.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
-    filePath.endsWith('package.json') ||
-    filePath.endsWith('definition.json') ||
-    // include old async deasync versions so this runs seamlessly across node versions
-    filePath.endsWith(path.join('bin', 'linux-x64-node-10', 'deasync.node')) ||
-    filePath.endsWith(path.join('bin', 'linux-x64-node-12', 'deasync.node')) ||
-    filePath.endsWith(path.join('bin', 'linux-x64-node-14', 'deasync.node')) ||
-    filePath.endsWith(
-      // Special, for zapier-platform-legacy-scripting-runner
-      path.join('bin', `linux-x64-node-${nodeMajorVersion}`, 'deasync.node'),
-    ) ||
```
eliangcs, line 135: These deasync files are no longer required for quite some time. I've removed them from the new implementation.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
-    filePath.match(
-      path.sep === '\\' ? /aws-sdk\\apis\\.*\.json/ : /aws-sdk\/apis\/.*\.json/,
```
eliangcs, line 137: These `aws-sdk` files are still needed when an integration still depends on aws-sdk v2. These are still in the new implementation.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
+  ).sort();
+
+  const workspaceRoot = (await findWorkspaceRoot(workingDir)) || workingDir;
+
+  if (workspaceRoot !== workingDir) {
+    const appDirRelPath = path.relative(workspaceRoot, workingDir);
+    const linkNames = ['zapierwrapper.js', 'index.js'];
+    for (const name of linkNames) {
+      zip.symlink(name, path.join(appDirRelPath, name), 0o644);
+    }
```
eliangcs: If there's an upper level workspace root, we need to add symlinks to `zapierwrapper.js` and `index.js` in the app directory, as such:

```
monorepo/   <- will also be the zip file's root
  zapierwrapper.js -> packages/app-1/zapierwrapper.js
  index.js         -> packages/app-1/inde.js
  packages/
    app-1/
      zapierwrapper.js
      index.js
```

I tested it. Lambda is fine with it and can execute `zapierwrapper.js` as if it were a regular `.js` file.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
-  const [dumbPaths, smartPaths, appConfig] = await Promise.all([
-    listFiles(dir),
-    requiredFiles({ cwd: dir, entryPoints }),
-    getLinkedAppConfig(dir).catch(() => ({})),
-  ]);
+    const filenames = ['package.json', 'definition.json'];
+    for (const name of filenames) {
+      const absPath = path.resolve(workingDir, name);
+      zip.file(absPath, { name, mode: 0o644 });
+    }
```
eliangcs, line 340: `zapierwrapper.js` and `index.js` can be symlinks, but `package.json` and `definition.json` can't. Our upload endpoint validates them and doesn't seem to handle symlinks. So we're copying physical files here.

packages/cli/src/utils/build.js
```diff
@@ -468,43 +629,27 @@ const _buildFunc = async ({
...
-
-    // tries to do a reproducible build at least
-    // https://content.pivotal.io/blog/barriers-to-deterministic-reproducible-zip-files
-    // https://reproducible-builds.org/tools/ or strip-nondeterminism
-    await runCommand(
-      'find',
-      ['.', '-exec', 'touch', '-t', '201601010000', '{}', '+'],
-      { cwd: tmpDir },
-    );
-  }
```
eliangcs, line 497: The old code did this _after_ building the zip, which didn't make sense. If you wanted to build a reproducible zip, you should do this _before_ creating the zip, right? I think this code has been redundant all this time. I've removed it from the new code.

packages/cli/src/tests/utils/build.js
```diff
@@ -63,29 +60,33 @@ describe('build (runs slowly)', function () {
...
 
     [
       'safe.js',
+      '.DS_Store',
       '.env',
       '.environment',
-      '.git/HEAD',
-      'build/the-build.zip',
+      `.git${path.sep}HEAD`,
+      '.DS_Store',
```
szchenghuang: Oops, `.DS_Store` is added twice.

packages/cli/src/utils/files.js
```diff
@@ -200,19 +201,52 @@ const makeTempDir = () => {
...
+
+// Iterates files and symlinks in a directory recursively, up to a specified
+// number of levels deep (maxLevels). The minimum value for maxLevels is 1.
+// Yields fs.Dirent objects.
+function* walkDirLimitedLevels(dir, maxLevels, currentLevel = 1) {
+  const entries = fs.readdirSync(dir, { withFileTypes: true });
+  for (const entry of entries) {
+    if (entry.isDirectory()) {
+      if (currentLevel < maxLevels) {
+        const subDir = path.join(entry.parentPath, entry.name);
```
szchenghuang, line 226: `parentPath` is a property added to `fs.Dirent` in [Node.js ≥ v18.20](https://nodejs.org/api/fs.html#direntparentpath). But the package.json still keeps  `"engines": { "node": ">=18.18" }`. Should it be updated accordingly?

packages/cli/src/constants.js
szchenghuang: Awesome!

packages/cli/src/utils/files.js
```diff
@@ -200,19 +201,52 @@ const makeTempDir = () => {
...
+      yield entry;
+    }
+  }
+}
+
+// Iterates files and symlinks in a directory recursively, up to a specified
+// number of levels deep (maxLevels). The minimum value for maxLevels is 1.
+// Yields fs.Dirent objects.
+function* walkDirLimitedLevels(dir, maxLevels, currentLevel = 1) {
+  const entries = fs.readdirSync(dir, { withFileTypes: true });
```
szchenghuang, line 222: Not a blocker. Have you thought about pulling `fs.readdirAsync`? It seems a good alternative to make `walkDirLimitedLevels` async. So is `walkDir`.  `zapier build` could run faster in the end.

packages/cli/src/utils/ignore.js
```diff
@@ -1,20 +1,40 @@
...
+];
+const BLOCKLISTED_SUFFIXES = ['.gitkeep', '.env', '.environment', '.zip'];
 
-const constants = require('../constants');
-const isBlocklisted = (filePath) => {
-  return constants.BLOCKLISTED_PATHS.find((excluded) => {
-    return filePath.search(excluded) === 0;
-  });
+// Tells if we should exclude a file from build.zip or source.zip
+const isBlocklisted = (relPath) => {
```
szchenghuang, line 18: ### suggestion

`.some` is a built-in feature to replace the for loops:

```
const isBlocklisted = (relPath) => {
  if (BLOCKLISTED_PREFIXES.some(prefix => relPath.startsWith(prefix))) {
    return true;
  }
  if (BLOCKLISTED_SUFFIXES.some(suffix => relPath.endsWith(suffix))) {
    return true;
  }
  return false;
};
```

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
-    return constants.BLOCKLISTED_PATHS.find((excluded) => {
-      return filePath.search(excluded) === 0;
-    });
-  };
+// From a file path relative to workingDir, traverse up the directory tree until
+// it finds a directory that looks like a package directory, which either
+// contains a package.json file or whose path matches a pattern like
+// 'node_modules/(@scope/)package-name'.
+// Returns null if no package directory is found.
+const getPackageDir = (workingDir, relPath) => {
```
szchenghuang, line 96: This is a common use case and there're packages for that such as [package-up](https://www.npmjs.com/package/package-up) and [find-up](https://www.npmjs.com/package/find-up) which handle edge cases. Is there a reason to roll our own?

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
+  const nm = `node_modules${path.sep}`;
+  let i = relPath.lastIndexOf(nm);
+  if (i < 0) {
+    let dir = path.dirname(relPath);
+    for (let j = 0; j < 100; j++) {
+      const packageJsonPath = path.resolve(workingDir, dir, 'package.json');
+      if (fs.existsSync(packageJsonPath)) {
```
kola-er, line 103:
```suggestion
      if (await fse.pathExists(packageJsonPath)) {
```
Being async, it is non-blocking for better performance. This requires adding `async` to the `getPackageDir` function definition.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
 };
 
-const writeZipFromPaths = (dir, zipPath, paths) => {
-  return new Promise((resolve, reject) => {
-    const output = fs.createWriteStream(zipPath);
-    const zip = archiver('zip', {
-      zlib: { level: Z_BEST_COMPRESSION },
-    });
+const looksLikeWorkspaceRoot = async (dir) => {
+  if (fs.existsSync(path.join(dir, 'pnpm-workspace.yaml'))) {
```
kola-er, line 224:
```suggestion
  if (await fse.pathExists(path.join(dir, 'pnpm-workspace.yaml'))) {
```
For non-blocking IO performance

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
+    return true;
+  }
 
-    // listen for all archive data to be written
-    output.on('close', function () {
-      resolve();
-    });
+  const packageJsonPath = path.join(dir, 'package.json');
+  if (!fs.existsSync(packageJsonPath)) {
```
kola-er, line 232:
```suggestion
  const packageJsonPath = path.resolve(dir, 'package.json');
  if (! await fse.pathExists(packageJsonPath)) {
```
For non-blocking IO performance

kola-er, line 229:
```suggestion
  if (! await fse.pathExists(packageJsonPath)) {
```
For non-blocking IO performance

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
+    return false;
+  }
 
-    zip.on('error', function (err) {
-      reject(err);
-    });
+  let packageJson;
+  try {
+    packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
```
kola-er, line 235:
```suggestion
    packageJson = await fse.readJson(packageJsonPath);
```
For non-blocking IO performance

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
-    zip.finalize();
-  });
+const getNearestNodeModulesDir = (workingDir, relPath) => {
+  if (path.basename(relPath) === 'package.json') {
+    const nmDir = path.resolve(
+      workingDir,
+      path.dirname(relPath),
+      'node_modules',
+    );
+    return fs.existsSync(nmDir) ? path.relative(workingDir, nmDir) : null;
```
kola-er, line 268:
```suggestion
    return await fse.pathExists(nmDir) ? path.relative(workingDir, nmDir) : null;
```
For non-blocking IO performance

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
-const makeZip = async (dir, zipPath, disableDependencyDetection) => {
-  const entryPoints = [path.resolve(dir, 'zapierwrapper.js')];
+const writeBuildZipDumbly = async (workingDir, zip) => {
+  for (const entry of walkDirWithPresetBlocklist(workingDir)) {
+    const absPath = path.resolve(entry.parentPath, entry.name);
+    const relPath = path.relative(workingDir, absPath);
+    if (entry.isFile()) {
+      zip.file(absPath, { name: relPath });
+    } else if (entry.isSymbolicLink()) {
+      const target = path.relative(entry.parentPath, fs.realpathSync(absPath));
```
kola-er, line 292:
```suggestion
      const target = path.relative(entry.parentPath, await fs.realpath(absPath));
```
For non-blocking IO performance. It would be good to wrap it in try...catch to handle errors.

packages/cli/src/utils/build.js
```diff
@@ -71,171 +70,396 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
+    for (const symlink of symlinks) {
+      const absPath = path.resolve(
+        workingDir,
+        symlink.parentPath,
+        symlink.name,
+      );
+      const nameInZip = path.relative(workspaceRoot, absPath);
+      const targetInZip = path.relative(
+        symlink.parentPath,
+        fs.realpathSync(absPath),
```
kola-er, line 387:
```suggestion
        await fs.realpath(absPath),
```
For non-blocking IO performance

packages/cli/src/utils/build.js
```diff
@@ -290,129 +500,83 @@ const _buildFunc = async ({
...
-    constants.PLATFORM_PACKAGE,
-  );
-  if (!fs.existsSync(corePath)) {
-    throw new Error(
-      'Could not install dependencies properly. Error log:\n' + output.stderr,
-    );
+    // `npm install` may fail silently without returning a non-zero exit code,
+    // need to check further here
+    const corePath = path.join(workingDir, 'node_modules', PLATFORM_PACKAGE);
+    if (!fs.existsSync(corePath)) {
```
kola-er, line 609:
```suggestion
    if (! await fse.pathExists(corePath)) {
```
For non-blocking IO performance

packages/cli/src/utils/build.js
```diff
@@ -290,129 +500,83 @@ const _buildFunc = async ({
...
   );
-  const fileWriteError = await writeFile(
-    path.join(tmpDir, 'definition.json'),
-    prettyJSONstringify(rawDefinition),
-  );
 
-  if (fileWriteError) {
-    debug('\nFile Write Error:\n', fileWriteError, '\n');
+  try {
+    fs.writeFileSync(
```
kola-er, line 632:
```suggestion
    await fs.writeFile(
```

packages/cli/src/utils/ignore.js
eliangcs: I use `for` because it's faster and doesn't really hurt code readability much. This is a function that will be called many times, so it's worth applying a bit of optimization.

For reference, ChatGPT helped me write [this script](https://gist.github.com/eliangcs/071782755ea7b6bb92a7f25aaf6e49d7) to benchmark `for` vs. `.some`. The result:

```
Array size:  10,000,000
Rounds:      5
Target pos:  last (9999999)

Round 1: for=6.613 ms   Array.some=38.822 ms
Round 2: for=5.507 ms   Array.some=41.534 ms
Round 3: for=4.467 ms   Array.some=39.862 ms
Round 4: for=3.990 ms   Array.some=39.944 ms
Round 5: for=3.983 ms   Array.some=39.914 ms

Averages:
  for loop   → 4.912 ms
  Array.some → 40.015 ms
```

packages/cli/src/utils/files.js
eliangcs: Good catch! Updated in d62c03a20f2dd867a359c980a0e205857d32ed17.

packages/cli/src/tests/utils/build.js
eliangcs: Removed in 20c5bc58217df3c9ac0f753435b44d708c458cfe. Thanks!

packages/cli/src/utils/build.js
eliangcs: I didn't know about package-up or find-up, so thanks for pointing them out! But what I need here is pretty specific, as mentioned in the comment:

> traverse up the directory tree until it finds a directory that looks like a package directory, which either contains a package.json file or whose path matches a pattern like 'node_modules/(@scope/)package-name'

So even if we used package-up, we'd still need to write the second half of the logic ourselves (matching `node_modules/(@scope/)package-name`). That small bit of code doesn’t seem worth pulling in another dependency (which itself brings along more dependencies).

Also, looking at their code, I don't see what edge cases [package-up](https://github.com/sindresorhus/package-up/blob/5230f5c41ca316d0c0d73faf7b50b4788546294c/index.js) or [find-up](https://github.com/sindresorhus/find-up/blob/f9d7aafb15947e0eccf940247daaa04cf746f697/index.js) will help us handle.

packages/cli/src/utils/build.js
eliangcs: @kola-er do you mean wrap the `realpath` call or the entire function body in `try ... catch`? And what error(s) do you expect us to handle?

eliangcs:
@szchenghuang @kola-er thanks for the review! Regarding the async I/O suggestion: in general, yes, it's a good practice to use Promise-based async functions instead of their synchronous counterparts. However, simply replacing `readdirSync`, `realpathSync`, etc., with their async versions doesn't automatically make the code faster in this case. To actually benefit from async I/O performance, we'd need to use `Promise.all()` at some higher level to increase concurrency — which `zapier build` doesn't currently do in this PR.

This PR is primarily focused on adding support for npm/yarn/pnpm workspaces. Improving performance would be a bonus, but it's not the main goal here. What I'm looking for is feedback on the correctness of the logic and any potential bugs. So I'd appreciate it if we could focus the review on those aspects. How does that sound?

packages/cli/src/tests/utils/build.js
```diff
@@ -472,42 +434,33 @@ describe('build in workspaces', function () {
...
     uuidPackageJson.version.should.equal('8.3.2');
-
-    // Make sure node_modules/app-1 and node_modules/app-2 are not included
-    // in the build
-    fs.existsSync(
-      path.join(unzipPath, 'node_modules', 'app-1'),
-    ).should.be.false();
-    fs.existsSync(
-      path.join(unzipPath, 'node_modules', 'app-2'),
-    ).should.be.false();
```
kola-er, line 491: It seems this isn't applicable anymore because of the symlinks to the actual `node_modules` that would be in each workspace. Right?

kola-er: Oh, I think it is because of the `skipNpmInstall: true`. Right? Does that mean `zapier build` works with workspaces only when `--skip-npm-install` is used? If so, is it documented?

kola-er:
Noted your pushback on my suggestions for performance improvement. I agree that the review should focus on the support being added considering how big the changes are. It is a new area for me so it took me some time to go through the changes. I'm assuming that you have tested it extensively and the build works well on Lambda. Left a comment just for my knowledge. Approving anyways.

kola-er (dismissed):
Noted your pushback on my suggestions for performance improvement. I agree that the review should focus on the support being added considering how big the changes are. It is a new area for me so it took me some time to go through the changes. I'm assuming that you have tested it extensively and the build works well on Lambda. Left a comment just for my knowledge. Approving anyways.

packages/cli/src/utils/files.js
szchenghuang: Perf!

packages/cli/src/utils/build.js
szchenghuang: Yep, additional logic might still be needed because they are intentionally general-purpose utilities, but at the core `find_up`, for example, is robust in a few ways:

1. `relPath.indexOf(path.sep, i)` works for most platforms but can fail subtly with odd path structures like UNC paths on Windows (\\\\server\share). `find-up` uses `path.normalize` (via `locate-path` and `path-exists`) and internal checks to robustly support Windows-specific behavior (like drive letters, UNC paths).
2. `getPackageDir` takes `workingDir` and `relPath`, but it assumes that `relPath` is relative and `workingDir` is a known constant. `find-up` automatically normalizes the starting point and performs upward traversal regardless of current working directory context, making it more flexible and reusable across modules.
3. `getPackageDir` limits the traversal to 100 iterations, which is good for preventing infinite loops. `find-up` uses directory boundary detection (/ or root detection) which is safer and context-aware, e.g., doesn’t rely on arbitrary magic numbers.

Where it shines especially is that it can function in async, which correlates to my other feedback.

packages/cli/src/utils/ignore.js
szchenghuang: I like how functional programming favors declarative constructs over imperative constructs in JS.

The optimization feels premature in terms of performance. 40ms vs. 4ms on an array of 10,000,000 entries is little and atypical in our use case.

Not a blocker though. It's not the hill to die on.

szchenghuang (approved):

packages/cli/src/utils/build.js
eliangcs: Note that `getPackageDir` is an internal function of the module. It's not reused elsewhere. So what you see as as a drawback (being specific rather than general-purpose) is what I see as an advantage. Since we know the `relPath` argument passed to `getPackageDir` is always a relative path, we can make some assumptions and write more efficient code that does less. We don't need it to handle cases like, for example, paths starting with `/` or  a drive letter.

Regarding:

> find-up uses directory boundary detection (/ or root detection) which is safer and context-aware

`getPackageDir` actually does both: a fixed upper bound of 100 iterations _and_ the root detection via `path.dirname(dir) === dir` [here](https://github.com/zapier/zapier-platform/blob/d62c03a20f2dd867a359c980a0e205857d32ed17/packages/cli/src/utils/build.js#L111). On this regard, `getPackageDir` is actually even safer. The fixed 100-iteration cap is a practice I learned from [NASA's coding rules](https://en.wikipedia.org/wiki/The_Power_of_10:_Rules_for_Developing_Safety-Critical_Code), specifically the second rule "All loops must have fixed bounds."

packages/cli/src/utils/ignore.js
eliangcs: Yes, we shouldn't optimize too early, but only if doing so requires some effort. Writing `for` takes me no more time than writing `some`. And it will save a bit of CPU time on users' machines, so why not? Given equal effort and code readability, I'd choose performance over personal programming preference.

packages/cli/src/tests/utils/build.js
eliangcs: Suppose we have a project using npm/yarn workspaces:

```
(workspace root)
  package.json
  packages/
    app-1/
      package.json
      index.js
      node_modules/  (app-1's deps)
      ...
    app-2/
      package.json
      index.js
      node_modules/  (app-2's deps)
      ...
  node_modules/  (hoisted deps)
    app-1 -> ../packages/app-1
    app-2 -> ../packages/app-2
    zapier-platform-core/
      ...
```

Notice that npm/yarn created symlinks, `node_modules/app-1` and `node_modules/app-2` pointing to each workspace, `packages/app-1` and `packages/app-2`.

Previously, running `zapier build --skip-npm-install` in `packages/app-1` would create a zip like this:

```
package.json  (app-1's package.json)
index.js
zapierwrapper.js
definition.json
node_modules/  (app-1's deps + hoisted deps)
  zapier-platform-core/
    package.json
    ...
  lodash/
    ...
```

Previously we pack the files from `package/app-1` and grab some hoisted deps from `(workspace root)/node_modules` and copy to `(zip root)/node_modules`.

So in the previous test code, we asserted that `node_modules/app-1` and `node_modules/app-2` shouldn't exist in the zip.

Now, with this PR, running `zapier build --skip-npm-install` in `packages/app-1` will create a zip like this:

```
package.json  (app-1's package.json)
index.js -> packages/app-1/index.js
zapierwrapper.js -> packages/app-1/zapierwrapper.js
definition.json
packages/
  app-1/
    package.json
    index.js
    zapierwrapper.js
    ...
    node_modules/  (app-1's deps)
      ...
node_modules/  (hoisted deps)
  app-1 -> ../packages/app-1
  zapier-platform-core/
    package.json
    ...
  lodash/
    ...
```

The new code packs from `(workspace root)` so the zip now actually contains the symlink `node_modules/app-1`. That's why I removed the two `fs.existsSync` assertions.


Regarding:

> Does that mean `zapier build` works with workspaces only when `--skip-npm-install` is used? If so, is it documented?

Both `zapier build` and `zapier build --skip-npm-install` _will_ work with workspaces. To clarify, here are tables comparing behavior before and after:

Before:

Command | npm/yarn workspace | pnpm workspace
---|---|---
`zapier build` | ✅ | ✅ 
`zapier build --skip-npm-install` | ❗ (limited support) | ❌ (not supported)

After:

Command | npm/yarn workspace | pnpm workspace
---|---|---
`zapier build` | ✅ | ✅ 
`zapier build --skip-npm-install` | ✅ | ✅

kola-er: Thanks for the clarification @eliangcs

packages/cli/src/utils/build.js
```diff
@@ -71,171 +64,398 @@ const requiredFiles = async ({ cwd, entryPoints }) => {
...
+      '../test/userapp',
+      'zapier-platform-core/src/http-middlewares/before/sanatize-headers', // appears in zapier-platform-legacy-scripting-runner/index.js
+      './request-worker', // appears in zapier-platform-legacy-scripting-runner/zfactory.js
+      './xhr-sync-worker.js', // appears in jsdom/living/xmlhttprequest.js
+    ],
     format,
-    conditions,
+    // Setting conditions to an empty array to exclude 'module' condition,
+    // which Node.js doesn't use. https://esbuild.github.io/api/#conditions
+    conditions: [],
```
eliangcs, line 76: This fixes [PDE-6245](https://zapierorg.atlassian.net/browse/PDE-6245). For more context, see this [Slack message](https://zapier.slack.com/archives/C04P5J5R8E4/p1751015786041659?thread_ts=1750384204.158269&cid=C04P5J5R8E4).

packages/cli/package.json
```diff
@@ -46,6 +46,7 @@
     "@oclif/plugin-not-found": "3.2.51",
     "@oclif/plugin-version": "2.2.28",
     "adm-zip": "0.5.16",
+    "decompress": "4.2.1",
```
eliangcs, line 49: `decompress` was originally a dev dependency. I moved it to the `dependencies` section because now we're using it to extract build.zip in the "Testing build" step. The existing `adm-zip` package doesn't support extracting symlinks in a zip file, so I use `decompress` instead.

eliangcs (commented):
I pushed a commit with two changes:
- A additional "Testing build" step. This step extracts the build.zip file and run `node zapierwrapper.js` and `node index.js`. This step effectively verifies the build.zip file includes all the necessary dependent files.
- Setting `conditions: []` regardless of CJS or ESM when calling `esbuild.build`: This fixes [PDE-6245](https://zapierorg.atlassian.net/browse/PDE-6245).

@kola-er can you take another look? I'll run a few more tests before merging this PR.

kola-er (approved):

packages/cli/src/utils/build.js
eliangcs: cc @kreddlear as this changed your earlier change.

PR #1055, opened 2025-06-12
chore(core): PDE-6220 bump fernet to latest

Natay:
Bump to [latest fernet](https://www.npmjs.com/package/fernet) to resolve
```
node_modules/crypto-js
  fernet  <=0.3.1 || >=0.4.0
  Depends on vulnerable versions of crypto-js
  node_modules/fernet
    zapier-platform-core  >=17.2.0
    Depends on vulnerable versions of fernet
    node_modules/zapier-platform-core 
```

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -55,7 +55,7 @@
     "@zapier/secret-scrubber": "^1.1.2",
     "content-disposition": "0.5.4",
     "dotenv": "16.5.0",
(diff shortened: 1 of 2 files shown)


Natay:
> Oh [interesting](https://www.npmjs.com/package/fernet?activeTab=versions), so they went to 0.4.0 and then back to 0.3.2 and then up to 0.3.3 😅

That confused me too!

PR #1062, opened 2025-06-27
[PDE-6273] Defensive when content-type is absent

szchenghuang:
This PR defensively handles HTTP responses without the `content-type`.

### DIFF
--- a/packages/core/src/tools/fetch-logger.js
+++ b/packages/core/src/tools/fetch-logger.js
@@ -56,11 +56,16 @@ const isZapierUserAgent = (headers) =>
   _.get(headers, 'user-agent', []).indexOf('Zapier') !== -1;
 
 const shouldIncludeResponseContent = (contentType) => {
+  if (!contentType) {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


packages/core/src/tools/fetch-logger.js
```diff
@@ -56,11 +56,16 @@ const isZapierUserAgent = (headers) =>
   _.get(headers, 'user-agent', []).indexOf('Zapier') !== -1;
 
 const shouldIncludeResponseContent = (contentType) => {
+  if (!contentType) {
+    return false;
```
szchenghuang, line 60: @eliangcs Before I merge this..does `false` make more sense than `true` for you?

Natay: IMO `false` makes sense, so we don't return the response content if we don't know the content-type.

eliangcs: +1. `false` makes sense. We don't want to log content for an unknown content-type.

PR #1064, opened 2025-07-02
[PDE-6175] request-worker added to bundle while require.resolve warning suppressed

szchenghuang:
This PR re-adds `request-worker` to the bundle built by ESBuild while the warning of `require.resolve` suppressed.

### DIFF
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -70,6 +70,7 @@ const findRequiredFiles = async (workingDir, entryPoints) => {
     external: [
       '../test/userapp',
       'zapier-platform-core/src/http-middlewares/before/sanatize-headers', // appears in zapier-platform-legacy-scripting-runner/index.js
+      './xhr-sync-worker.js', // appears in jsdom/living/xmlhttprequest.js
     ],
     format,
     // Setting conditions to an empty array to exclude 'module' condition,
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/build.js
```diff
@@ -64,11 +64,12 @@ const findRequiredFiles = async (workingDir, entryPoints) => {
     platform: 'node',
     metafile: true,
     logLevel: 'warning',
+    logOverride: {
+      'require-resolve-not-external': 'silent',
+    },
```
kola-er, line 69: I think we should not silence the `require-resolve-not-external` warnings. We can always ignore the warnings, but we don't want to miss a "crucial" warning that needs to be marked as external.

kola-er: Argh...I see how untidy it looks for DX.

kola-er: Sounds like a fair trade-off to silent it.

packages/cli/src/utils/build.js
```diff
@@ -64,11 +64,12 @@ const findRequiredFiles = async (workingDir, entryPoints) => {
...
     external: [
       '../test/userapp',
       'zapier-platform-core/src/http-middlewares/before/sanatize-headers', // appears in zapier-platform-legacy-scripting-runner/index.js
-      './request-worker', // appears in zapier-platform-legacy-scripting-runner/zfactory.js
-      './xhr-sync-worker.js', // appears in jsdom/living/xmlhttprequest.js
```
kola-er, line 71: This is actually external - I mean from 3rd-party package. I think we can leave this.

PR #1066, opened 2025-07-03
PDE-6289 - Add user and account filters to canary

standielpls:
To improve the usability of `zapier canary`, I am looking to add several "audience" type target filters to help isolate canary tests.
The general percentage applies the change to all users, with no way to isolate any problematic invocations to test out a bug fix for instance. 
This PR then aims to add the following filters:
- user (email): to filter by user, can be used with accountId
- accountId: to filter by account

### DIFF
--- a/packages/cli/docs/cli.md
+++ b/packages/cli/docs/cli.md
@@ -55,11 +55,11 @@ This command is typically followed by `zapier upload`.
 
 Only one canary can be active at the same time. You can run `zapier canary:list` to check. If you would like to create a new canary with different parameters, you can wait for the canary to finish, or delete it using `zapier canary:delete a.b.c x.y.z`.
 
-For migrating a specific user, you can pass in the --user flag. 
+To canary traffic for a specific user, use the --user flag.
 
-For migrating a specific user within a specific account, you can pass in the --user flag with the --accountId flag.
+To canary traffic for a specific user within a specific account, use both --user and --accountId flags.
 
-For migrating a specific account, you can pass in the --accountId flag with an --owner flag. This is the only use case for the --owner flag, as it helps isolate the account filter.
+To canary traffic for an entire account, use both --accountId and --owner flags. The --owner flag is only used with --accountId to isolate the account filter.
 
 Note: this is similar to `zapier migrate` but different in that this is temporary and will "revert" the changes once the specified duration is expired.
 
@@ -72,17 +72,16 @@ Note: this is similar to `zapier migrate` but different in that this is temporar
 **Flags**
 * (required) `-p, --percent` | Percent of traffic to route to new version
 * (required) `-d, --duration` | Duration of the canary in seconds
-* `-u, --user` | Canary this user across all accounts, unless account_id is specified.
-* `-o, --owner` | Canary the account_id. Only used when account_id is also used.
-* `-a, --accountId` | The account id to target. If owner is specified, Canary applies to all traffic for the account. If user is specified, only canary the user within this account.
+* `-u, --user` | Canary this user across all accounts, unless `accountId` is specified.
+* `-o, --owner` | Canary the accountId. Only used when `--accountId` is also used.
+* `-a, --accountId` | The account ID to target. If owner is specified, canary applies to all traffic for the account. If user is specified, only canary the user within this account.
 * `-d, --debug` | Show extra debugging output.
 
 **Examples**
-* `zapier canary:create 1.0.0 1.1.0 -p 25 -d 720`
-* `zapier canary:create 2.0.0 2.1.0 --percent 50 --duration 300`
-* `zapier canary:create 2.0.0 2.1.0 --percent 50 --duration 300 --user user@example.com`
-* `zapier canary:create 2.0.0 2.1.0 --percent 50 --duration 300 --account_id 123 --owner admin@example.com`
-* `zapier canary:create 2.0.0 2.1.0 --percent 50 --duration 300 --account_id 123 --user user@example.com`
+* `zapier canary:create 1.0.0 1.1.0 -p 10 -d 3600`
+* `zapier canary:create 2.0.0 2.1.0 --percent 25 --duration 1800 --user user@example.com`
+* `zapier canary:create 2.0.0 2.1.0 -p 15 -d 7200 -a 12345 -u user@example.com`
+* `zapier canary:create 2.0.0 2.1.0 -p 50 -d 600 -a 12345 -o admin@example.com`
 
 
 ## canary:delete
--- a/packages/cli/docs/cli.md
+++ b/packages/cli/docs/cli.md
@@ -72,8 +72,8 @@ Note: this is similar to `zapier migrate` but different in that this is temporar
 **Flags**
 * (required) `-p, --percent` | Percent of traffic to route to new version
 * (required) `-d, --duration` | Duration of the canary in seconds
-* `-u, --user` | Canary this user across all accounts, unless `accountId` is specified.
-* `-o, --owner` | Canary the accountId. Only used when `--accountId` is also used.
+* `-u, --user` | Canary this user (email) across all accounts, unless `accountId` is specified.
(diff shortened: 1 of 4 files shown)


packages/cli/docs/cli.md
```diff
@@ -64,8 +64,11 @@ Note: this is similar to `zapier migrate` but different in that this is temporar
 * (required) `versionTo` | Version to canary traffic to
 
 **Flags**
-* (required) `-p, --percent` | Percent of traffic to route to new version
+* `-p, --percent` | Percent of traffic to route to new version
 * (required) `-d, --duration` | Duration of the canary in seconds
+* `--user` | Specify a user within all accounts to canary
+* `--account` | Canary all accounts for which the specified user is a member
+* `--includeEnterprise` | Include enterprise accounts
```
FokkeZB: I suggest we drop this. I don't think (certainly external) builders should be able to canary enterprise users unless they specifically target (and thus know) a user or account.

```suggestion
```

packages/cli/docs/cli.md
```diff
@@ -64,8 +64,11 @@ Note: this is similar to `zapier migrate` but different in that this is temporar
 * (required) `versionTo` | Version to canary traffic to
 
 **Flags**
-* (required) `-p, --percent` | Percent of traffic to route to new version
+* `-p, --percent` | Percent of traffic to route to new version
 * (required) `-d, --duration` | Duration of the canary in seconds
+* `--user` | Specify a user within all accounts to canary
+* `--account` | Canary all accounts for which the specified user is a member
```
FokkeZB: I find this U/DX confusing. I would expect to canary a specific account, not all accounts that a user is member of. I also expect to be able to target a specific user within the context of a specific account.

How about:

- `zapier canary:create 1.0.0 2.0.0 -p 50` to canary 50% of all traffic except enterprise.
- `zapier canary:create 1.0.0 2.0.0 -p 50 -u user@example.com` to canary 50% of traffic from given user.
- `zapier canary:create 1.0.0 2.0.0 -p 100 -u user@example.com` to canary all traffic from given user.
- `zapier canary:create 1.0.0 2.0.0 -p 50 -a 123 -o owner@example.com` to canary 50% of traffic from given account owned by given user.
- `zapier canary:create 1.0.0 2.0.0 -p 100 -a 123 -o owner@example.com` to canary all traffic from given account owned by given user.
- `zapier canary:create 1.0.0 2.0.0 -p 50 -a 123 -u user@example.com` to canary 50% of traffic from given account, limited to given user.
- `zapier canary:create 1.0.0 2.0.0 -p 50 -a 123 -u user@example.com` to canary all traffic from given account, limited to given user.

We can't just accept `-o owner@example.com` as one user can own multiple accounts, but we also can't just accept `-a 123`, as it's an incremental ID and thus too easy to iterate over.

And `-p` IMO should always be required. Even when targeting a specific user and/or account, I should be able to target a slice of their traffic.

I also wonder if we should default to target free users only, and add a `--paid` flag to include paid users, while still excluding enterprise users. That way the default settings have the least possible impact on revenue.

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -25,13 +44,36 @@ If you would like to stop this canary now, run \`zapier canary:delete ${existing
       return;
     }
 
+    const body = {
+      duration,
+    };
+
+    if (user || account) {
+      this.startSpinner(
+        `Starting migration from ${versionFrom} to ${versionTo} for ${
```
FokkeZB:
```suggestion
        `Starting canary from ${versionFrom} to ${versionTo} for ${
```

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -25,13 +44,36 @@ If you would like to stop this canary now, run \`zapier canary:delete ${existing
...
+          user || account
+        }`,
+      );
+      body.email = user || account;
+      body.email_type = flagType;
+    } else {
+      this.startSpinner(
+        `Starting migration from ${versionFrom} to ${versionTo} for ${percent}%`,
```
FokkeZB:
```suggestion
        `Starting canary from ${versionFrom} to ${versionTo} for ${percent}%`,
```

FokkeZB (dismissed):
I believe the U/DX can and should be improved.

packages/cli/docs/cli.md
standielpls: I see, I was looking to preserve the behaviour of accepting emails over ids
But considering how limited that is, and how confusing the UDX is (agreed), ids are unavoidable. 

I don't quite understand the owner flag:
```
zapier canary:create 1.0.0 2.0.0 -p 50 -a 123 -o owner@example.com to canary 50% of traffic from given account owned by given user.
zapier canary:create 1.0.0 2.0.0 -p 100 -a 123 -o owner@example.com to canary all traffic from given account owned by given user.
```

Is it essentially an account canary? The way I read the example above is, 50% of traffic for account 123 but only if owner@example.com owns account 123?

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -8,6 +8,21 @@ class CanaryCreateCommand extends ZapierBaseCommand {
...
+    const owner = this.flags.owner;
+    const accountId = this.flags.accountId;
+
+    if (user && owner) {
+      this.error(
+        'Cannot specify both `--user` and `--owner`. Use only one or the other.',
+      );
+    }
+
+    if (accountId && !user && !owner) {
```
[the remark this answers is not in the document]
standielpls: This feedback considered this block of code in isolation. The block above ensures only one of `user` or `owner` is provided so it shouldn't violate what your feedback is considering. However, I'll optimize this to make it super clear what the valid filter combinations are.

packages/cli/docs/cli.md
standielpls: The owner field adds too much complexity. Discussed more [here](https://zapier.slack.com/archives/C04QN139NV6/p1754266215639389?thread_ts=1749001243.760169&cid=C04QN139NV6)

Natay (dismissed):
Thank you Stan! Just a heads up we might need to update the public facing docs to include these new features: https://docs.zapier.com/platform/build/test-tools#q-how-do-i-monitor-the-canary-test

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -71,6 +89,16 @@ CanaryCreateCommand.flags = buildFlags({
       description: 'Duration of the canary in seconds',
       required: true,
     }),
+    user: Flags.string({
+      char: 'u',
+      description:
+        'Canary this user (email) across all accounts, unless `accountId` is specified.',
+    }),
+    accountId: Flags.string({
```
eliangcs: For naming consistency, can we use `--account-id` instead of `--accountId`? Otherwise, the code changes look good to me, assuming you've tested it.

eliangcs (dismissed):

packages/cli/src/oclif/commands/canary/create.js
```diff
@@ -71,6 +89,16 @@ CanaryCreateCommand.flags = buildFlags({
...
+      char: 'a',
+      description:
+        'The account ID to target. If user is specified, only canary the user within this account.',
```
FokkeZB: Should we document that only internally user can be omitted?

standielpls:
> Thank you Stan! Just a heads up we might need to update the public facing docs to include these new features: https://docs.zapier.com/platform/build/test-tools#q-how-do-i-monitor-the-canary-test

Yep I got a ticket queued up [PDE-6290](https://zapierorg.atlassian.net/browse/PDE-6290)!

standielpls:
> > Thank you Stan! Just a heads up we might need to update the public facing docs to include these new features: https://docs.zapier.com/platform/build/test-tools#q-how-do-i-monitor-the-canary-test
> 
> Yep I got a ticket queued up [PDE-6290](https://zapierorg.atlassian.net/browse/PDE-6290)!

Accompanying docs: 
https://gitlab.com/zapier/public-api/public-api-docs/-/merge_requests/138

PR #1067, opened 2025-07-03
feat(cli): Add additional Typescript auth options to `zapier init` [PDE-6191]

kreddlear:
We should offer more `zapier init` template options that support Typescript. This PR adds corresponding Typescript support for the auth templates, including:

- basic-auth
- custom-auth
- digest-auth
- oauth1
- oauth2
- session-auth

It implements this via the `--language` flag, so that devs can now run commands like `zapier init --template oauth2 --language typescript`.

This PR also removes the `typescript` and `typescript-esm` templates, since they can now be reproduced by running:
```bash
# create the authentication part with the oauth2 template
zapier init --template oauth2 --language typescript

# add a trigger - scaffold auto-detects typescript
zapier scaffold trigger contact

# add an action - scaffold auto-detects typescript
zapier scaffold create contact
```

Because of this, the `zapier build` tests also had to be updated, since there is no longer a `typescript` example app that uses CommonJS. The ESM `zapier build` tests were replaced with `oauth2-typescript` example app, since that uses ESM.

This MR touches a lot of files, but that's mostly adding example apps for each `-typescript` template, and deleting the original `typescript` standalone template. The actual changes occur in `packages/cli/src/`, in these files:

- [oclif/commands/init.js](https://github.com/zapier/zapier-platform/pull/1067/files#diff-6ff196fc5a4414c7bbda6239b0469d72f2d65ff893577161ffedad1509203d5f) - adds `language` flag to `init` command
- [generators/index.js](https://github.com/zapier/zapier-platform/pull/1067/files#diff-d1c4c0c707357d4dc03db588db62e696c939bd08e4ec76a607139d592247e171) - adds logic to take in `language` flag and route it as needed
- [utils/auth-files-codegen.js](https://github.com/zapier/zapier-platform/pull/1067/files#diff-da6dc4800666b5026dbad8f3981e08538b6e1367ae363b27fa3ad423937b64d6) - updates auth template generating functions to add logic to handle the `language` param
- [utils/codegen.js](https://github.com/zapier/zapier-platform/pull/1067/files#diff-63723557017b77ac6ae190dae588dee503e84745eeebb7b1b4edb4a1eddfd330) - adds Typescript-formatted functions to handle exporting a module and creating a typed object


kola-er (commented):
Just two observations/recommendations:

- I could see addition of typescript templates for Oauth2 and Basic Auth. Should we be adding for Custom Auth too because most of these AI integrations would likely be using API key.
- Having a new `zapier init` option `language` with choice list ['javascript', 'typescript'] may be confusing with the existing `typescript` choice of the option `template`. Would it be less confusing to build on the existing `typescript` value for the option `template`? With that, we would have the following new typescript auth templates; `typescript-basic-auth`, `typescript-custom-auth`, and `typescript-oauth2`. The existing template `typescript` would be for the general typescript template that has auth and actions. For example, `zapier init ./path/myapp --template typescript-oauth2` would generate for typescript Oauth2 integration.

packages/cli/scaffold/resource.template.ts
```diff
@@ -1,29 +1,50 @@
-import type { PerformFunction, Resource } from 'zapier-platform-core';
```
kreddlear, line 1: `PerformFunction` has been deprecated and replaced with types like `CreatePerform`, `PollingTriggerPerform`, and `SearchPerform`. These changes bring this file in line with the updates made to `create.template.ts` for example ([seen here](https://github.com/zapier/zapier-platform/blob/1b3eb1f12c284f11694453c47d3ef26af682cfaa/packages/cli/scaffold/create.template.ts)).

packages/cli/src/generators/index.js
```diff
@@ -67,19 +67,7 @@ const writeGenericPackageJson = (gen, packageJsonExtension) => {
   );
 };
 
-const writeTypeScriptPackageJson = (gen, packageJsonExtension) => {
-  const moduleExtension =
-    gen.options.module === 'esm'
```
kreddlear, line 72: The `module` option is now no longer available for Typescript templates. The new Typescript templates are only available in ESM format. This removes some functionality, but the original CommonJS Typescript template was somewhat clunky and compiled into very difficult-to-read JS, so it's very likely that Typescript devs will prefer ESM.

packages/cli/src/generators/index.js
```diff
@@ -205,36 +210,6 @@ const writeForStandaloneTemplate = (gen) => {
   );
 };
 
-const writeForStandaloneTypeScriptTemplate = (gen) => {
```
kreddlear, line 208: No longer needed now that the standalone `typescript` template has been removed.

packages/cli/src/tests/utils/build.js
```diff
@@ -270,15 +271,6 @@ describe('build (runs slowly)', function () {
       new Set(['index.js', 'README.md', '.zapierapprc', '.gitignore']),
     );
   });
-
-  it('should run the zapier-build script', async () => {
```
kreddlear, line 274: `zapier-build` script does not exist in non-Typescript example apps, like the `oauth2` one being used here.

packages/cli/src/utils/auth-files-codegen.js
```diff
@@ -45,84 +58,101 @@ const authFileExport = (
...
-            "The test method allows Zapier to verify that the credentials a user provides are valid. We'll execute this method whenever a user connects their account for the first time.",
-          ),
-          test,
-          comment(
-            `This template string can access all the data returned from the auth test. If you return the test object, you'll access the returned data with a label like \`{{json.X}}\`. If you return \`response.data\` from your test, then your label can be \`{{X}}\`. This can also be a function that returns a label. That function has the standard args \`(${standardArgs.join(
-              ', ',
-            )})\` and data returned from the test can be accessed in \`bundle.inputData.X\`.`,
-          ),
-          objProperty('connectionLabel', connectionLabel),
-        ),
```
kreddlear, line 70: Copy/pasted these lines into the `configProps` array above, to enable passing them into the `configObj` and dynamically generate either a Typescript object (with `satisfies Authentication` at the end) or a vanilla JS object.

packages/cli/src/utils/auth-files-codegen.js
```diff
@@ -45,84 +58,101 @@ const authFileExport = (
...
     comment(
       'You want to make a request to an endpoint that is either specifically designed to test auth, or one that every user will have access to. eg: `/me`.',
     ),
     comment(
       'By returning the entire request object, you have access to the request and response data for testing purposes. Your connection label can access any data from the returned response using the `json.` prefix. eg: `{{json.username}}`.',
     ),
-    fatArrowReturnFunctionDeclaration('test', standardArgs, zRequest(testUrl)),
+    fatArrowReturnFunctionDeclaration(
+      'test',
+      standardArgs(language),
```
kreddlear, line 108: `standardArgs` -> `standardArgs(language)` is the actual change here, since `standardArgs` has been changed from an array to a function.

packages/cli/src/utils/auth-files-codegen.js
```diff
@@ -131,17 +161,22 @@ const beforeMiddlewareFunc = (funcName, ...statements) =>
...
-const afterMiddlewareFunc = (funcName, ...statements) =>
+const afterMiddlewareFunc = (funcName, language, ...statements) =>
   block(
     comment(
       "This function runs after every outbound request. You can use it to check for errors or modify the response. You can have as many as you need. They'll need to each be registered in your index.js file.",
     ),
-    functionDeclaration(funcName, { args: afterMiddlewareArgs }, ...statements),
+    functionDeclaration(
+      funcName,
+      { args: afterMiddlewareArgs(language) },
```
kreddlear, line 171: Similar to above, `afterMiddlewareArgs` -> `afterMiddlewareArgs(language)` is the actual change here, the rest is formatting.

packages/cli/src/generators/templates/index.template.ts
```diff
@@ -1,24 +1,21 @@
...
+import zapier, { defineApp } from 'zapier-platform-core';
 
-import packageJson from '../package.json';
+import packageJson from '../package.json' with { type: 'json' };
 
-import MovieCreate from './creates/movie';
-import MovieTrigger from './triggers/movie';
-import authentication from './authentication';
-import { addBearerHeader } from './middleware';
+import authenticationModule from './authentication.js';
```
kola-er, line 5:
```suggestion
import authenticationModule from './authentication';
```

Should it be without the `.js` since it's for typescript?

packages/cli/src/generators/templates/authTests/basic.test.ts
```diff
@@ -0,0 +1,42 @@
+import { describe, expect, it } from 'vitest';
+import zapier from 'zapier-platform-core';
+
+import App from '../index.js';
```
kola-er, line 4: Similar to https://github.com/zapier/zapier-platform/pull/1067#discussion_r2213955287. And applies to other auth test templates for typescript.

packages/cli/src/generators/index.js
```diff
@@ -306,6 +288,25 @@ class ProjectGenerator extends Generator {
...
+        }
+        // if they try to combine typescript with commonjs, throw an error
+        if (this.options.module === 'commonjs') {
+          throw new Error('Typescript is not supported for commonjs');
+        }
+      }
+    } else {
+      // default to javascript for the language if it's not set
+      this.options.language = 'javascript';
+    }
```
kola-er: Seems this check should come before the check on line 283-286, else `--template basic-auth --module esm --language typescript` would throw an error.

packages/cli/src/generators/templates/tsconfig.template.json
```diff
@@ -0,0 +1,18 @@
...
+      "isolatedModules": true,
+      "noImplicitAny": false,
+      "skipLibCheck": true,
+      "outDir": "./dist",
+      "rootDir": "./src",
+      "strict": true
+    },
+    "include": ["./src/**/*.ts"],
+    "exclude": ["./**/*.test.ts"]
+  }
```
kola-er, line 18: Seems this is meant to replace the existing `tsconfig.json` in the same folder?

example-apps/basic-auth-typescript/src/index.ts
```diff
@@ -0,0 +1,21 @@
+import zapier, { defineApp } from 'zapier-platform-core';
+
+import packageJson from '../package.json' with { type: 'json' };
+
+import authenticationModule from './authentication.js';
```
kola-er, line 5: Same here and other examples

```suggestion
import authenticationModule from './authentication';
```

example-apps/oauth1-trello-typescript/src/authentication.ts
```diff
@@ -0,0 +1,114 @@
...
+// response data for testing purposes. Your connection label can access any data
+// from the returned response using the `json.` prefix. eg: `{{json.username}}`.
+const test = (z: ZObject, bundle: Bundle) =>
+  z.request({ url: 'https://auth-json-server.zapier-staging.com/me' });
+
+export default {
+  config: {
+    // OAuth2 is a web authentication standard. There are a lot of configuration
+    // options that will fit most any situation.
+    type: 'oauth2',
```
kola-er: This example is meant for Oauth1

packages/cli/src/generators/templates/tsconfig.template.json
kreddlear: Hmm, could you clarify this comment? This is meant to replace the `tsconfig.json` which lived in the `typescript` folder, since that whole folder has been deleted but we still need a template for `tsconfig.json`. Github doesn't show that very helpfully, it shows that it was [moved to the basic-auth-typescript example app here](https://github.com/zapier/zapier-platform/pull/1067/files#diff-7bd4839f12e83af3e80a2e79418f0b6086959d8ebfe943e91c136c0841a54726), but there shouldn't be another `tsconfig.json` in this folder 🤔

packages/cli/src/generators/templates/index.template.ts
kreddlear: This actually tripped me up quite a bit while working on this, using a `.ts` extension results in Typescript error, but it's not possible to just import `./authentication` because these new templates are EcmaScript modules, and ESM requires file extensions for imports.

I spent a while reading Typescript and Node docs and it seems like this is convention when using Typescript and ESM - internally we only have one app that uses both so far, but it also uses this pattern of importing with JS extensions.

Prompted by your comment though, I asked Thomas [in this thread](https://zapier.slack.com/archives/C07SVQPEE22/p1752790423772959?thread_ts=1752790004.987889&cid=C07SVQPEE22) if using JS imports is something that is indeed convention or if it's something we may change in the future. He confirmed this is convention and how these should be set up, so these imports don't need to be changed and are correct as `.js` 👍

example-apps/oauth1-trello-typescript/src/authentication.ts
kreddlear: Good catch 🤦‍♀️ Updated, and double-checked the rest of the generated apps to make sure they are correct.

packages/cli/src/generators/index.js
kreddlear: You're totally right. I switched them and added an extra check to the ESM if block to only handle `--language javascript`.

Here's the combos I tested, which all had the proper behavior - let me know if I've missed any:

- `--template basic-auth --module esm --language typescript`: Should succeed with no errors since ESM is compatible with Typescript and the template is Typescript-supported.
- `--template basic-auth --module esm --language javascript`: Should fail with an error that ESM is not supported for this template.
- `--template minimal --module esm --language typescript`: Should fail with an error that Typescript is not supported for this template.
- `--template minimal --module esm --language javascript`: Should succeed with no errors since `minimal` is explicitly ESM-supported.

kreddlear:
Thanks so much for the detailed review @kola-er , I really appreciate it 🙏 

The lack of unit tests made me somewhat worried as well. Before submitting this for review, I tested via the following:
- Re-run `zapier init` to re-populate all the existing templates that were affected by this change (basic, custom, digest, oauth1, oauth2, session), and confirmed there were no major changes for each (the `.gitignore` would change for each to ignore `dist/`, but I discarded those changes as they aren't actually relevant for non-Typescript apps - but won't hurt the apps to be included in the future.)
- Confirmed `zapier build` and `zapier test` still work for the existing example apps that were affected
- Generated the new templates using `zapier init`
- Confirmed `zapier build` and `zapier test` work for both language type templates as well

The smoke tests for `zapier init` did help me catch some issues, so hopefully that plus the above bullets will cover most testing cases. If you can think of other tests to run, please let me know, happy to do so.

Please let me know if these updates help resolve the issues you called out!

packages/cli/src/generators/templates/index.template.ts
kola-er: Looks confusing, thanks for confirming that.

packages/cli/src/generators/templates/tsconfig.template.json
kola-er: That's it. I missed it. Thanks

kola-er (approved):
This is a huge change, thanks for taking care of it.

PR #1068, opened 2025-07-08
PDE-6286 fix(cli): `zapier build` misses local package files when building in a lerna monorepo

eliangcs:
Fixes a bug where `zapier build --skip-npm-install` failed to recognize a Lerna monorepo. In which case, it should build the app at the project root rather than the app directory.

For example, for a Lerna monorepo that looks like this:

```
(project root)
  lerna.json
  package.json
  common/
    libary-1/
      package.json
  packages/
    app-1/
      package.json
      ...
```

`zapier build --skip-npm-install` should build at the project root instead of `packages/app-1`. When it builds in `packages/app-1`, like it does currently, `common/library-1` will be missing in the output build.zip file.

The fix is simple. We only need to add a few lines to make `looksLikeWorkspaceRoot()` return true when there's a `lerna.json` file.

While I was there, I added test coverage not only for the Lerna case, but also for yarn and pnpm cases.

### DIFF
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -223,6 +223,10 @@ const looksLikeWorkspaceRoot = async (dir) => {
     return true;
   }
 
+  if (fs.existsSync(path.join(dir, 'lerna.json'))) {
+    return true;
+  }
+
   const packageJsonPath = path.join(dir, 'package.json');
   if (!fs.existsSync(packageJsonPath)) {
     return false;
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/build.js
```diff
@@ -67,11 +67,7 @@ const findRequiredFiles = async (workingDir, entryPoints) => {
     logOverride: {
       'require-resolve-not-external': 'silent',
     },
-    external: [
-      '../test/userapp',
-      'zapier-platform-core/src/http-middlewares/before/sanatize-headers', // appears in zapier-platform-legacy-scripting-runner/index.js
-      './xhr-sync-worker.js', // appears in jsdom/living/xmlhttprequest.js
```
eliangcs, line 73: I misused the `external` option of `esbuild.build()`. We should include these files if they are imported, so we shouldn't mark them "external". `'require-resolve-not-external': 'silent'` is the right way to silence the warning on `require.resolve`. (thanks @szchenghuang)

PR #1070, opened 2025-07-09
PDE-6286 fix(cli,core): improve error messages on `zapier build`

eliangcs:
Two changes:

- Hide the temporary directory in the "Testing build" step in `zapier build`
- Improve the error message when zapier-platform-core fails to import the app's ESM entry point

### Hide the temporary directory in the "Testing build" step in `zapier build`

In the "Testing build" step, `zapier build` extracts build.zip to a new [temporary directory](https://github.com/zapier/zapier-platform/blob/fecdd0a2e7b0610ca4be437717407aff50decdc2/packages/cli/src/utils/build.js#L519) before [running](https://github.com/zapier/zapier-platform/blob/fecdd0a2e7b0610ca4be437717407aff50decdc2/packages/cli/src/utils/build.js#L537) `node zapierwrapper.js`.

If there's an error, the error message may contain the path to the temporary directory, which may cause confusion. Integration developers shouldn't need to know the "Testing build" step was using a temporary directory internally. It'd be friendlier if the command should just print where the problem is, relative to the current working directory (i.e. the app directory).

With this PR, we hide the path to the temporary directory from the error message.

### Improve the error message when ZPC fails to import the app's ESM entry point

The `zapierwrapper.mjs` entry point prints this suggestion when it [fails to import the app's entry point](https://github.com/zapier/zapier-platform/blob/fecdd0a2e7b0610ca4be437717407aff50decdc2/packages/core/include/zapierwrapper.mjs#L8):

> Make sure you specify a valid entry point using `exports` in package.json.

Based on the reports from some developers, this suggestion might be misleading in some cases, because the [import error](https://github.com/zapier/zapier-platform/blob/fecdd0a2e7b0610ca4be437717407aff50decdc2/packages/core/include/zapierwrapper.mjs#L7) might be caused by someone else instead of the app module itself.

So this PR adds another condition to at least make sure the app's module name is in the error message before giving such suggestion. Also, the suggestion is moved from the end of the message to the beginning, so it's more visible. And, the suggestion is revised so it's more clear:

> It seems you are using ESM because your package.json has `"type": "module"`. For ESM to work, make sure you specify a valid entry point using `exports` (instead of `main`) in package.json.

### DIFF
--- a/packages/cli/src/utils/build.js
+++ b/packages/cli/src/utils/build.js
@@ -554,7 +554,7 @@ const testBuildZip = async (zipPath) => {
             `in your ${colors.bold.underline('.zapierapprc')} file.`,
         );
       } else if (error.message) {
-        // Hide the unzip tempoary directory
+        // Hide the unzipped temporary directory
         error.message = error.message
           .replaceAll(`file://${testDir}/`, '')
           .replaceAll(`${testDir}/`, '');


packages/cli/src/utils/build.js
```diff
@@ -553,6 +553,11 @@ const testBuildZip = async (zipPath) => {
             `You may have to add it to ${colors.bold.underline('includeInBuild')} ` +
             `in your ${colors.bold.underline('.zapierapprc')} file.`,
         );
+      } else if (error.message) {
+        // Hide the unzip tempoary directory
```
kola-er:
```suggestion
        // Hide the unzipped temporary directory
```

eliangcs: Fixed in [f0812c0](https://github.com/zapier/zapier-platform/pull/1070/commits/f0812c0143b921b7386b4bbb7c16b5487803356c).

kola-er (approved):

PR #1073, opened 2025-07-14
refactor(cli): `zapier init` for auth should have befores/afters in middleware, not in auth files

kreddlear:
While working on https://github.com/zapier/zapier-platform/pull/1067, it came to my attention that the auth templates currently add `befores` and `afters` (middleware) to the `authentication.js` file. This goes against convention in our Integrations teams, and also against convention suggested by the `middleware` init template itself.

This PR splits those out into their own `middleware` file, and should be finished and merged after https://github.com/zapier/zapier-platform/pull/1067 is merged.


packages/cli/src/generators/index.js
```diff
@@ -130,11 +130,11 @@ const authTypes = {
...
   const authType = authTypes[gen.options.template];
   const content = authFilesCodegen[authType](gen.options.language);
-  const destPath =
-    gen.options.language === 'typescript'
-      ? 'src/authentication.ts'
-      : 'authentication.js';
-  gen.fs.write(gen.destinationPath(destPath), content);
+  const destPath = (key) =>
+    gen.options.language === 'typescript' ? `src/${key}.ts` : `${key}.js`;
+  Object.entries(content).forEach(([key, value]) => {
```
kreddlear, line 135: Now that `authFilesCodegen[authType](gen.options.language)` returns this:

```
{
    middleware: fileContents,
    authentication: fileContents,
}
```

instead of just a string of file contents, we need to create both `src/middleware.ts`/`middleware.js` and `src/authentication.ts`/`authentication.js`.

Natay (approved):

PR #1082, opened 2025-07-16
feat(core, schema): Add search pagination support

kreddlear:
As part of releasing Search Pagination support, we need to:
- Allow `canPaginate` to be set for `BasicSearchOperationSchema`
- Validation if `canPaginate` is true:
    - Require an object like `{ results: [...], paging_token: '123' }` to be returned from a search if `canPaginate` is `true`
    - Ensure that the `paging_token` is a string (even an empty one is okay) or `null`
    - Ensure that the included `results` are an array
- Preserve existing validation to confirm non-paging search actions still return an array
- Type updates: Allow an object to be returned in the `SearchPerform` type


packages/core/src/checks/index.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   createIsObject: require('./create-is-object'),
 
-  searchIsArray: require('./search-is-array'),
+  searchIsArrayOrObject: require('./search-is-array'),
```
FokkeZB:
```suggestion
  searchIsArrayOrEnvelope: require('./search-is-array-or-envelope'),
```

packages/core/src/checks/search-is-array.js
```diff
@@ -5,19 +5,53 @@ const { simpleTruncate } = require('../tools/data');
...
+
+    if (hasCanPaginate(searchKey, compiledApp)) {
+      // must be an object
+      if (!_.isObject(results)) {
+        return [
+          `Paginating search results must be an object, got: ${typeof results}, (${truncatedResults})`,
+        ];
+      }
+      // must have results and paging_token
+      if (!_.has(results, 'results') || !_.has(results, 'paging_token')) {
```
FokkeZB: Do we also need to check that `paging_token` is a string or `null` (or not/undefined?) and that `results` is an `Array` (of objects)?

packages/core/src/checks/search-is-array.js
```diff
@@ -5,19 +5,53 @@ const { simpleTruncate } = require('../tools/data');
...
+        return [
+          `Paginating search results must be an object containing results and paging_token, got: ${truncatedResults}`,
```
FokkeZB:
```suggestion
          `Paginated search results must be an object containing results and paging_token, got: ${truncatedResults}`,
```

packages/core/types/functions.d.ts
```diff
@@ -130,7 +134,7 @@ export type CreatePerformGet<
 export type SearchPerform<
   $InputData extends DefaultInputData = DefaultInputData,
   $Return extends {} = {},
-> = (z: ZObject, bundle: Bundle<$InputData>) => $Return[] | Promise<$Return[]>;
+> = (z: ZObject, bundle: Bundle<$InputData>) => $Return[] | $Return | Promise<$Return[] | $Return>;
```
FokkeZB: Shouldn't this be:

```suggestion
> = (z: ZObject, bundle: Bundle<$InputData>) => $Return[] | { results: $Return[], paging_token: string } | Promise<$Return[] | { results: $Return[], paging_token: string }>;
```

Likely would benefit from a helper type to be less verbose.

packages/schema/lib/schemas/BasicSearchOperationSchema.js
```diff
@@ -30,6 +30,11 @@ BasicSearchOperationSchema.properties = {
       'How will Zapier get a single record? If you find yourself reaching for this - consider resources and their built-in get methods.',
     oneOf: [{ $ref: RequestSchema.id }, { $ref: FunctionSchema.id }],
   },
+  canPaginate: {
+    description:
+      'Does this endpoint support pagination via temporary cursor storage?',
```
FokkeZB:
```suggestion
      'Does this search support pagination?',
```

FokkeZB (commented):
I know you're still working on this, but left some comments while I'm here.

packages/cli/src/oclif/commands/invoke.js
```diff
@@ -1378,6 +1379,10 @@ InvokeCommand.flags = buildFlags({
       description:
         'EXPERIMENTAL: Instead of using the local .env file, use the production authentication data with the given authentication ID (aka the "app connection" on Zapier). Find them at https://zapier.com/app/assets/connections (https://zpr.io/z8SjFTdnTFZ2 for instructions) or specify \'-\' to interactively select one from your available authentications. When specified, the code will still run locally, but all outgoing requests will be proxied through Zapier with the production auth data.',
     }),
+    'paging-token': Flags.string({
+      description:
+        'Set bundle.meta.paging_token. Used for search pagination or bulk reads. When used in production, this indicates which page of items you should fetch. First page is 0.',
```
FokkeZB: The "First page is 0." bit only applies to `bundle.meta.page`.

```suggestion
        'Set bundle.meta.paging_token. Used for search pagination or bulk reads. When used in production, this indicates which page of items you should fetch.',
```

kreddlear: 🤦‍♀️ Good catch. Cursor kept adding this, I missed it the last time before committing.

packages/core/src/checks/search-is-array.js
kreddlear: I'm setting `results = results.results` so that the existing check for an array can take care of that 👍 Re: `paging_token`, yes, I've added a check for `paging_token` to be a string.

packages/core/src/checks/search-is-array-or-envelope.js
```diff
@@ -0,0 +1,60 @@
...
+    if (hasCanPaginate(searchKey, compiledApp)) {
+      // if paging is supported and results is an object (indicating pagination), it must have results and paging_token
+      if (_.isPlainObject(results)) {
+        if (!_.has(results, 'results') || !_.has(results, 'paging_token')) {
+          return [
+            `Paginated search results must be an object containing results and paging_token, got: ${truncatedResults}`,
+          ];
+        }
+        if (
+          !_.isString(results.paging_token) &&
```
Natay, line 35: Should we validate if `paging_token` is an empty string as well or is that more of a responsibility on the product side?

kreddlear: Hmm, interesting question 🤔 looks like `_.isString()` returns `true` for an empty string, so that should be covered here.

I tested this with my test integration and a review lab, and it looks like if the integration returns `paging_token` as `null` or an empty string, this check (correctly) does not trigger an error (~although the invoke results don't contain a paging token at all, which I think is expected~ Updated this so that the paging token is returned from invoke regardless of whether a paging token is provided in the initial call). So, I think this should be all good, but let me know if there's something in there I haven't thought of!

Natay: That makes sense! I hadn't realized paging_token got omitted from the results when it's an empty string ( thereby behaving the same as other other null values ) 👍

packages/core/types/functions.d.ts
kreddlear: Updated this to the proper types, and added a helper type [here](https://github.com/zapier/zapier-platform/pull/1082/commits/acada57fa4ac7ed998bec5d35937cbd11cb3d80a#diff-919ea6742a9ba5b605c793cb47e7c553ee662d83ee82fbb1fbbc278e81260d57) to make this more readable!

PR #1085, opened 2025-07-17
PDE-6323: improve decryptBundleWithSecret

Natay:

### DIFF
--- a/packages/core/src/tools/bundle-encryption.js
+++ b/packages/core/src/tools/bundle-encryption.js
@@ -2,7 +2,7 @@
 
 const crypto = require('crypto');
 const fernet = require('fernet');
-
+const zlib = require('zlib');
 /**
  * Decrypt a bundle using secret key
  *
@@ -10,10 +10,12 @@ const fernet = require('fernet');
  * 1. Hash the secret with SHA256 to get 32 bytes
  * 2. Base64url encode those bytes to make Fernet-compatible key
  * 3. Use Fernet library to decrypt (handles all token parsing internally)
+ * 4. Base64 decode the decrypted string to get compressed binary data
+ * 5. Decompress the data using gzip
  *
  * @param {string} bundle - The bundle represented as an encrypted token
  * @param {string} secret - The secret key for decryption
- * @returns {Object} The decrypted bundle object
+ * @returns {Object} The decrypted and decompressed bundle object
  */
 const decryptBundleWithSecret = (bundle, secret) => {
   try {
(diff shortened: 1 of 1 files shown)


packages/core/src/tools/bundle-encryption.js
```diff
@@ -26,32 +28,57 @@ const decryptBundleWithSecret = (bundle, secret) => {
...
+    // Step 4: The data is compressed, so we need to decompress it
+    let decompressed;
+    try {
+      // Try to decompress first (for new format with compression)
+      decompressed = zlib.gunzipSync(compressedBytes).toString('utf8');
+    } catch (decompressionError) {
+      // If decompression fails, assume it's the old format without compression
+      // This provides backward compatibility
+      decompressed = decryptedString;
+    }
```
standielpls: What happens if there's a real decompression error?

standielpls: I was going to suggest, since `bundle` is always going to be in json, we could just assert it's json parsable, but we already do that in the next step, so this should be fine?

packages/core/test/app-middleware.js
```diff
@@ -335,5 +336,203 @@ describe('app middleware', () => {
...
+            },
+          },
+        })),
+        pagination: {
+          page: 1,
+          limit: 50,
+          total: 1000,
+          hasMore: true,
+        },
+      };
```
standielpls: In theory this can be any payload to encrypt/compress, but for the sake of how it is used, and how others may use tests to understand how something works, can we use a mock bundle instead? and for example, jack up the rawRequest etc..

packages/core/test/app-middleware.js
```diff
@@ -335,5 +336,203 @@ describe('app middleware', () => {
...
+
+    it('should decrypt and set stashed bundle with new compressed format', async () => {
+      const testSecret = 'test-secret-key';
+      const testData = {
+        compressed: 'data'.repeat(500), // Large data that benefits from compression
+        nested: {
+          values: [1, 2, 3, 4, 5],
+          metadata: { created: '2023-01-01', type: 'test' },
+        },
+      };
```
standielpls: same with using a "bundle" test data suggestion

packages/core/test/tools/bundle-encryption.test.js
```diff
@@ -168,5 +169,167 @@ describe('bundle-encryption', () => {
...
+            },
+          },
+        })),
+        pagination: {
+          page: 1,
+          limit: 100,
+          total: 1000,
+          hasMore: true,
+        },
+      };
```
standielpls: same with using test "bundle" data suggestion

packages/core/src/tools/bundle-encryption.js
Natay: > but we already do that in the next step, so this should be fine?

That was my line of thinking as well, the one caveat is that real decompression errors would manifest  themselves  as json.parse errors, we could add a console.log here so we can at least spot it in the logs, i'll add that!

packages/core/test/app-middleware.js
Natay: Good point, I created a test fixture for bundles and a method that make it easy to ratchet up the size of the whole bundle.

packages/core/src/tools/bundle-encryption.js
```diff
@@ -26,32 +28,61 @@ const decryptBundleWithSecret = (bundle, secret) => {
...
+
+    // Step 4: The data is compressed, so we need to decompress it
+    let decompressed;
+    try {
+      // Try to decompress first (for new format with compression)
+      decompressed = zlib.gunzipSync(compressedBytes).toString('utf8');
+    } catch (decompressionError) {
+      // If decompression fails, assume it's the old format without compression
+      // This provides backward compatibility
+      console.warn(
```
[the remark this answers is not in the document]
Natay: I've found `console.log` used in server spots ([1](https://github.com/zapier/zapier-platform/blob/de14a28d90399d18536df8ec122b5a6b4dd63729/packages/core/src/tools/create-logger.js#L118), [2](https://github.com/zapier/zapier-platform/blob/de14a28d90399d18536df8ec122b5a6b4dd63729/packages/core/src/tools/memory-checker.js#L29) ) that run in production code and are used for debugging purposes. I believe this is okay to add taking that into consideration, @standielpls what do you think?

Natay:
Thanks for the feedback!

>  because I was on the fence about that. But to double down, using a large bundle is not super useful since it bloats the env where tests run, and we get the same result by asserting some reduction in data due to compression with a smaller test size.

Good point! I saw the RSS memory didn't change much locally so I ratchetd up the number, but you and copilot make very good points

> My other general feedback is, IMO, the comments are overly detailed and bloats the readability of the code. For example:

 I’ll make sure to trim comments that are redundant or don’t add much clarity going forward 👍 This flow ( stashing, decrytping, and decompressing ) is not straightforward so I figured more explicit comments will benefit a future devs seeing the code for the first time.

PR #929, opened 2024-12-11
[PDE-4089] chore(cli, core, schema): Release v16.0.0

rnegron:
## 16.0.0

_released `2024-12-16`_

Version `16.0.0` is a breaking change release that contains some important housekeeping updates that put us in a better spot to support some exciting changes coming up in future versions. **Here is a quick breakdown of the two main breaking changes:**

**1. The minimum supported version of Node.js is now v18.18.** We’ve officially dropped support for Node.js v16.x, which has been designated [end-of-life since September 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol).

**2. Integrations using `zapier-platform-core@16.0.0` and onwards will default to running on the ARM architecture in AWS Lambda.**  Note: If you believe your integrations have run into runtime incompatibilities due to this change, please [reach out to our support team](https://developer.zapier.com/contact) and we may be able to opt your integration out of running on ARM.

Here are the detailed release notes for this release (**note that ❗ denotes a breaking change**):

### cli

- :hammer: The `oclif` suite of packages have been upgraded to support the latest `oclif` major version (v4) ([#890](https://github.com/zapier/zapier-platform/pull/890)).

### core

- **:exclamation: Node.js v16.x is no longer supported ([#890](https://github.com/zapier/zapier-platform/pull/890))**

- **:exclamation: Integrations running on `zapier-platform-core` v16 and onwards will run on the ARM architecture by default.** This means means they will run on AWS Graviton2 processors (previously, the AWS Lambda runtime was based on x86 architecture). For more details on what this means on a technical level, please check out: [AWS Lambda on Graviton 2](https://github.com/aws/aws-graviton-getting-started/blob/main/aws-lambda/README.md#aws-lambda-on-graviton2).

- **:exclamation: Due to changes in the underlying `dotenv` library ([see changelog](https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md)) ([#926](https://github.com/zapier/zapier-platform/pull/926)), when using `zapier.tools.env.inject()`, we always recommend "double quoting" your environment variable values.** Please make sure to consider if the following cases would apply to your `.env` files:
  * ```SECRET_HASH=something-with-a-#-hash```: change it to ```SECRET_HASH="something-with-a-#-hash"```
  * ```SECRET_BACKTICK=something-with-a-`-backtick```: change it to ```SECRET_BACKTICK="something-with-a-`-backtick"```
[...]

PR #1086, opened 2025-07-18
fix(cli): PDE-6346 remove empty array at the end of zapier versions -f json

Natay:
https://zapierorg.atlassian.net/browse/PDE-6346

`zapier versions -f json` returns an extra array with three objects, this fixes it so they show what was intended ( error types ). I am questioning if we should remove them all together though? 

```
➜  ZapierWebhookAppCopy zapier versions -f json 
✔ Loading versions
[
  {
    "Version": "1.0.19",
    "Platform": "17.2.0",
    "Zap Users": "0",
    "State": "private",
    "Legacy Date": "",
    "Deprecation Date": "null",
    "Timestamp": "2025-05-19T13:19:26+00:00"
  }
]
[
  {},
  {},
  {}
]
```

Looks like this after this PR

### DIFF
--- a/packages/cli/src/oclif/commands/versions.js
+++ b/packages/cli/src/oclif/commands/versions.js
@@ -1,4 +1,3 @@
-const colors = require('colors/safe');
 const { Flags } = require('@oclif/core');
 const BaseCommand = require('../ZapierBaseCommand');
 const { buildFlags } = require('../buildFlags');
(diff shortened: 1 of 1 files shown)


eliangcs:
The Error Types table only makes sense in the [`zapier validate`](https://github.com/zapier/zapier-platform/blob/808d884a497a563fa130a58fde813c7fef91af73/packages/cli/src/oclif/commands/validate.js#L88) command. It seems the table was accidentally copied to `zapier versions` in this [PR](https://github.com/zapier/zapier-platform/pull/929). @Natay I think you're right. We should remove it from `zapier versions`.

PR #1097, opened 2025-07-30
PDE-5447 fix(core): some authData keys are safe to be logged uncensored

eliangcs:
Currently, we censor every value appearing in `bundle.authData` in the log unless the value [looks like a URL without a secret](https://github.com/zapier/zapier-platform/blob/0784942e39981c33ba35bbc74bb21d86727c96da/packages/core/src/tools/create-logger.js#L151-L153). However, there are many fields in `bundle.authData` that are known to be safe to log uncensored, such as `scope`, `team_id`, and `account`. Censoring them in logs just makes troubleshooting hard.

### DIFF
--- a/packages/core/src/tools/create-logger.js
+++ b/packages/core/src/tools/create-logger.js
@@ -51,7 +51,6 @@ const initSafeAuthDataKeys = () => {
     'environment',
     'location',
     'org',
-    'org',
     'organization',
     'project',
     'region',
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/create-logger.js
```diff
@@ -32,16 +32,68 @@ const LOG_STREAM_BYTES_LIMIT = 15 * 1024 * 1024;
...
   }
+  const commonProtocols = [
+    'https://',
+    'http://',
+    'ftp://',
+    'ftps://',
+    'file://',
+  ];
+  for (const protocol of commonProtocols) {
+    if (value.startsWith(protocol)) {
```
eliangcs, line 86: I added `if (value.startsWith(protocol))` because I found `new URL('chat:write,chat:read')` actually passes and it's not a URL at all.

PR #1106, opened 2025-08-12
PDE-6144 ci(cli): enable Windows in CI

eliangcs:


packages/cli/package.json
```diff
@@ -25,16 +25,16 @@
     "docs": "ZAPIER_BASE_ENDPOINT='' node scripts/docs.js",
     "preversion": "git pull && yarn validate",
     "prepack": "oclif manifest",
-    "postpack": "rm -f oclif.manifest.json",
+    "postpack": "rimraf oclif.manifest.json",
```
eliangcs, line 28: `rm` isn't available on Windows. Changed it to cross-platform `rimraf`.

packages/cli/package.json
```diff
@@ -25,16 +25,16 @@
...
     "lint": "eslint src",
     "lint:fix": "eslint --fix src",
-    "test": "cross-env NODE_ENV=test mocha -t 50s --recursive src/tests --exit",
-    "test:debug": "cross-env NODE_ENV=test node inspect ../../node_modules/.bin/mocha -t 50s --recursive src/tests --exit",
-    "smoke-test": "cross-env NODE_ENV=test mocha -t 2m --recursive src/smoke-tests --exit",
-    "smoke-test:debug": "cross-env NODE_ENV=test node inspect ../../node_modules/.bin/mocha -t 2m --recursive src/smoke-tests --exit",
+    "test": "cross-env NODE_ENV=test mocha -t 200s --recursive src/tests --exit",
+    "test:debug": "cross-env NODE_ENV=test node inspect ../../node_modules/.bin/mocha -t 200s --recursive src/tests --exit",
+    "smoke-test": "cross-env NODE_ENV=test mocha -t 6m --recursive src/smoke-tests --exit",
+    "smoke-test:debug": "cross-env NODE_ENV=test node inspect ../../node_modules/.bin/mocha -t 6m --recursive src/smoke-tests --exit",
```
eliangcs, line 37: GitHub Actions's Windows runners are [terribly slow](https://github.com/actions/runner-images/issues/7320), so I increased the time limits to avoid timeouts.

.github/workflows/ci.yaml
```diff
@@ -32,9 +32,9 @@ jobs:
         node-version: [18.x, 20.x]
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
kreddlear, line 35: Looks like for both `actions/checkout@v4` and `actions/setup-node@v4`, the default runtime changes from Node 16 to Node 20 ([example changelog](https://github.com/actions/setup-node/releases/tag/v4.0.0)) - I'm assuming that's okay because we're using `setup-node` to manual setting the node version to either 18 or 20 below, is that right?

packages/cli/package.json
kreddlear: Yikes, that's unfortunate. Sounds like a good call.

packages/cli/src/smoke-tests/smoke-tests.js
```diff
@@ -357,7 +358,8 @@ describe('smoke tests - setup will take some time', function () {
             'trigger',
             'people',
             '-i',
-            '{"species_id":"1"}',
+            // Windows's cmd.exe somehow requires you to escape quotes in JSON
+            IS_WINDOWS ? '{\\"species_id\\":\\"1\\"}' : '{"species_id":"1"}',
```
kreddlear, line 362: Apologies for a possibly obvious question - why wouldn't this be `'{\"species_id\":\"1\"}'`? Do we also need to escape the backslash itself?

eliangcs: Here we're executing Windows's cmd.exe from Node.js. cmd.exe then executes the `zapier invoke` command in another a Node.js process.

Before the string goes to cmd.exe, it's `'{\"species_id\":\"1\"}'` (backslashes are actually in the string). Then it appears cmd.exe would unescape that string into `'{"species_id":"1"}'` before passing it to the Node.js process that executes `zapier invoke`.

.github/workflows/ci.yaml
eliangcs: That's right. The default Node.js version used in the image doesn't matter. We use the one specified in in `matrix.node-verison` anyway.

PR #1143, opened 2025-08-29
fix(schema): Refactor InputFieldGroupsSchema to have its property displayed

kola-er:
**BEFORE**

**AFTER**


packages/cli/src/tests/utils/build.js
```diff
@@ -1446,7 +1446,7 @@ describe('build ESM (runs slowly)', function () {
     smartPaths.filter((p) => p.endsWith('.ts')).length.should.equal(0);
     smartPaths.should.not.containEql('tsconfig.json');
 
-    smartPaths.length.should.be.within(200, 327);
+    smartPaths.length.should.be.within(200, 328);
```
kola-er, line 1449: Don't know why this was failing.

Natay: I think it's because of the newly added [schema file](https://github.com/zapier/zapier-platform/pull/1143/files#diff-9bc213d72ad82baca1fc0db2ccc8e1360e904e97e5cc616361753f976fc05713), I've noticed this happen with previous schema changes

kola-er: Odd that schema files affect builds in packages/cli/tests. I'm sure of the schema/tests/index.js [here](https://github.com/zapier/zapier-platform/pull/1143/files#diff-2962a33ff09947d011d42b132e48e362a25138f5284ab3fb514fac5e98dff658R12).

Natay: Yeah I suspect it's because this test checks the build size and that includes the new schema file in this PR. I have seen similar oddities when [updating a dependency](https://github.com/zapier/zapier-platform/pull/1096/files) in core/, cause this test in cli/ to fail

Natay (approved):

PR #1147, opened 2025-09-02
PDE-6429 fix(core): censor sensitive info in `ResponseError`

eliangcs:
To reproduce, use this example trigger and `beforeRequest` middleware:

```js
// Trigger definition
{
  key: "recipe",
  operation: {
    perform: {
      url: "https://httpbin.zapier-tooling.com/status/400",
    },
  },
  // ...
}
```

```js
// beforeRequest middleware
const includeAccessToken = (request, z, bundle) => {
  if (bundle.authData.access_token) {
    request.headers.Authorization = `Bearer ${bundle.authData.access_token}`;
    request.params.access_token = bundle.authData.access_token;
  }
  return request;
};
```

Then run it with `zapier invoke trigger recipe --debug`.

### Before

The value of `access_token` is visible in the response error:

```
(redacted)
 ›   Error: {"status":400,"headers":{"content-type":"text/plain; charset=utf-8","retry-after":null},"content":"","request":{"url":"https://httpbin.zapier-tooling.com/status/400?access_token=a_token"}}
```

(The value of `access_token` is not censored in the response error)

### After

The value of `access_token` is censored in the response error:

```
(redacted)
 ›   Error: {"status":400,"headers":{"content-type":"text/plain; charset=utf-8","retry-after":null},"content":"","request":{"url":"https://httpbin.zapier-tooling.com/status/400?access_token=:censored:7:da1d3a47b7:"}}
```

### DIFF
--- a/packages/cli/docs/cli.md
+++ b/packages/cli/docs/cli.md
@@ -937,12 +937,14 @@ Run the standard validation routine powered by json-schema that checks your inte
 
 **Flags**
 * `--without-style` | Forgo pinging the Zapier server to run further checks.
+* `--skip-build` | Skip running the _zapier-build script before validation.
 * `-d, --debug` | Show extra debugging output.
 * `-f, --format` | Change the way structured data is presented. If "json" or "raw", you can pipe the output of the command into other tools, such as jq. One of `[plain | json | raw | row | table]`. Defaults to `table`.
 
 **Examples**
 * `zapier validate`
 * `zapier validate --without-style`
+* `zapier validate --skip-build`
 * `zapier validate --format json`
 
 
--- a/packages/core/src/tools/secret-scrubber.js
+++ b/packages/core/src/tools/secret-scrubber.js
@@ -1,0 +1,98 @@
+'use strict';
+
+const { recurseExtract } = require('@zapier/secret-scrubber');
+const {
+  isUrlWithSecrets,
+  isSensitiveKey,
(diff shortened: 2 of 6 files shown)


packages/core/src/tools/secret-scrubber.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+  return false;
+};
+
+const findSensitiveValuesFromAuthData = (authData) =>
+  recurseExtract(authData, (key, value) => {
+    // for the most part, we should censor all the values from authData
+    // the exception is safe urls, which should be filtered out - we want those to be logged
+    // but, we _should_ censor-no-matter-what sensitive keys, even if their value is a safe url
+    // this covers the case where someone's password is a valid url ¯\_(ツ)_/¯
```
standielpls, line 83: lol

packages/cli/docs/cli.md
```diff
@@ -937,12 +937,14 @@ Run the standard validation routine powered by json-schema that checks your inte
 
 **Flags**
 * `--without-style` | Forgo pinging the Zapier server to run further checks.
+* `--skip-build` | Skip running the _zapier-build script before validation.
```
standielpls, line 940: Did you manually update this or was this a pre-commit hook change?

eliangcs: It was a pre-commit hook. Looks like it was missed in a previous PR.

PR #1153, opened 2025-09-05
 feat(schema): add skipCleanArrayInputData option to actions ( PDE-6533 )

Natay:
This pull request adds a new experimental property, `skipCleanArrayInputData `, to operation schemas. This allows users to specify whether Zapier should automatically clean up **array** input data by removing null, undefined, or empty values. 

Schema Files Updated
- Added `skipCleanArrayInputData` to packages/schema/lib/schemas/BasicPollingOperationSchema.js
- Added `skipCleanArrayInputData` to packages/schema/lib/schemas/BasicActionOperationSchema.js
- Added `skipCleanArrayInputData` to packages/schema/lib/schemas/BasicSearchOperationSchema.js
- Added `skipCleanArrayInputData` to packages/schema/lib/schemas/BasicHookOperationSchema.js
- Added `skipCleanArrayInputData` to packages/schema/lib/schemas/BasicHookToPollOperationSchema.js

In the coming major version, we will replace this experimental flag with `skipCleanInputData`.

### DIFF
--- a/packages/schema/lib/schemas/BasicOperationSchema.js
+++ b/packages/schema/lib/schemas/BasicOperationSchema.js
@@ -69,9 +69,9 @@ module.exports = makeSchema(
           'Zapier uses this configuration to apply throttling when the limit for the window is exceeded.',
         $ref: ThrottleObjectSchema.id,
       },
-      skipCleanInputData: {
+      skipCleanArrayInputData: {
         description:
-          'Should empty strings, `null`, `undefined`, and empty Arrays or objects be removed from `inputData`?',
+          '(Experimental) Should empty strings, `null`, `undefined`, and empty Arrays or objects be removed from `inputData`?',
         type: 'boolean',
       },
     },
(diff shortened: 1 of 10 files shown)


eliangcs (approved):
Can we label this flag as "experimental" in the description? Just in case we want to remove it without releasing a major version. Otherwise, the PR looks good to me!

packages/schema/lib/schemas/BasicOperationSchema.js
```diff
@@ -69,6 +69,11 @@ module.exports = makeSchema(
           'Zapier uses this configuration to apply throttling when the limit for the window is exceeded.',
         $ref: ThrottleObjectSchema.id,
       },
+      skipCleanInputData: {
+        description:
+          'Should Zapier clean inputs by removing empty strings, None, empty lists, and dicts before passing them to the perform function?',
```
FokkeZB:
```suggestion
          'Should empty strings, `null`, `undefined`, and empty Arrays or objects be removed from `inputData`?',
```

FokkeZB: Let's not use Python language here. And it doesn't only apply to `perform`, so let's focus on `inputData`.

eliangcs (approved):
Thanks! I'll include this in v17.7.1.

PR #1093, opened 2025-07-24
feat(schema, cli) Support version with label [PDE-6384]

FokkeZB:
Adds support for `version` in `package.json` that has a label (e.g. `1.2.3-dev`).

PR #1155, opened 2025-09-08
feat(cli): Only allow standard (non-labeled) versions to be migrated or promoted to [PDE-6496]

kreddlear:
This other PR: https://github.com/zapier/zapier-platform/pull/1093 adds support for labeled versions, i.e. `2.1.3-beta`. However, we only want to allow those during development - actual production versions or versions that users will use should be standard numbers-only versions, i.e. `2.1.3`.

This PR adds a small utility function that checks the version format when `zapier promote` or `zapier migrate` is run, and throws an error if the version being migrated/promoted to is not numbers-only.


eliangcs:
We probably should include this in [v17.7.1](https://github.com/zapier/zapier-platform/pull/1154). So I'll wait a bit on this before releasing. @kreddlear

PR #1160, opened 2025-09-15
fix(cli): Allow deleting old non-semver versions

standielpls:
In https://github.com/zapier/zapier-platform/pull/1093, we released a change to introduce a stricter semver. However, before that takes into effect, we have limited the ability to delete non-strict semver versions. This MR releases that check for the version delete CLI command


FokkeZB (approved):

standielpls:
> @standielpls should `deprecate` also be allowed? As we don't allow deleting (available) versions before deprecating them first?

Yep, `deprecate` already doesn't validate the version when it reads the argument.
Was able to deprecate just now https://cdn.zappy.app/5845859c08e6617db94fae6be1c2da8b.png

PR #1169, opened 2025-09-24
PDE-6592 docs(cli): fix incorrect docs in `zapier migrate`

eliangcs:


eliangcs:
Some tests are failing because https://httpbin.zapier-tooling.com/get somehow now responds with `url: http://httpbin.zapier-tooling.com/get` instead of `url: https://httpbin.zapier-tooling.com/get` (`http://` vs. `https://`). Let's not "fix" the tests for now. Let's see if our httpbin will fix itself first.

PR #1172, opened 2025-10-01
feat(cli) - Add natural snapshot support for zapier push

standielpls:
https://zapierorg.atlassian.net/browse/PDE-6609

Zapier push currently takes the package.json version to push to. This makes for a poor DX for creating snapshots because we need to change and then revert. Instead, we can add a option to push to a snapshot label (akin to pushing to a branch that isn't the one you're checked out on in git).

### DIFF
--- a/packages/cli/src/oclif/commands/push.js
+++ b/packages/cli/src/oclif/commands/push.js
@@ -14,7 +14,7 @@ class PushCommand extends ZapierBaseCommand {
     const definition = await localAppCommand({ command: 'definition' });
 
     const snapshotLabel = this.flags.snapshot;
-    if (snapshotLabel && snapshotLabel.length >= 18) {
+    if (snapshotLabel && snapshotLabel.length > 18) {
       throw new Error('Snapshot label cannot exceed 18 characters');
     }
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/oclif/commands/push.js
```diff
@@ -12,7 +12,15 @@ class PushCommand extends ZapierBaseCommand {
   async perform() {
     const skipNpmInstall = this.flags['skip-npm-install'];
     const definition = await localAppCommand({ command: 'definition' });
-    const version = definition.version;
+
+    const snapshotLabel = this.flags.snapshot;
+    if (snapshotLabel && snapshotLabel.length >= 18) {
```
FokkeZB: 18 is fine, and also matches the error.

```suggestion
    if (snapshotLabel && snapshotLabel.length > 18) {
```

FokkeZB (changes requested):
>= 18 should be > 18

packages/cli/src/oclif/commands/push.js
standielpls: 🤦 lol

FokkeZB (approved):

Natay (approved):

PR #1180, opened 2025-10-14
fix(cli): Display failure reasons for env:set

kola-er:

### DIFF
--- a/packages/cli/src/oclif/commands/env/set.js
+++ b/packages/cli/src/oclif/commands/env/set.js
@@ -73,12 +73,12 @@ class SetEnvCommand extends BaseCommand {
       const failedKeys = e.json.errors[0].split('update: ')[1].split(', ');
       const successfulResult = omit(payload, failedKeys);
       if (!Object.keys(successfulResult).length) {
-        this.error(e.json.errors[0]);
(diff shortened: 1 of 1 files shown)


packages/cli/src/oclif/commands/env/set.js
```diff
@@ -73,12 +73,12 @@ class SetEnvCommand extends BaseCommand {
       const failedKeys = e.json.errors[0].split('update: ')[1].split(', ');
       const successfulResult = omit(payload, failedKeys);
       if (!Object.keys(successfulResult).length) {
-        this.error(e.json.errors[0]);
+        this.error(e.json.errors.join('\nError: '));
```
kola-er, line 76: **Before**




**After**

packages/cli/src/oclif/commands/env/set.js
```diff
@@ -73,12 +73,12 @@ class SetEnvCommand extends BaseCommand {
...
       }
 
       this.warn(successMessage(version));
       this.logJSON(successfulResult);
-      this.warn(`However, these keys failed to update: ${failedKeys}`);
+      this.warn(e.json.errors.join('\nWarning: '));
```
kola-er, line 81: **Before**




**After**

PR #1181, opened 2025-10-15
PDE-6485 feat(cli): add executable name `zapier-platform` and deprecate `zapier`

eliangcs:
This PR:

- adds `zapier-platform` as an additional executable name to the CLI tool
- deprecates the `zapier` executable name with a deprecation warning (it still works)

What it will look like:

```
$ zapier help
The CLI for managing integrations in Zapier Developer Platform.

(irrelevant output content...)

[DEPRECATION WARNING] The executable name "zapier" is deprecated and will be remove in a future version. Please use "zapier-platform" instead.
```

### DIFF
--- a/packages/cli/src/bin/run-deprecated
+++ b/packages/cli/src/bin/run-deprecated
@@ -8,7 +8,7 @@ const path = require('node:path');
 
   console.warn(
     '[DEPRECATION WARNING] ' +
-      'The executable name "zapier" is deprecated and will be remove in a future version. ' +
+      'The executable name "zapier" is deprecated and will be removed in a future version. ' +
       'Please use "zapier-platform" instead.',
   );
 })();


packages/cli/src/bin/run.cmd
```diff
@@ -1,3 +0,0 @@
-@echo off
-
-node "%~dp0\run" %*
```
eliangcs, line 3: This `.cmd` file doesn't seem necessary, even for Windows, so I deleted it. I tested without this file on Windows and the CLI still works. It was added probably due to some historical reason.

Examples of high-profile projects without a `.cmd` file:
- [eslint](https://github.com/eslint/eslint/tree/004577eda2f2f4b2829e0364f8b41893cebfc859/bin)
- [vite](https://github.com/vitejs/vite/tree/ef411cee2696af3ba791879fdae9aad165f178b2/packages/vite/bin)

packages/cli/src/bin/run-deprecated
```diff
@@ -0,0 +1,14 @@
...
+
+const path = require('node:path');
+
+(async () => {
+  const oclif = await import('@oclif/core');
+  await oclif.execute({ development: false, dir: __dirname });
+
+  console.warn(
+    '[DEPRECATION WARNING] ' +
+      'The executable name "zapier" is deprecated and will be remove in a future version. ' +
```
kola-er:
```suggestion
      'The executable name "zapier" is deprecated and will be removed in a future version. ' +
```

kola-er (approved):

PR #1183, opened 2025-10-17
PDE-6542 feat!(schema): replace `skipCleanArrayInputData` with `cleanInputData`

eliangcs:
Adds global and per-action `cleanInputData` flags to the schema. Example usage:

```js
const App = {
  flags: {
    cleanInputData: false,  // global flag
  },
  triggers: {
    recipe: {
      key: 'recipe',
      operation: {
        cleanInputData: true,  // only enable for this trigger
      },
    },
  },
  creates: {
    recipe: {
      key: 'recipe',
      operation: {
        cleanInputData: false,
      },
    },
  },
};
```

`cleanInputData` defaults to true in v18, so it's backward compatible with the current behavior. Developers won't need to change code when upgrading to zapier-platform-core v18, unless you use the experimental `skipCleanArrayInputData` flag.

❗ This is technically a breaking change, but only to those who are using the experimental `skipCleanArrayInputData` flag. But since the flag is experimental and only one app is using it, this change is not so breaking.

### DIFF
--- a/packages/core/types/schemas.generated.d.ts
+++ b/packages/core/types/schemas.generated.d.ts
@@ -302,6 +302,17 @@ export interface AppFlags {
    * `throwForThrottlingEarly` directly on the request options.
    */
   throwForThrottlingEarly?: boolean;
+
+  /**
+   * If true, Zapier removes empty strings, `null`, `undefined`, and
+   * empty Arrays or objects from `bundle.inputData` recursively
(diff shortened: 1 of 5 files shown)


kola-er (approved):
Looks good. From the ticket details, I could see the intention of adding some validation checks. Is that meant to be implemented as functional constraints checks or it is decided to be not necessary?

Natay (approved):

PR #1187, opened 2025-10-23
PDE-6640 - fix(cli) - Zapier push does not rely on localAppCommand

standielpls:
`zapier push` uses `zapier build` and `zapier upload` under the hood. However, we added `localAppCommand` in a [prior commit](https://github.com/zapier/zapier-platform/pull/1160/files#diff-df2d6779fb4e776ae0f642dcd697afdfe7a1f15b643221dbb04676c2a33207f5) which runs before `zapier build`, which caused `zapier push` to fail if it haven't been built yet. It can fail for esm apps if it has a entry that references the `dist/` directory

This PR fixes the issue by not using `localAppCommand` and adding a suggestion along in the error message.

If for whatever reason, you may not get this fix in and you're running into this, just run `zapier build` first and then run `zapier push`

### DIFF
--- a/packages/cli/src/oclif/ZapierBaseCommand.js
+++ b/packages/cli/src/oclif/ZapierBaseCommand.js
@@ -8,6 +8,7 @@ const { recordAnalytics } = require('../utils/analytics');
 const { getWritableApp } = require('../utils/api');
 
 const inquirer = require('inquirer');
+const { throwForInvalidVersion } = require('../utils/version');
 
 const DATA_FORMATS = ['json', 'raw'];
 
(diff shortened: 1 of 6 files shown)


standielpls:
> @standielpls since `zapier build` runs as part of `zapier push` and does call `localAppCommand()` [in these lines](https://github.com/zapier/zapier-platform/blob/88cc522c59e3239f8e5574bd274bb5d47ee0eb13/packages/cli/src/utils/build.js#L656-L660), could we instead call `throwForInvalidVersion()` on `rawDefinition.version` immediately after `rawDefinition` is defined? Does that make sense for this use case?
> 
> That would also have the benefit of throwing on an invalid version whether someone is running `zapier build` and then `zapier upload` separately, or if they're running `zapier push`.
> 
> Or, do we need the validation in `zapier push` specifically since that contains the snapshot label maybe? Lmk if I'm missing something!

Yeah I was thinking of doing it in `zapier push` because we have the snapshot label there and it was simpler to validate and move on, but you make a good point. Folks can bypass the invalid version check by using `zapier build and then zapier upload`. I think I'll refactor it so we push the validation later

packages/cli/src/oclif/ZapierBaseCommand.js
```diff
@@ -90,16 +91,7 @@ class ZapierBaseCommand extends Command {
...
-    if (
-      !version.match(
-        // this is mirrored in schemas/VersionSchema.js and developer_cli/constants.py
-        /^(?:0|[1-9]\d{0,2})\.(?:0|[1-9]\d{0,2})\.(?:0|[1-9]\d{0,2})(?:-(?=.{1,12}$)[A-Za-z0-9]+(?:-[A-Za-z0-9]+)*)?$/g,
-      )
-    ) {
-      throw new Error(
-        `${version} is an invalid version str. Try something like \`1.2.3\` or \`0.0.0-TICKET\``,
-      );
-    }
```
standielpls, line 102: I moved this to `utils/version.js`

PR #1188, opened 2025-10-24
chore(core): Fix Children Input Types when line items present

cwernert:
This adds typing for `children` input fields when line items are present.
Previously only the hoisted, flattened types at the top level were available, but this corrects that to match Platform behaviour.

### DIFF
--- a/packages/core/types/inputs.plainfields.test-d.ts
+++ b/packages/core/types/inputs.plainfields.test-d.ts
@@ -19,7 +19,12 @@ expectAssignable<singleRequiredChildren>({
   string_required: 'some_string',
   parent_input: [{ string_required: 'value1' }, { string_required: 'value2' }],
 });
-expectType<singleRequiredChildren>(singleRequiredChild);
+expectType<singleRequiredChildren>(
+  singleRequiredChild as {
+    string_required: string;
+    parent_input?: { string_required: string }[];
+  },
+);
 
 // Complete `children: [{...}]` case.
 const multipleRequiredChild = {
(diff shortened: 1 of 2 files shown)


packages/core/types/inputs.d.ts
```diff
@@ -52,12 +52,11 @@ type UnionToIntersection<U> = (
  * type result = LastInUnion<1 | 2>;
  * // 2
  */
-type LastInUnion<U> =
-  UnionToIntersection<U extends unknown ? (x: U) => 0 : never> extends (
-    x: infer L,
-  ) => 0
-    ? L
-    : never;
+type LastInUnion<U> = UnionToIntersection<
```
tkcranny: This MR changes a lot of formatting, and in a way against the standard prettier setup for this repo – I'd avoid making these changes.

packages/core/types/inputs.d.ts
```diff
@@ -240,8 +239,18 @@ export type PlainFieldContribution<$Field extends PlainInputField> =
...
+  // When line items are mapped into any child input, then the parent will also
+  // be included as an array of objects containing the children's inputs'
+  // contributions. This may or may not apply on a per-zap basis, so this parent
+  // array is an optional member of inputData.
+  Partial<
+    Record<
+      $Field['key'],
+      Array<PlainFieldArrayContribution<$Field['children']>>
+    >
+  >;
```
tkcranny: image
Looks like we'll want to wrap this in a `Simplify<…>`, I didn't realise the introduction of the `Partial` made it pretty inscrutable.

I.e.:
image

packages/core/types/inputs.plainfields.test-d.ts
```diff
@@ -14,7 +14,11 @@ type singleRequiredChildren = PlainFieldContribution<{
   key: 'parent_input';
   children: [{ key: 'string_required'; type: 'string'; required: true }];
 }>;
-expectType<singleRequiredChildren>(singleRequiredChild);
+expectAssignable<singleRequiredChildren>(singleRequiredChild);
+expectAssignable<singleRequiredChildren>({
+  string_required: 'some_string',
+  parent_input: [{ string_required: 'value1' }, { string_required: 'value2' }],
+});
```
tkcranny, line 21: `expectAssignable` is useful for testing compatible subtypes, but it's still useful to test the entire whole type too, I'd like to see some `expectType` tests as well. E.g.:
image

tkcranny (dismissed):
Looking good. We need to preserve the existing prettier formatting to reduce the diff, wrap the new result in a `Simplify`, and add some tests that check for the exact result type too.

packages/core/types/inputs.plainfields.test-d.ts
cwernert: I think I understand, so I gave it a go with the latest push. Would you mind taking a look to let me know whether it's what you were expecting? Apologies for the tedium, this is super unfamiliar to me so I'm sort of learning on the fly...

PR #1191, opened 2025-10-31
fix(cli): Fix npx resolution issue with dual binary entries (PDE-6667)

Natay:
## Problem

After introducing two binary entry points in v18.0.0 (`zapier` and `zapier-platform`), npx was unable to determine which executable to run when using `npx zapier-platform-cli`. This broke CI pipelines and any workflows relying on npx to run the CLI.

**Root Cause:** When a package has multiple bin entries pointing to different files, npx cannot determine which one is the "main" executable and fails with:

```
npm error could not determine executable to run
```

## Solution

Both binary entries now point to the same file (`./src/bin/run`), which allows npx to successfully resolve the executable.

### Deprecation Warning Implementation

The script detects which binary name was used to invoke it and shows a deprecation warning for the old `zapier` command:

#### Unix/macOS/Linux:
- Precise detection via `process.argv[1]` which contains the symlink name
- Shows warning **only** when invoked via `zapier` command
- No warning when invoked via `zapier-platform` command

#### Windows:
- Cannot reliably detect which wrapper (`.cmd`/`.ps1`) was invoked
- npm's wrappers both call the same script, and `process.argv[1]` only shows the script path, not the wrapper name
- Shows generic deprecation notice for **all** Windows users

## Changes

- Updated `package.json` to point both `zapier` and `zapier-platform` bin entries to `./src/bin/run`
- Modified `./src/bin/run` to detect the invoked binary name and show platform-appropriate deprecation warnings
- Added CI job to check npx compatibility 
- Removed the separate `./src/bin/run-deprecated` file (no longer needed)

## Testing

✅ `npx zapier-platform-cli` now works correctly  
✅ `npx .` from the CLI directory resolves successfully  
✅ Deprecation warning appears when using `zapier` command (Unix)  
✅ No deprecation warning when using `zapier-platform` command (Unix)  
✅ Generic deprecation notice shown on Windows (both commands)  
✅ Both `zapier` and `zapier-platform` commands execute the same CLI functionality  
✅ Tested on Windows, macOS

### DIFF
--- a/packages/cli/src/bin/run
+++ b/packages/cli/src/bin/run
@@ -1,36 +1,30 @@
 #!/usr/bin/env node
 
 const path = require('node:path');
-const fs = require('node:fs');
 
 (async () => {
   const oclif = await import('@oclif/core');
   await oclif.execute({ development: false, dir: __dirname });
 
   // Show deprecation warning for the old "zapier" binary name
-  if (process.platform === 'win32') {
-    // On Windows, we cannot reliably detect which wrapper (zapier vs zapier-platform) was invoked.
-    // Both npm-generated .cmd/.ps1 wrappers call the same script (this file), and process.argv[1]
-    // always points to the script path, not the wrapper name. The parent process is just the shell
-    // (powershell.exe/cmd.exe), and no environment variables are set to indicate which wrapper ran.
-    // Therefore, we show a generic notice for all Windows users rather than no notice at all.
-    console.warn(
-      '\n[DEPRECATION NOTICE] ' +
-        'The "zapier" command is deprecated and will be removed in a future version. ' +
-        'Please use "zapier-platform" instead.\n',
-    );
-  } else {
-    // On Unix-like systems (macOS/Linux), process.argv[1] contains the symlink name that was
-    // invoked, allowing us to detect precisely which command was used.
-    const scriptPath = process.argv[1] || '';
-    const binName = path.basename(scriptPath, path.extname(scriptPath));
+  // Users can suppress this warning by setting ZAPIER_SUPPRESS_DEPRECATION_WARNING
+  if (!process.env.ZAPIER_SUPPRESS_DEPRECATION_WARNING) {
+    const msg =
+      'The "zapier" command is deprecated and will be removed in a future version. ' +
+      'Please use "zapier-platform" instead.\n' +
+      'To suppress this warning, set ZAPIER_SUPPRESS_DEPRECATION_WARNING in your environment.\n';
     
-    if (binName === 'zapier') {
-      console.warn(
-        '\n[DEPRECATION WARNING] ' +
-          'The "zapier" command is deprecated and will be removed in a future version. ' +
-          'Please use "zapier-platform" instead.\n',
-      );
+    if (process.platform === 'win32') {
+      // On Windows, we cannot reliably detect which wrapper was invoked (zapier vs zapier-platform)
+      // because both wrappers point to the same script. Show a generic deprecation notice instead of a warning..
+      console.warn('\n[DEPRECATION NOTICE] ' + msg);
+    } else {
+      // On Unix, we can detect which symlink was used
+      const scriptPath = process.argv[1] || '';
+      const binName = path.basename(scriptPath, path.extname(scriptPath));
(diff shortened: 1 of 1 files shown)


packages/cli/src/bin/run
```diff
@@ -1,6 +1,20 @@
...
 
+const path = require('node:path');
+
 (async () => {
   const oclif = await import('@oclif/core');
   await oclif.execute({ development: false, dir: __dirname });
+
+  // Show deprecation warning if invoked via the old "zapier" binary name
+  const scriptPath = process.argv[1] || '';
+  const binName = path.basename(scriptPath, path.extname(scriptPath));
```
eliangcs: Note that this doesn't work on Windows. When I was working on https://github.com/zapier/zapier-platform/pull/1181, I found on Windows you can't get the executable name in `process.argv`.

Natay: Yeah, I wasn't able to get it working.

This is what I noticed happen in windows

- npm creates wrapper scripts (`zapier.cmd/.ps1` and `zapier-platform.cmd/.ps1`)
- Both wrappers execute: `node.exe "<path-to-script>" %args%`
- The wrapper launches Node with the actual script path
- When Node runs, `process.argv[1]` only sees the script file path, not which wrapper launched it
- The wrapper name is lost in the execution chain
- No way to detect which command was used

🤔 Here are the things I tried in the windows machine that did not work. 

```
### ❌ 1. process.argv[1]
**Attempted:** Check `process.argv[1]` for the wrapper script name
**Result:** Always points to `node_modules\zapier-platform-cli\src\bin\run` regardless of which command was invoked
**Why:** npm's Windows wrappers (`.cmd`/`.ps1` files) both call the same script directly

### ❌ 2. Parent Process Command Line
**Attempted:** Query parent process (PowerShell/CMD) to see which wrapper it executed
```javascript
wmic process where ProcessId=${ppid} get CommandLine

**Result:** Only shows `powershell.exe`, not which wrapper script it ran
**Why:** The parent process is the shell itself, not the intermediate wrapper

### ❌ 3. Environment Variables
**Attempted:** Check for env vars set by PowerShell wrappers (`PSCommandPath`, `PSScriptRoot`, etc.)
**Result:** None are set or passed through to the Node process
**Why:** npm's auto-generated wrappers don't set custom environment variables

### ❌ 4. Filesystem Access Timestamps
**Considered but rejected:** Check modification/access times of `zapier.cmd` vs `zapier-platform.cmd`
**Why rejected:** Too hacky, unreliable (caching, antivirus, concurrent executions), and fragile
```

Notably, having two different files would work here but fails for `npx`

I opted to show a **generic deprecation notice** here, but it shows for both commands. What do you think?

eliangcs: @Natay thanks for the thorough test! I'm also convinced there's no way for Windows to print the deprecation message conditionally without two separate `.js` files.

How about we show a generic deprecation notice **only on Windows**? On macOS or Linux,  we can still print the deprecation notice conditionally. The two `bin` entires would point to the same `.js` file:

```js
{
  "bin": {
    "zapier": "./src/bin/run",
    "zapier-platform": "./src/bin/run"
  }
}
```

The logic in `run.js` would be like:

```js
if (os.platform() === 'win32') {
  console.warn("<Generic deprecation notice>");
} else if (process.argv[1] === 'zapier') {
  console.warn("<Specific deprecation notice>");
}
```

Natay: @eliangcs Agreed, that is currently what I have in the [PR](https://github.com/zapier/zapier-platform/pull/1191/files#diff-a48d4a7103160e2c67a3dc346e2158c2104ed105a2e18f28675c2eb4e8286da3R11) ( my comment above was not the best at explaining this! )

eliangcs: @MaxwellGBrown mentioned to me that it'd be better if we have an option for developers using Windows to suppress the warning. Maybe we could use an environment variable?

Natay: Good idea, I'll add that in and update the docs to mention that!

Natay: @eliangcs I updated it so `ZAPIER_SUPPRESS_DEPRECATION_WARNING` can suppress this warning for all OS!

PR #1282, opened 2026-04-14
PDE-7035 feat(core): add getAuthTemplate and renderAuthTemplate commands

eliangcs:
Add `getAuthTemplate` and `renderAuthTemplate` commands to `zapier-platform-core`. These extract and render authentication templates from an app's definition by running the app's middleware with placeholder credentials and capturing the result.

Read more details in [this internal doc](https://gitlab.com/zapier/team-developer-platform/dev-platform-shared-brain/-/blob/main/team/working/changhung/PDE-7006-relay-auth-coverage/pr-description-internal.md).


packages/core/src/auth-template/render-auth-template.js
```diff
@@ -0,0 +1,228 @@
...
+
+  // Ensure all bundle fields are populated so middleware doesn't crash
+  // accessing e.g. bundle.inputData.someField or bundle.meta.isLoadingSample
+  const bundle = {
+    authData: eventBundle.authData || {},
+    inputData: eventBundle.inputData || {},
+    meta: eventBundle.meta || {},
+    subscribeData: eventBundle.subscribeData || {},
+    targetUrl: eventBundle.targetUrl || '',
+  };
```
standielpls, line 112: super nit: should we be doing this on the backend anyways?

packages/core/src/render-auth-template.js
```diff
@@ -0,0 +1,228 @@
...
+  ];
+
+  if (authType === 'basic') {
+    httpBefores.push(addBasicAuthHeader);
+  } else if (authType === 'digest') {
+    // Digest auth requires a challenge-response — can't capture statically,
+    // but we still run the middleware for renderAuthTemplate.
+    // For now, skip digest middleware since it makes a real HTTP request.
+  } else if (authType === 'oauth1') {
+    const oauth1SignRequest = require('./http-middlewares/before/oauth1-sign-request');
```
standielpls: Curious why does this one need an inline import?

packages/core/src/render-auth-template.js
```diff
@@ -0,0 +1,228 @@
...
+    };
+  }
+
+  const template = extractTemplate(capturedReq);
+
+  // Inline auth fallback: if pipeline produced an empty template and
+  // there's no middleware/requestTemplate, render from authentication.test
+  const hasBeforeRequest =
+    Array.isArray(compiledApp.beforeRequest) &&
+    compiledApp.beforeRequest.length > 0;
```
standielpls: Technically, `beforeRequest` can be an array of functions or just a function (though I am doubtful anyone actually implemented it as just a function) - I think we'd accomplish the same if we removed the Array.isArray() check wdyt?

packages/core/src/auth-template/render-auth-template.js
```diff
@@ -0,0 +1,228 @@
...
+      status: 200,
+      headers: {},
+      getHeader: () => undefined,
+      content: '{}',
+      data: {},
+      request: preparedReq,
+    });
+  };
+
+  const stubZ = {
```
standielpls, line 199: Any concern here for beforeRequest functions that make an API call **AND** then manipulate the `request` fields? My understanding is that this will stub any z.request call, so if we had that pattern, say a refresh token call to then add to the headers, this stub will always block the api call, and the app wont have the properly rendered fields.
Does this make sense? I am very skeptical that this is an actual concern, and likely an extreme edge case, but wanted to flag in case.

packages/core/src/render-auth-template.js
eliangcs: I did encounter a real-world example that implements `beforeRequest` as a function, which was why I added the check. And if `beforeRequest` is a function, `beforeRequest.length` [will be the number of arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) it expects, so we still need the `Array.isArray()` check.

packages/core/src/render-auth-template.js
eliangcs: Good catch! Fixed in [18abab0](https://github.com/zapier/zapier-platform/pull/1282/commits/18abab0b5b97a8fe12e25344d183f6560996fc1e). It also fixes a bug where unresolved curlies could get included in the output.

packages/core/src/auth-template/render-auth-template.js
eliangcs: Good point! I'll see if there's a way to only stub the main `z.request()` call. I haven't found a working solution yet.

packages/core/src/auth-template/render-auth-template.js
eliangcs: Do you mean we can just use `eventBundle` here and there's no need for `bundle`?

standielpls: I mean [here](https://gitlab.com/zapier/zapier/-/blob/a597ac1d7f9f718106422069e474cdd168cc237c/web/backend/zapier/app_platform/developer_cli/api_base.py#L1519) should we be ensuring bundle fields are populated in before passing into the lambda payload?

packages/core/src/get-auth-template.js
```diff
@@ -0,0 +1,1173 @@
...
+    return { supported: true, authType: null, source: 'none', template: {} };
+  }
+
+  // Digest and OAuth1 can't be expressed as static templates
+  if (authType === 'digest') {
+    return { supported: false, reason: 'digest', authType };
+  }
+  // if (authType === 'oauth1') {
+  //   return { supported: false, reason: 'oauth1', authType };
+  // }
```
standielpls: Should this be commented out?

packages/core/src/get-auth-template.js
```diff
@@ -0,0 +1,1173 @@
...
+
+  const placeholderAuthData = buildPlaceholderAuthData(auth);
+  let beforeRequestTemplate;
+  let beforeRequestFailed = false;
+
+  // --- Step 1: requestTemplate ---
+  // If the app declares a requestTemplate, that IS the auth template.
```
standielpls: What if the app declares a `requestTemplate` AND `beforeRequest` and they BOTH add auth creds into the request?

packages/core/src/auth-template/get-auth-template.js
```diff
@@ -0,0 +1,1173 @@
...
+              supported: false,
+              reason: 'middleware_not_static',
+              authType,
+            };
+          }
+          // else: fall through to authentication.test
+        } else {
+          // URL divergence check
+          const urlProbeTrue = createUrlProbe('https://example.com', true);
+          const urlProbeFalse = createUrlProbe('https://example.com', false);
```
standielpls, line 685: Looking for confirmation:
URL divergence check is used to evaluate whether or not there are url specific paths in a `beforeRequest` that may alter the request -  if there's divergence (and no fallback), we'll likely need to use `renderAuthTemplate` for these kind of apps?

packages/core/src/auth-template/get-auth-template.js
```diff
@@ -0,0 +1,1173 @@
...
+            // beforeRequest succeeded. Store the template — if authentication.test
+            // produces a superset (e.g., adds per-operation auth headers from
+            // legacy scripting hooks), we'll prefer that instead.
+            beforeRequestTemplate = cleanTemplate(template);
+          }
+        } // end else (proxy check passed)
+      }
+
+      // No auth placeholders survived, or divergence with auth.test available.
+      if (!auth.test) {
```
standielpls, line 738: curious, i've seen this pop up a bunch, `authentication.test` is a required field in the schema, so do we actually need the defensive checks here?

packages/core/src/get-auth-template.js
```diff
@@ -0,0 +1,1173 @@
...
+    capture({
+      url,
+      headers:
+        headers instanceof Headers
+          ? Object.fromEntries(headers.entries())
+          : headers,
+      method: init?.method || input?.method || 'GET',
+    });
+    return new Response('{}', { status: 200, headers: {} });
+  };
```
standielpls: I believe AWS Lambda function containers are isolated if there's 2 concurrent requests by spinning up a new container with its own Node process  - but wanted to double check. The risk is, if this patched the global fetch, will it affect a concurrent request that's using fetch to make an API call.

packages/core/src/auth-template/render-auth-template.js
```diff
@@ -0,0 +1,228 @@
...
+  const hasRequestTemplate =
+    compiledApp.requestTemplate &&
+    Object.keys(compiledApp.requestTemplate).length > 0;
+
+  if (
+    !hasBeforeRequest &&
+    !hasRequestTemplate &&
+    Object.keys(template).length === 0 &&
+    auth.test &&
+    typeof auth.test !== 'function'
```
standielpls, line 254: One edge case:
For apps with inline auth (defined in the request for each action), and auth.test **is** a function, I believe this will return an empty auth template. 
I think what's better is if we error here, instead of Relay executing an empty template.

standielpls: I'm not sure if the `getAuthTemplate` path also runs into this but would it be worth double checking given we have a chunk of inline auth apps?

packages/cli/src/oclif/commands/invoke/index.js
```diff
@@ -253,10 +255,20 @@ class InvokeCommand extends BaseCommand {
           }
           return;
         }
+        case 'template': {
+          const output = await templateAuth(context);
+          console.log(JSON.stringify(output, null, 2));
+          return;
+        }
+        case 'render': {
```
standielpls, line 307: My wires are all crossed for this comment but maybe you'll make some sense from it:

If we ran `zapier invoke auth render -a 123` - this is essentially Relay mode right? But the authData we pass in here will be [templated auth data ](https://github.com/zapier/zapier-platform/blob/231b688cfb9e70c9575cfba1df4c91a0aaf1bff3/packages/cli/src/oclif/commands/invoke/index.js#L176) - so would the render step just render back these templated strings?

packages/core/src/auth-template/get-auth-template.js
```diff
@@ -0,0 +1,1173 @@
...
+// should have survived.
+// Build a supported:true result, but check if auth fields were consumed
+// by encoding (e.g., base64). If so, override to supported:false.
+const supportedResult = (authType, source, template, auth) => {
+  if (template && Object.keys(template).length > 0) {
+    const s = JSON.stringify(template);
+    const hasAuthData = /\{\{\s*bundle\.authData\./.test(s);
+    const hasProcessEnv = /\{\{\s*process\.env\./.test(s);
+    const hasDeclaredFields =
+      auth && auth.fields && auth.fields.some((f) => f.key);
```
standielpls, line 86: Is this sufficient to denote whether there is "placeholders consumed" by only looking at auth.fields? What about the ones we get depending on auth type i.e. basic: username:password, oauth2: access_token?

packages/core/src/auth-template/get-auth-template.js
eliangcs: Exactly. It's a common pattern to use `req.url` to branch conditionally in `beforeRequest`:

```
if (req.url.startsWith('something')) {
  req.headers.authorization = '...';
} else if (req.url.includes('else')) {
  req.headers.authorization = '...';
}
```

The URL probe is used to detect this kind of branching.

packages/core/src/get-auth-template.js
eliangcs: This is a hard problem. I was thinking we should give integration developers a way to declare template-based auth if they want. So if they use `requestTemplate`, we'd use that for Relay and ignore `beforeRequest`.

But currently, when both `requestTemplate` and `beforeRequest` are defined, platform-core applies both of them (`requestTemplate` first, then `beforeRequest`). So what you're suggesting may actually happen.

I think correctness is more important here. I'll merge Steps 1 (`requestTemplate`) and 2 (`beforeRequest`) so we cover the case you mentioned. Thanks!

packages/core/src/auth-template/get-auth-template.js
eliangcs: The only time `authentication.test` might be undefined is during local development, when the developer hasn't pushed their code. When they run `validate` command, they'll see an error about the missing `authentication.test`. But when they run `invoke auth template`, I'd like to see it have a fallback output or show a meaningful error message.

packages/core/src/get-auth-template.js
eliangcs: I forgot to delete it. Deleted it in aa437cf8. Full implementation of OAuth1 requires signing, which is not supported by `getAuthTemplate`. But I found at least one example (Trello if I recall) that implements OAuth1 without signing, which is why this code was commented and is now deleted.

packages/core/src/get-auth-template.js
eliangcs: Lambda isolates concurrent requests as you noted, so I don't think there's an issue.

However, a container might be reused for the sequential request. If we didn't properly restore the patches, it'd pollute the next request. The patch restoration is done in [this `finally` block](https://github.com/zapier/zapier-platform/blob/aa437cf8f1ba05c55d551a1a6e26bcdca43f3a2c/packages/core/src/get-auth-template.js#L761-L765). There's also [restoration](https://github.com/zapier/zapier-platform/blob/aa437cf8f1ba05c55d551a1a6e26bcdca43f3a2c/packages/core/src/get-auth-template.js#L204) for the `process.env` patch too. I could use a second pair of eyes on that.

packages/cli/src/oclif/commands/invoke/index.js
eliangcs: `invoke auth template` and `invoke auth render` will only work in local mode. If you specify `-a` (relay mode) or `-r` (remote mode), you'll get an error. It was implemented in [9be012f](https://github.com/zapier/zapier-platform/pull/1282/commits/9be012f0f23624d442d1d96a99fd6bd8ea3e45cb).

By design, we never pass auth credentials from production to local, so `render -a 123` (rendering the request using authentication ID 123) is not actually applicable.

packages/core/src/auth-template/get-auth-template.js
eliangcs: Good question! I dug into this. Two parts:

**1. Standard placeholder coverage in `buildPlaceholderAuthData`.** Comparing against how the monolith auto-injects auth fields based on the auth type, our placeholder builder was missing `oauth1` (`oauth_token`, `oauth_token_secret`) and `custom + customConfig.sendCode` (`code`). I added those in 30c92774 so locally-run `getAuthTemplate` matches what the monolith would inject in production.

**2. Extending the consumed-fields check itself.** I tried it — adding standard fields to `hasDeclaredFields` so `type: 'basic'` apps without declared `auth.fields` would still trigger the consumed check. Ran the compare script against the top 250 apps (relay top 150 + invocations top 100):

| | Before | After |
|---|---|---|
| Matches | 231 | 230 |

One regression: [an app](https://gitlab.com/zapier/team-developer-platform/dev-platform-shared-brain/-/blob/0ea86ddb17c6713baeb6c091f3d5ebdd35d28343/team/working/changhung/PDE-7006-relay-auth-coverage/devlog.md#trying-to-extend-the-consumed-fields-check) uses `process.env` for its token exclusively and never references the standard `access_token` field. The check can't distinguish "field was base64-encoded" from "field was never used" — both produce the same result (`process.env` present, no `bundle.authData`).

So I reverted that part. Only counting **declared** `auth.fields` keeps it as a real "developer expects this in the request" signal. Added a comment explaining the trade-off. The hypothetical case (basic-auth app with no declared fields + base64 middleware) didn't show up in the top ~200 — if we hit one later we can revisit, probably with a heuristic that looks for actual encoding evidence rather than absence of placeholders.

packages/core/src/get-auth-template.js
standielpls: Indeed this is a hard problem, and agreed, the correctness of merging 1+2 would be better. Alternatively, we could also check to see if there is _any_ `beforeRequest` functions defined. If none, then we can move return at Step 1. 

In an ideal world, I would love to offer developers a way to define "auth middleware" that is uniquely isolated away from other request middleware. 
I noticed a painful misuse of `beforeRequest` where developers add `Authorization Bearer {{bundle.authData.access_token}}` in there over using the `requestTemplate`

packages/core/src/get-auth-template.js
standielpls: Hmm it looks fine to me - though my AI review bot mentioned that for the patching, we may have an edge case if we ever refactored the code to call them inside a `Promise.all()`:

```
getAuthTemplate mutates global state — process.env, http.request,
  https.request, fetch — to capture what an app's auth pipeline would
  emit, then restores the originals in a finally block. This
  save-and-restore is safe when calls run sequentially, but it breaks
  under Promise.all: if two captures run concurrently, the second one's
   "save" picks up the first one's patch instead of the real value, so
  when both finally blocks run, the real value never gets restored —
  leaving the global pinned to a stale patch. Today this only actively
  leaks for process.env (because only runMiddlewareSurvival is called
  in Promise.all), but the http/fetch patches share the same
  vulnerability and would leak the same way if anyone ever wrapped
  runTestFunctionSurvival in parallel calls.

...

In practice, the current impact is small. The CLI is a one-shot
  process — it leaks the proxy on exit and the OS reclaims everything,
  so no observable effect. AWS Lambda is the same: each container
  serializes invocations on its own V8 heap, so the leak is wiped
  between calls. The only places it actually surfaces are (a) test
  harnesses that exercise the URL-divergence path and then read
  process.env afterwards (subsequent reads of unset vars return
  placeholder strings instead of undefined), and (b) hypothetical
  long-lived in-process hosts that nobody runs today. So the concern is
   more about future-proofing than fixing a live bug — worth flagging
  in the review and serializing the Promise.all defensively, but not a
  merge blocker.
```

standielpls:
I've tested it locally, it looks great! Thanks! 
I think once we've addressed [this](https://github.com/zapier/zapier-platform/pull/1282#discussion_r3144790597), we can release.

packages/core/src/render-auth-template.js
eliangcs: I just realized I misunderstood your comment. Fixed in 49863f22636f9. Thanks!

packages/core/src/auth-template/render-auth-template.js
eliangcs: I see. It's not needed technically, but I think this code can serve as a hint for the expected data shape.

packages/core/src/get-auth-template.js
eliangcs: > Alternatively, we could also check to see if there is any beforeRequest functions defined. If none, then we can move return at Step 1.

I think your alternative makes more sense. Implemented in [6e284a7](https://github.com/zapier/zapier-platform/pull/1282/commits/6e284a78d3f556e29541eb4c8cc253cdc1dc5c04).

packages/core/src/auth-template/render-auth-template.js
eliangcs: Good catch! Fixed in [ef6d5b5](https://github.com/zapier/zapier-platform/pull/1282/commits/ef6d5b5776495c33865df72457df00389a2a9b60) (line 228). Instead of erroring, I think the expected result should be actually returning the template with `{{bundle.authData.x}}` placeholders.

I double checked, `getAuthTemplate` doesn't have the same issue.

packages/core/src/get-auth-template.js
eliangcs: Noted!

packages/core/src/auth-template/render-auth-template.js
eliangcs: Fixed in ea689ffc0. `renderAuthTemplate` should use a working `z.request()` rather than a stubbed `z.request()`.

eliangcs:
@standielpls fyi, I'll add test cases for both `getAuthTemplate` and `renderAuthTemplate`.

packages/core/src/auth-template/get-auth-template.js
```diff
@@ -0,0 +1,997 @@
...
+// restores them by checking if empty param values correspond to known
+// auth fields from placeholderAuthData.
+const restoreStrippedParams = (template, placeholderAuthData) => {
+  if (!template.params) {
+    return;
+  }
+  for (const [key, value] of Object.entries(template.params)) {
+    if (value === '' || value === null || value === undefined) {
+      // Check if this key matches an authData field
+      const placeholder = placeholderAuthData[key];
```
standielpls: I think this will be buggy if the template uses a key that **doesn't** match the auth data one. See my general comment for more info

standielpls:
So sorry, I did some more manual testing given the new commits on a test app and found a new small bug worth flagging. 

If you have both `requestTemplate.params` and a `beforeRequest`, the param gets dropped silently when its key name doesnt match the authData field name like params: 
```
{ api_key:  '{{bundle.authData.access_token}}' }
``` 
AND if there's a `beforeRequest` header, the `api_key` just disappears. 
If you rename the param key to `access_token` it works. I figured its possible for developers may map key names to a different auth values so this would break.
I poked at it and I believe the cause is `restoreStrippedParams` matching by key name against what's available in the `authData`. 

There's two ways I think we can fix this:
1. Replace the curly destroy and replace behaviour with encode and decode sentinel strings. 
    Main value is fewer moving parts. With curlies, the flow is: core's `normalize` destroys them, then we manually re-create them after capture, and we have to keep these two operations in sync. [This PR](https://github.com/zapier/zapier-platform/pull/1296) demonstrates what this may look like

2.  Move the merge-back into `cleanTemplate` and point it at both compiledApp.requestTemplate.params and testReq.params. Same idea as the existing merge-back, just generalized so it covers both auth sources instead of only one.

I'm fine either one! LMK if I can help clarify Option 1 better.

eliangcs:
@standielpls nice catch, thanks! I merged and tested your PR. I found the sentinel strings you used caused a regression with real-world apps.

To reproduce, I added [two test cases](https://github.com/zapier/zapier-platform/pull/1282/commits/fbe8870764344e8ae0c534628b3bbc96e30ed663):

1. "captures requestTemplate params whose key differs from the authData field"
2. "captures params when test URL embeds an authData field in the hostname"

My branch was failing on 1 (before merging your branch) while yours was failing on 2. It was because:

- Test 1: my `restoreStrippedParams` looks up `placeholderAuthData[param_key]`, so when the param key (`api_key`) doesn't match the authData field (`access_token`) the lookup misses and the param gets dropped. Your sentinel approach sidesteps this — the value carries its own field identity, independent of the param key.

- Test 2: the colon-separated sentinel format (`__placeholder_auth:KEY:end_placeholder__`) breaks URL parsing when an authData placeholder lands in a hostname. After substitution the URL becomes `https://__placeholder_auth:domain:end_placeholder__.example.com/`, `new URL()` throws on the colons in the hostname, and `extractTemplate`'s try/catch swallows the error — params extracted from the URL are silently lost.

To fix, I [changed](https://github.com/zapier/zapier-platform/pull/1282/commits/e76adb76434a032f2c18e70d616232365306b10d) the sentinel strings to use double-underscores instead of colons as the separator (`__placeholder_auth__KEY__end_placeholder__`). Underscores are valid in URL hostnames, so a sentinel survives substitution into any URL position without breaking `new URL()`. Lowercase to match `new URL()`'s hostname normalization. Regex switched from `[^:]*` to a non-greedy `.+?` since underscores are common in real keys.

standielpls (approved):
Thank you! This was a much more complex one than I anticipated - but I appreciate your diligent approach to use the test bed to validate it.

PR #1303, opened 2026-05-20
feat(auth): add OIDC federation authentication support

joseph-zapier:
Add OIDC federation authentication support

PR #1306, opened 2026-05-25
[PDE-7120] feat(auth): add OIDC federation authentication support

joseph-zapier:
## What

Adds `oidcFederation` as a new authentication type for Zapier integrations that use OIDC token exchange to obtain cloud provider credentials (e.g. AWS STS, GCP, Azure, Vault).

## Changes

**schema**
- New `AuthenticationOIDCFederationConfigSchema` with required `perform` and `audience` fields
- Added `oidcFederation` to the `type` enum and `oidcFederationConfig` property in `AuthenticationSchema`
- Regenerated `exported-schema.json`, `schema.md`, and TypeScript types

**core**
- `get-auth-template.js`: builds placeholder `authData` sentinels for common OIDC credential keys (AWS STS, GCP/Azure, Vault) so request templates resolve correctly during testing
- `create-app-request-client.js`: enables auto-refresh for `oidcFederation` auth the same way `session` auth does

**cli**
- `auth-files-codegen.js`: generates `authentication.js`/`.ts` and `middleware.js`/`.ts` scaffolding for `oidcFederation` integrations via `zapier init`
- `generators/index.js`: registers the new auth type in the generator
- `describe.js`: includes `authentication.oidcFederationConfig.perform` in described paths

**Note:** `zapier invoke` is not supported for `oidcFederation` — the auth type relies on a token injected by Zapier infrastructure that isn't available in local invocation.

## Testing

- Added unit tests in `packages/core/test/auth-template/get-auth-template.js` and `packages/core/test/create-app.js`
- Added generated test templates `oidc_federation.test.js` / `.ts`
- Schema validated via `packages/schema/test/index.js`

Note: A previous [PR](https://github.com/zapier/zapier-platform/pull/1303) was created for this work due to insufficient permissions at the time.

### DIFF
--- a/packages/cli/src/generators/index.js
+++ b/packages/cli/src/generators/index.js
@@ -126,6 +126,7 @@ const authTypes = {
   'oauth1-trello': 'oauth1',
   oauth2: 'oauth2',
   'session-auth': 'session',
+  'oidc-federation-auth': 'oidcFederation',
 };
 
 const writeGenericAuth = (gen) => {
@@ -224,6 +225,7 @@ const TEMPLATE_ROUTES = {
   openai: writeForStandaloneTemplate,
   'search-or-create': writeForStandaloneTemplate,
   'session-auth': writeForAuthTemplate,
+  'oidc-federation-auth': writeForAuthTemplate,
 };
 
 const ESM_SUPPORTED_TEMPLATES = ['minimal'];
@@ -236,6 +238,7 @@ const TS_SUPPORTED_TEMPLATES = [
   'oauth1-trello',
   'oauth2',
   'session-auth',
+  'oidc-federation-auth',
 ];
 
 const TEMPLATE_CHOICES = Object.keys(TEMPLATE_ROUTES);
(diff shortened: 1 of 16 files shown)


kola-er:
Looks good. Thanks for the detailed description. To prevent mistakenly merging, I will delay the approval till the other team's work for the complete support is ready and we are ready to publish in a new version.

kola-er:
Btw, have you been able to test with an integration using the oidc auth type against the backend support? Instead of using `zapier-platform push`, you could manually upload a locally built zip file to your test integration.

joseph-zapier:
> Btw, have you been able to test with an integration using the oidc auth type against the backend support? Instead of using `zapier-platform push`, you could manually upload a locally built zip file to your test integration.

I have manually tested the new auth type for OIDC vs the backend, we're getting an [error](https://cdn.zappy.app/04deca39e007308ba534479e93ff9297.png) but it is expected cuz the token creation isn't implemented yet. What we can see is that the audience is getting through the error

cc: @kola-er

PR #1340, opened 2026-09-18
fix(core) Keep declared env values out of the auth template

kola-er:
Screenshot 2026-09-21 at 1 52 22 PM

## The problem

The app's environment is loaded during auth template capture — natively on the function, or copied out of the event by `applyEnvironment` — so middleware reading `process.env.SOME_VAR` receives the real value, not a placeholder. `hasAuthPlaceholders` passes if the template contains *any* placeholder anywhere; it does not require every value to be one. So a template like

```js
Authorization: Bearer {{bundle.authData.access_token}}   <- passes the check
X-App-Key:     <real value of a declared env var>        <- rides along
```

is cached and served to every connection with a live secret inside it. Nothing downstream inspects values: `cleanTemplate` only converts sentinels to curlies, and the caller caches any response carrying a `supported` key.

Confirmed in code; no instance confirmed in a cache.

A template with *no* placeholders is already safe — `hasAuthPlaceholders` is false and `auth_fields_consumed` is returned, so an app whose auth is purely a server-side secret is rejected rather than cached. Only the mixed case leaks.

## Why not scrub non-placeholder values

Integrations legitimately send constants on every request — a fixed `Prefer` header, a fixed `Host`. Removing every value that is not a placeholder would make those templates diverge from what the app actually does.

## The fix: reverse-map

A captured value that equals a declared env var's value becomes `{{process.env.NAME}}`, which the backend resolves at render time. Secrets stay out of the cache, the template stays faithful, and genuine constants match no env value and are left alone.

- `buildEnvPlaceholderPairs` builds `[value, placeholder]` pairs from `bundle.declaredEnvNames`, taking values from an environment snapshot — taken before any capture runs, so it is the real environment rather than the `withProxiedEnv` Proxy — unioned with `event.environment`.
- `withEnvPlaceholders` replaces matches **as substrings**, so a value embedded in a larger string (`v1=<secret>=end`, a URL) is caught, not just a whole-value match. Pairs are sorted longest-value-first so a shorter value cannot clobber part of a longer one.
- Applied in the single `supported(...)` funnel that every supported return already goes through, so no path can skip it. The gating checks still run on the raw template, so no app's supported/unsupported verdict changes.

**`declaredEnvNames` is the only source; there is no inference fallback.** Guessing which `process.env` keys belong to the app would need a minimum-length guard — leaving short secrets in the template — and an exclusion list for the runtime's own variables, or core would emit a `{{process.env.NAME}}` that nothing can resolve at render time and break that
app's requests. The caller knows which names it owns and can resolve, so core does not guess: with no names, nothing is mapped.

## Scope correction found while testing

**Params were never exposed.** `extractTemplate` drops a param whose value carries no placeholder, and `findStrippedDerivedAuthParams` then returns `supported: false, reason: 'stripped_derived_params'`, so a real env value in a param already demotes the app instead of being cached. Headers and body were the exposed surfaces; both are covered now, and the params behaviour is pinned by a test so it stays that way.

## CLI

`zapier invoke auth template` passed `bundle: {}`, which would have made it the one caller with no protection. It now sends the names too: `invoke/index.js` already parsed `.env` via dotenv and discarded the result, so it keeps the names on `context.declaredEnvNames`, and `auth/template.js` puts them on the bundle.

The names come from the parsed `.env` rather than `process.env` deliberately: the latter is the whole shell environment (80 keys against 3 on a typical machine), and claiming `PATH` or `TERM_PROGRAM` is app-owned would reintroduce exactly the mis-mapping the no-inference rule avoids. Values still come from `process.env`, which dotenv has populated from the file. `authData_*` entries are excluded — that prefix is the command's stand-in for a connection, not app environment.


packages/core/src/auth-template/get-auth-template.js
```diff
@@ -425,13 +425,87 @@ const buildProxyAuthData = (placeholderAuthData, knownAbsentKeys = new Set()) =>
...
+    return value;
+  }
+  if (typeof value === 'string') {
+    let replaced = value;
+    for (const [envValue, placeholder] of envPlaceholderPairs) {
+      if (replaced.includes(envValue)) {
+        replaced = replaced.replaceAll(envValue, placeholder);
+      }
+    }
+    return replaced;
```
pedro-louro: Is there a chance that `replaceAll()` could wrongly replace placeholders previously replaced? 
e.g. 
- in the first iteration of the loop, the `envValue = hello` is replaced by `placeholder ={{process.env.world}}`
- in the second iteration, the is `envValue = world` by coincidence, and because `replaceAll()` is called on `replaced`, it also replaces the `placeholder ={{process.env.world}}` generated from the first loop iteration, generating a wrong {{env.variable}}. 

The chances of this coincidence seem very small, but if it does happen, it would generate wrong placeholders. 🤔

pedro-louro: Also,  could this pick the wrong env var if two env vars have the same value? 
Since the template cached depends on this, it could cache the wrong placeholder. This would still work well (due to having the same value), up until one of the env vars is changed  and we would send the wrong secret. 

This also like something with low chances of happening, but still seems like a possible cause of creating issues.

kola-er: Thank you. I've pushed a fix to handle the first concern. Not sure how we can handle the second. I'm fine with the trade-off as it is less likely to occur.

pedro-louro (approved):

